PR #2, opened 2020-05-02
WIP: Cleanup and finalize POC with functions, classes, and tests

mds1:
To do:

- [x] Add classes to manage values and keys to avoid the repeated formatting of values
- [x] Add tests
- [ ] Occasionally some random runs fail in check 2, which is stealthWallet.address === utils.getAddress(stealthAddress2). Need to figure out the cause of this bug and resolve it.


mds1:
@apbendi For demo purposes I went ahead and published this as an npm package with the name `umbra-protocol`. So when you review this you can try two things:

1. Clone the repo and run tests
2. Outside of this repo, follow the steps in the README to install and use umbra in a simple node script

mds1:
> Running npm install caused the lock file to change, so I pushed a commit with that in sync with the package file.

Awesome, thanks!

> I didn't experience this with about 10 tries. Maybe we can try writing a script that automates this step, runs it a million times and records the failure cases & rates so we can look for any patterns?

Yea I was thinking that too. Maybe we merge this and move that work to a separate branch/PR though?

mds1:
> Occasionally some random runs fail in check 2, which is stealthWallet.address === utils.getAddress(stealthAddress2). Need to figure out the cause of this bug and resolve it.

This will be tracked in a separate issue

Issue #7, opened 2020-05-05
Confirm that protocol works as expected for all inputs


mds1:
When developing the current package, there were a few unexpected failures. Need to check and see if those are still present, and if so figure out the cause and resolve it

mds1:
Closed by #8

PR #8, opened 2020-05-06
Contract Setup

apbendi:
This PR updates the projects directory structure to give a top level directory for "umbra-js" and another for contracts. It stubs an empty `Umbra` contract and tests.

Thought it would be good to merge this first, since it changes the structure so significantly, in case you're doing anything in parallel. Would get ugly trying to merge later.

To make sure contract setup is working for you:

```bash
cd contracts
npm install
npm test
```


mds1:
Looks good! Added the `npx` prefix to the scripts so the OpenZeppelin CLI doesn't need to be installed globally.

Also pushed a fix for #7 on this branch—The problem was when you have a private key with leading zeros they'll get stripped and that causes issues. Added a test that checks 1000 random numbers and wallets to look for any failures as well to help catch any other edge cases.

If it's good with you, I'll also push an `app` folder to this branch to scaffold the frontend. That way this PR quickly finalizes the repo structure

mds1:
Ok, everything should be good to go here. Feel free to review the last few commits and merge unless you have questions

apbendi:
LGTM thanks!

PR #9, opened 2020-05-06
Can now specify recipient by tx hash and recover public key

mds1:
Add support for specifying recipient / generating `KeyPair` instance by providing a transaction hash. The public key of the sender of the specified transaction is recovered and used to instantiate the `KeyPair` instance.

### DIFF
--- a/umbra-js/classes/KeyPair.js
+++ b/umbra-js/classes/KeyPair.js
@@ -28,59 +28,18 @@ function pad32ByteHex(hex) {
   return hex.padStart(64, 0);
 }
 
-/**
- * @notice Given a transaction hash, return the public key of the transaction's sender
- * @dev See https://github.com/ethers-io/ethers.js/issues/700 for an example of
- * recovering public key from a transaction with ethers
- * @param {String} txHash Transaction hash to recover public key from
- */
-async function recoverPublicKeyFromTransaction(txHash) {
-  // Get transaction data
-  const tx = await provider.getTransaction(txHash);
-
-  // Get original signature
-  const splitSignature = {
-    r: tx.r,
-    s: tx.s,
-    v: tx.v,
-  };
-  const signature = ethers.utils.joinSignature(splitSignature);
-
-  // Reconstruct transaction data that was originally signed
-  const txData = {
-    chainId: tx.chainId,
-    data: tx.data,
-    gasLimit: tx.gasLimit,
-    gasPrice: tx.gasPrice,
-    nonce: tx.nonce,
-    to: tx.to, // this works for both regular and contract transactions
-    value: tx.value,
-  };
-
-  // Properly format it to get the correct message
-  const resolvedTx = await ethers.utils.resolveProperties(txData);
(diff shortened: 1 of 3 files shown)


umbra-js/classes/KeyPair.js
```diff
@@ -47,24 +91,39 @@ class KeyPair {
...
+      // PUBLIC KEY
+      // Save off public key as hex, other forms computed as getters
       this.publicKeyHex = key;
-      this.publicKeyHexCoords = {
-        x: pad32ByteHex(this.publicKeyHexSlim.slice(0, 64)),
-        y: pad32ByteHex(this.publicKeyHexSlim.slice(64)),
-      };
+    } else if (key.length === 66 && isTxHash) {
+      // TRANSACTION HASH
+      return (async () => {
```
apbendi, line 106: This line effectively means that sometimes the constructor returns a Promise, while sometimes it returns an instantiated instance. To me, a constructor should always do the latter. But is this a common pattern in JS or something?

If I were implementing this, I would have left the constructor as non-asynchronous, and used something like a static async generator function to return an instance created from a txHash. What do you think?

umbra-js/classes/KeyPair.js
```diff
@@ -27,17 +28,60 @@ function pad32ByteHex(hex) {
...
   /**
-   * @notice Creates new instance from a public or private key
-   * @param {String} key public or private key to initialize instance with, as hex with prefix
+   * @notice Creates new instance from a public key, private key, or transaction hash
+   * @param {String} key Can be either (1) hex public key with 0x04 prefix, (2) hex private
+   * key with 0x prefix, or (3) transaction hash
+   * @param {Boolean} isTxHash must be true if key is a transaction hash
    */
-  constructor(key) {
+  constructor(key, isTxHash = false) {
```
apbendi, line 78: Generally I consider boolean flags being passed to functions to be a code smell. Maybe combined with extracting the async action, as mentioned in my other comment, we can avoid requiring this?

umbra-js/classes/KeyPair.js
mds1: I doubt this is a common pattern as I haven't come across it yet, so it probably isn't the most elegant approach and I'm open to other solutions.

I'm not totally sure I follow your suggestion—are you saying to use a two step approach where you'd do something like this?

```javascript
const publicKey = await KeyPair.recoverPublicKeyFromTransaction(txHash);
const keyPairInsance = new KeyPair(publicKey);
```

umbra-js/classes/KeyPair.js
mds1: Agree with this, but can probably avoid it using whatever solution we decide upon in the other comment

umbra-js/classes/KeyPair.js
apbendi: Yup, basically. Where the two lines above would live in a static factory function.

mds1: Updated in fd5508b

PR #10, opened 2020-05-08
MVP Payments Only Contract

apbendi:
Barebones implementation for pay to address (no CREATE2 fun) payments contract.


contracts/contracts/Umbra.sol
```diff
@@ -1,10 +1,63 @@
...
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/SafeERC20.sol";
 
-contract Umbra is Ownable {
+contract Umbra is Ownable, ReentrancyGuard {
+    using SafeMath for uint256;
 
-    constructor() public {
+    event EthAnnouncement(address indexed receiver, uint256 indexed amount, string note);
+    event TokenAnnouncement(address indexed receiver, uint256 indexed amount, address indexed token, string note);
```
mds1, line 13: Should we only have one type of event announcement, that way it's a little cleaner when scanning events from the frontend? ETH Announcements can either use the zero address or `0x00eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` which is what Kyber uses

contracts/contracts/Umbra.sol
```diff
@@ -1,10 +1,63 @@
...
+import "@openzeppelin/contracts-ethereum-package/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/SafeERC20.sol";
 
-contract Umbra is Ownable {
+contract Umbra is Ownable, ReentrancyGuard {
+    using SafeMath for uint256;
 
-    constructor() public {
+    event EthAnnouncement(address indexed receiver, uint256 indexed amount, string note);
```
mds1, line 12: We might be able to change `string` to `bytes` which should save some gas, but we can wait until we have actual encrypted payloads to test with to ensure that works

apbendi:
Updated to unify events and use `byte` arguments for announcement data. Let me know what you think @mds1.

contracts/contracts/Umbra.sol
```diff
@@ -9,8 +9,18 @@ import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC20/SafeERC20
...
     using SafeMath for uint256;
 
-    event EthAnnouncement(address indexed receiver, uint256 indexed amount, string note);
-    event TokenAnnouncement(address indexed receiver, uint256 indexed amount, address indexed token, string note);
+    event Announcement(
+        address indexed receiver,
+        uint256 indexed amount,
+        address indexed token,
+        bytes16 iv,  // Inivitalization Vector
+        bytes32 pk0, // Ephemeral Public Key
```
mds1, line 17: Small thing—instead of `pk0` and `pk1` we should use `pkx` and `pky`, just so it's more clear which is the x-coordinate of the public key and which is the y-coordinate.

Aside from this, looks and works great!

apbendi: Good call, done!

PR #12, opened 2020-05-11
Frontend ens support

mds1:
This branch got pretty big, so there's lots of changes here. It's probably sufficient to just review the code changes in the `umbra-js` folder, then try to use the app.

What's new in `umbra-js`:
- Add ENS helper functions
- Hardcoded to use Ropsten for now
- Exports protocol constants (i.e. the ENS signature key, the ENS bytecode key, and the message to be signed)

What's new in the app:
- Pretty much all of it
- Hardcoded to use Ropsten for now
- Use "ENS Setup" page to associate a signature with your ENS domain. This won't currently help you get a domain, so you may need to buy one on Ropsten first.
- Use the "Send" page to pretend you're sending funds to someone. To identify the recipient, you can enter either an ENS domain, transaction hash, address, or public key. Then you can choose a token (currently just ETH, DAI, and USDC) and enter an amount


mds1:
> A couple minor things came up. When running the umbra-js tests, though everything passes, I see this message several times while running the ENS functions test.

Yea as you could probably tell this is just because those functions are overloaded in the PublicResolver ABI. I added `ethers.utils.Logger.setLogLevel('error');` to turn off those warnings.

> Speaking of the frontend, after I got a Ropsten ENS name set up, I couldn't get past the button spinning forever like this. Note, I have NOT submitted any transaction. This is what it does as soon as I connect my account:

Whoops, temporarily hardcoded `isPending: true` at the end to clean up the UI, and forgot to change it back! Both of these are fixed in ba219f4

apbendi:
Looks and works great 👍

PR #14, opened 2020-05-13
More frontend updates

mds1:
Changes:

- Add interface for entering private key with placeholder explaining what will happen afterwards
- Add warning to header so users do not use accounts that hold funds on mainnet
- Add tutorial to home page explaining how to use the app
- Small change to login flow so you can navigate to pages without logging in first
- Cleanup CSS and improve mobile appearance

### DIFF
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -5,7 +5,7 @@
       class="q-my-sm"
       filled
       :label="label"
-      lazy-rules
+      :lazy-rules="lazyRules"
       :rules="[val => rules(val)]"
       :type="type"
       @input="handleInput"
@@ -39,6 +39,12 @@ export default {
       type: String,
       required: true,
       default: undefined,
+    },
+
+    lazyRules: {
+      type: Boolean,
+      required: false,
+      default: true,
     },
 
     rules: {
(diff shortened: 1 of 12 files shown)


mds1:
That is actually the first commit! Arguably that feature ("Handle entering private key") is still a WIP though because it doesn't actually do anything with the private key yet, since we just need the contract ready to do that. Let me know if you still think it should be changed

apbendi:
> Arguably that feature ("Handle entering private key") is still a WIP though because it doesn't actually do anything with the private key yet, since we just need the contract ready to do that. Let me know if you still think it should be changed

Got it, this makes sense. In my head "WIP" basically means a commit that itself isn't completed. I would have worded what you're describing something like "Partial implementation of..." or "Initial implementation of..." 

I know this is super annoying, but do you mind rewording? 🙈🙈🙈 Maybe it's pure OCD brain but something about seeing "WIP" in a commit message makes it feel slipshod. (Although apparently I didn't notice one in the last PR so apparently I'm not paying very close attention 🤷‍♂️).

mds1:
Ok, just reworded the commit message! Let me know how that looks.

apbendi:
> Ok, just reworded the commit message! Let me know how that looks.

Well, actually, can you also change.....

Just kidding, looks great— sorry for being annoying!

PR #16, opened 2020-05-19
Implement pbkdf

### LINKED ISSUE #15: Implement PBKDF2 support on frontend
The current app requires users to manually copy/paste their private key, which of course is not good. Instead, we'll use the following workflow. Steps with an asterisk are user actions, and steps without an asterisk happen automatically behind the scenes.

@apbendi Can you review this flow and let me know how it looks?

You'll notice one change here that we haven't discussed is that the app private key is now associated with their specific web3 wallet that owns the ENS address. This association would only live locally and be encrypted and saved in `localStorage`. This association is not required by the protocol for any reason, but I added it because it should provide an extra bit of security (at the cost of UX) by requiring password + correct web3 login before withdrawing funds. Curious to hear what you think of this.

## Account Setup

1. *User visits website
1. *User visits account setup page
1.  Prompt user to connect wallet
1. *User logs in with the wallet of their choice
    1. If user does not login with a web3 wallet, exit this flow and inform user they cannot do any account setup
1. Check local storage for encrypted key
    1. If found, setup was completed so we inform the user they do not need to set anything up
    1. Ability to repeat setup (e.g. if password is lost) will not be implemented yet
1. Check if wallet address has an ENS domain associated with it
1. *If not, tell user to acquire an ENS domain and configure it with `PublicResolver` and ensure the reverse record is set
1. Once ENS setup is complete, prompt user for a password, and ask them to enter it twice for confirmation
1. *User enters password twice
1. Generate random private key with `ethers.Wallet.createRandom()`
1. Encrypt key using the password with PBKDF2 (can do this with [browser-passworder](https://github.com/danfinlay/browser-passworder))
1. Save encrypted private key and associated web3 wallet address to `localStorage`
1. Read encrypted private key from `localStorage`, decrypt it, and verify the key and associated address match what was expected (this ensures nothing went wrong in the process)
    1. If this fails, discard password and private key, clear local storage, then throw an error and ask user to try again
1. Prompt user to save file containing the encrypted contents, and make sure they know that if they lose this file or lose the password they will lose access to any non-withdrawn funds and must repeat the setup process
1. *User downloads encrypted file
1. TBD Provide user the option of saving an unencrypted file with the private key as backup?
1. Use unencrypted private key to sign message on behalf of the user (since this is a raw private key and not a connected web3 wallet, the user will not be prompted for any confirmation)
1. Prompt user to send a transaction (from their regular wallet) which associates that signature with their ENS domain under the `vnd.umbra-v0-signature` record.
1. *User submits transaction
1. *User can now provide their ENS address to someone to send them funds

## Withdrawing Funds, Workflow 1

This is the primary, recommended workflow shown to users by default

1. *User visits website
1. *User visits withdraw page
1. Prompt user to connect wallet
1. *User logs in with the wallet of their choice
    1. If user does not login with a web3 wallet, exit this flow and inform user they cannot withdraw anything
1. Check `localStorage` for encrypted key
    1. If not found, exit this flow, inform user they need to finish setup before they can withdraw funds, and provide a link to the setup page
1. Prompt user for password to decrypt key
1. *User enters password
1. Decrypt key from `localStorage` and ensure the associated wallet address matches the one they are currently logged in with
    1. If not, dispose decrypted info and inform user of the issue
1. Create `KeyPair` instance from this private key
1. Check `localStorage` to see the last block we scanned events from
1. Scan for all Umbra announcements and parse out emitted values
1. Call `keyPair.decrypt(...)` for each announcement
1. If decrypted output starts with `umbra-protocol-v0`, it's for this private key
1. Save off block number of the last scanned block in `localStorage`
1. Show a table of accessible funds with details and withdraw options
1. *User performs withdrawals

## Withdrawing Funds, Workflow 2

To expand the set of potential recipients, the above app private key + ENS setup is not required. Instead, a user can provide the sender with a public key, transaction hash, or address and use that.

To withdraw funds with this approach, a user simply enters their private key (as they currently can) and we hide this feature behind a toggle

mds1:
Closes #15

Updated the account setup process to walk through the steps outlined in #15. The withdraw process is also implemented, but without steps 10–16 where we scan the blockchain for announcements.

Before testing, you'll need to remove the `vnd.umbra-v0-signature` record from your Ropsten ENS account. You can do this at https://app.ens.domains/name/yourDomain.eth

Note that this PR only covers the initial implementation, and I want to take another pass at the code for security reasons (e.g. minimizing the time unencrypted data is stored in state)

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -4318,6 +4318,14 @@
       "resolved": "https://registry.npmjs.org/brorand/-/brorand-1.1.0.tgz",
       "integrity": "sha1-EsJe/kCkXjwyPrhnWgoM5XsiNx8="
     },
+    "browser-passworder": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/browser-passworder/-/browser-passworder-2.0.3.tgz",
+      "integrity": "sha1-b90gguUWoXbtvLPc7gt/n85PeRc=",
+      "requires": {
+        "browserify-unibabel": "^3.0.0"
+      }
+    },
     "browserify-aes": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/browserify-aes/-/browserify-aes-1.2.0.tgz",
@@ -4388,6 +4396,11 @@
         "parse-asn1": "^5.1.5",
         "readable-stream": "^3.6.0"
       }
+    },
+    "browserify-unibabel": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/browserify-unibabel/-/browserify-unibabel-3.0.0.tgz",
+      "integrity": "sha1-WmuPD3BM44jTkn30czfiWDD3Hdo="
     },
     "browserify-zlib": {
       "version": "0.2.0",
--- a/app/package.json
+++ b/app/package.json
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "@quasar/extras": "^1.0.0",
+    "browser-passworder": "^2.0.3",
     "ethers": "^5.0.0-beta.185",
     "quasar": "^1.0.0"
   },
--- a/app/src/components/AccountSetupChoosePassword.vue
+++ b/app/src/components/AccountSetupChoosePassword.vue
@@ -1,0 +1,53 @@
+<template>
+  <div>
+    To maximize privacy and security, we'll generate a random private key that
+    will be used to manage your Umbra account. This key will be encrypted based
+    on your chosen password.
+
+    <div
+      class="q-mt-lg"
+      style="max-width: 350px"
+    >
+      <base-input
+        v-model="password1"
+        label="Enter Password"
+        :rules="isValidPassword1"
+        type="password"
+      />
+      <base-input
+        v-model="password2"
+        label="Confirm Password"
+        :lazy-rules="false"
+        :rules="isValidPassword2"
+        type="password"
+      />
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'AccountSetupChoosePassword',
+
+  data() {
+    return {
+      password1: undefined,
+      password2: undefined,
+    };
+  },
+
+  methods: {
+    isValidPassword1() {
+      if (!this.password1) return 'Please enter a password';
+      return this.password1.length > 7 || 'Password must be at least 8 characters';
+    },
+    isValidPassword2() {
+      if (!this.password2) return 'Passwords do not match';
+      const isValid = this.password1 === this.password2;
+      // NOTE: Mutation performed here
+      if (isValid) this.$store.commit('user/setPassword', this.password1);
+      return isValid || 'Passwords do not match';
+    },
+  },
+};
+</script>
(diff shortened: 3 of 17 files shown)


mds1:
> When MetaMask was locked (i.e. needed password) clicked login does nothing

Hmm, can you confirm this? Does this happen on all pages or only a certain page? For me it prompts for the password when locked, and we are calling `await window.ethereum.enable()` so I'm not sure why this would happen

> After I finished the setup flow, the "Finish" button is disabled. Is this something still to be implemented? What is supposed to happen on clicking finish?

It should be enabled and take you back to the home screen. It's not enabled because of the last issue you mentioned, so we'll cycle back to this at the end of this comment

> When on the Withdraw page, the little link to toggle to the other form of logging in seems to be backwards. Shouldn't the link right now be for the private key case, since we're already on the enter password case:

Oops, yes just fixed that in 6b27b2a

> Also, I wonder if that little link in the above screen wouldn't look better centered on the UI?

Tried centering it and agree it looks better. Updated in 6b27b2a. Also added 7761c33 to make the two withdraw login options look more consistent

> When I try to login with my password for withdraw checks, it says the password doesn't matched. I cleared local storage and went through again to make sure I hadn't fat fingered anything. I used the password password so I'm 99.9% certain I'm typing it right.

Perhaps this error message needs to be more clear. Your password was correct, and this error message is saying that the public key associated with the private key in local storage does not match the public key recovered from your signature on the ENS address associated with the wallet you connected. 

This check is what I was describing in #15, where this isn't strictly necessary—there's no technical reason you need your associated web3 wallet to be connected to withdraw. But I think it's useful because: 

1. It "synchronizes" the login options, i.e. the same login requirements are needed to setup your account as to withdraw.
2. It seems like a good sanity check in case users want different Umbra accounts in different browsers.
3. It feels a bit more secure, though this may just be security theater and make no real difference. Because if someone has both your password and access to the encrypted data, they can withdraw your funds anyway without the frontend

As for why it happened, my guess is that you didn't clear the existing signature on your ENS domain. Make sure that is cleared, also clear your local storage, then run through the setup process again. While doing so, keep the console open and see if it logs `Using user's web3 wallet` or `Using Umbra-specific key`. This indicates which account is doing the signing. The latter is what it should say, the former is for users who just want to enter the private key directly (this option will prompt you for a signature via MetaMask, and is not fully integrated into the onboarding flow, so should not be seen yet)

apbendi:
> Hmm, can you confirm this? 

Weird! I was unable to reproduce it. Well, good!

> Perhaps this error message needs to be more clear. Your password was correct, and this error message is saying that the public key associated with the private key in local storage does not match the public key recovered from your signature on the ENS address associated with the wallet you connected.

> As for why it happened, my guess is that you didn't clear the existing signature on your ENS domain.

Ahh, OK, this makes sense now. Yes, this happened and for exactly the reason you suspected. So, this is definitely a corner case, and not something we should worry about further before the hackathon. After clearing it out, everything worked fine. Flow is really intuitive and easy!

Moving forward, here are some things we should consider:

* If we detect an ENS record for the account, don't let them generate a new key at all. Put up a message that tells them they already have a public record, and ask them if they'd like to clear it first
* Append the logged in address to the `localStorage` key, so the configuration is on a per-account basis. This would remove the possibility of ambiguity around accounts, and also let the user use Umbra with two different accounts from the same MetaMask wallet

> This check is what I was describing in #15, where this isn't strictly necessary—there's no technical reason you need your associated web3 wallet to be connected to withdraw. But I think it's useful because

Agree with your reasons— you're right, it doesn't add security, but it is a good sanity check.

PR #24, opened 2020-05-22
Finish frontend

mds1:
~~Still a WIP so don't merge yet, but it's at a point were you can start reviewing it~~. What to test so far:

- Sending funds to an ENS address
    - You can send Dai or ETH, either to myself at `msolomon.eth` or yourself
    - I've sent you some Ropsten Dai so you don't have to mint it
- Withdrawing funds
    - You should be able to view a table showing all funds sent to you
    - You'll only be able to withdraw ETH currently
    - Once ETH withdrawal is complete, the UI will not yet mark it as withdrawn. Instead, go to Etherscan and check the address you withdrew to. Then view the balance of the `from` account and it should be zero

Left to do:
- [x] Implement token withdrawals using GSN
- [x] Have withdrawal table show which funds have been withdrawn.
- [x] Find the easiest way to send all ETH without leaving dust. Currently we hardcode a a gas limit of 21000 and compute the value to send as `value = ethBalance - 21000*gasPrice`. (Update: Confirmed this is the easiest way)

Is there anything left do that I'm missing?


mds1:
This is ready to review and merge. Also pushed the current state of this branch to https://ropsten.umbra.cash.

Once merged, I'll deploy the `umbra-js` library to npm and update this to use the npm version of the package instead of the local one

PR #27, opened 2020-05-25
Prevent stealth address reuse & token balance overwriting

apbendi:


mds1:
LGTM. And VSCode had no lint complaints! 🥳

PR #35, opened 2020-07-03
Fix tests

mds1:
Currently pulling down the repo and running `npm install && npm run test` in the contracts folder fails. Looks like something just got messed up with `package-lock.json`, so this should fix it.

Recommend the following process to test:
1. Clone fresh copy of the repo
2. Switch to `fix-tests` branch
3. Run:

```
cd contracts
npm install
npm run test
```

All tests should pass

### DIFF
--- a/contracts/package-lock.json
+++ b/contracts/package-lock.json
@@ -13,6 +13,19 @@
         "@babel/highlight": "^7.8.3"
       }
     },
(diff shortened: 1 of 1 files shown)


apbendi:
I'm an idiot, and was running this from the wrong directory. Everything seems to be working.

Issue #13, opened 2020-05-12
Remove bytecode functionality from umbra-js and export encryption prefix


mds1:
Once we confirm the new approach to meta-transactions works as planned we can remove the functionality of associating bytecode with an ENS address.

Additionally, the encryption prefix is currently hidden in the KeyPair class. This should be defined elsewhere and exported.

To do:
- [ ] Remove bytecode functionality
- [ ] Export prefix 
- [ ] Clean up all export of constants to ensure consistency

apbendi:
I believe you ripped this out @mds1, or am I mistaken?

mds1:
The bytecode functionality is still in there, so we should keep this issue open for now

mds1:
This is now outdated, so closing this issue

PR #38, opened 2020-07-22
Implement Unstoppable CNS support

mvlabat:
This PR adds support of Unstoppable domains.

Its implementation resides in the `./utils/cns.js` module. It's got the same set of functions as the ens counterpart, but their signatures have slight differences: while ens functions expect web3 provider, cns ones accept `Resolution` instance of our [resolution library](https://github.com/unstoppabledomains/resolution).

While having a separate module is useful for users that might target only CNS domains, such structure can be inconvenient to users that want to support both ENS and CNS for stealth payments. For this purpose I introduced another class `DomainService`, which basically wraps both ens and cns utils into an universal interface. Storing both `provider` and `Resolution` instances in `DomainService` and deciding which method to call based on domain extension makes it possible. It also proved to be quite useful in #39.

There's a lot of space for improvements and possibly open questions. These are some off the top of my head:
- As of now, there are two different code paths for calculating domain namehash. There's actually no need to have them both, so if you'll like the idea to rely on our resolution library, we could refactor even more
- As this PR proposes a new class that stores provider instances, the idea could be extended to storing contract instances

Also one of the topics I'd like to bring up is `umbraKeySignature` constant, which now equals `crypto.ETH.signature-v0` for CNS. Ideally we'd like to have the key to be `crypto.ETH.signature` to just make it consistent with all our conventional records (like `crypto.ETH.address`, `crypto.ZIL.address`, etc) and we'd like to assume that the signature format won't change. This is though inconsistent with `vnd.umbra-v0-bytecode` for ENS, and I'm not sure how bad (or not) it is.


mds1:
Awesome, this looks really great!

Two pieces of feedback before merging:
- Can you add some tests for this CNS support? You can just create a file called `cns.test.js` and may want to base them off the tests in `ens.test.js` (ignore the bytecode functionality test since that is outdated and will be removed anyway in #13). We run the ENS tests against a fork of Ropsten that has an ENS domain configured for testing—if CNS is deployed on Ropsten it may be easiest to test that way. (There's one ENS test that is not properly implemented that we need to update, just so you know)
- Regarding the duplicate `umbraMessage` definition, what if we rename `addresses.json` to `constants.json` and store the message there, and then import that variable?

As for the other topics you mentioned:
- Regarding namehash code paths: Using the resolution library for both seems reasonable, I just want to check out that package a bit more closely first. I'll create a new issue to track this
- Regarding a class for storing contract instance: Can you expand on what you mean here? I'm not sure I follow. Are you referring to a class that does something like this:
```javascript
// Right now this is how we get the CNS contract
const cnsRegistry = createContract(CNS_REGISTRY, cnsRegistryAbi, provider);
// And this is how we get the CNS contract
const publicResolver = createContract(ENS_PUBLIC_RESOLVER, publicResolverAbi, provider);

// Are you referring to a class that would let you do something like this, to abstract away the
// the above two lines?
const domainResolverContract = getResolverContract(provider, <input 'ens' or 'cns'>)
```

- Regarding the `umbraKeySignature`, I think we can leave this as is and I'll create a separate issue to discuss this further. There's a few different considerations here that I'll mention when opening the issue. As for the `vnd.umbra-v0-bytecode` record, that can be ignored as it's not used and will be removed as mentioned above.

mvlabat:
Thank you for you review @mds1, glad to hear it!

Regarding the feedback:
- Totally missed the `test` folder, what was a shameful oversight. :D Added the tests for `cns`. I think ideally we should also cover the new `DomainService` class, but practically that would be like concatenating `ens` and `cns` tests into one more file. Though maybe it's better than having no test coverage, and adding them is not a big deal, so let me know if you want to have that module tested.
- I like the idea, done

> Regarding a class for storing contract instance: Can you expand on what you mean here? I'm not sure I follow. Are you referring to a class that does something like this

I was referring to `DomainService` class. I think we might want to store contract instances there and possibly avoid recreating them in every function. Though this would probably lead to heavy refactoring of ens, cns modules

> Regarding the `umbraKeySignature`, I think we can leave this as is and I'll create a separate issue to discuss this further. There's a few different considerations here that I'll mention when opening the issue

Alright, sounds good

mds1:
> Totally missed the test folder, what a shameful oversight. :D Added the tests for cns. I think ideally we should also cover the new DomainService class, but practically that would be like concatenating ens and cns tests into one more file. Though maybe it's better than having no test coverage, and adding them is not a big deal, so let me know if you want to have that module tested.

Certainly can't hurt to have them, so I'll take you up on that offer 🙂

After that we should be good to merge!

> I was referring to DomainService class. I think we might want to store contract instances there and possibly avoid recreating them in every function. Though this would probably lead to heavy refactoring of ens, cns modules

Ah, I see, I like that idea. We'll likely refactor things a bit when we upgrade the library to Typescript (#49), so I'll make an issue for that and perhaps a good time to add in this change would be during that TS update

mvlabat:
Updated the branch with tests for `DomainService`

PR #39, opened 2020-07-22
Draft UI supporting UD and absence of reverse records

mvlabat:
This PR is based on #38 

I'm sharing my changes I've made to the app to test #38 PR. It's far from being polished (especially the UI), and probably you imagine the UX differently (this is why I didn't dwell into it too much), so I'm creating this as a draft PR.

One of the significant changes is making the demo app work without reverse records, suggesting user to enter their domain manually (in case the domain can't be found automatically). Since Unstoppable Domains doesn't have the concept of reverse resolution, this feature was needed to make the app work with our domains.

For supporting both ENS and CNS resolution, the proposed `DomainService` from the core library is actively used.
> While having a separate module is useful for users that target only CNS domains, such structure might be inconvenient for users that want to support both ENS and CNS for stealth payments. For this purpose I introduced another class `DomainService`, which basically wraps both ens and cns utils into an universal interface. Storing both `provider` and `Resolution` instances in `DomainService` and deciding which method to call based on domain extension makes it possible.

P.S. I don't know what's wrong with `app/package-lock.json` changes, probably it's caused by changing umbra-js dependency to
```
"umbra-js": "file:../umbra-js",
```
I guess we could change it to `0.0.2`, once a new version of the core library releases.


mds1:
Closing since this PR is outdated—the `app` folder no longer exists and replaced by a rewritten frontend in the `frontend` folder

Issue #31, opened 2020-05-29
Improve Encryption Process


mds1:
### Problem

The current encryption process is as follows:
1. Generate a random 32-byte number
2. Using an ephemeral keypair, sender computes shared secret with the receiver via ECDH
3. Use the shared secret to encrypt the random number with AES-256-CBC, which outputs the MAC, IV, and ciphertext
4. Announce the ciphertext, MAC, IV, and ephemeral public key on chain through an event

Note that steps 2 and 3 are handled by the current library we are using for encryption, [eccrypto](https://github.com/bitchan/eccrypto).

The current encryption process is suboptimal for a few reasons:
- IV is logged with the event, but this is not necessary since the payload to encrypt and the encryption key are always random
- MAC is logged with the event, but this is not necessary since the blockchain provides protects data integrity and authenticity

These are currently required by the eccrypto library, and results in unnecessary computation and gas costs.

### Potential solutions

1. Generate random number with the existing process, but use the WebCrypto API to encrypt it with AES-256 using the ECDH shared secret. Would need to decide which AES-256 mode of operation to use (currently using CBC which is probably sufficient). On chain we'd announce the ephemeral public key and encrypted random number
2. Generate random number with the existing process, but encrypt it with XOR. On chain we'd announce the ephemeral public key and encrypted random number. Can also use ECB mode of operation
3. Used the shared secret from ECDH as the random number. On chain we'd only need to announce the ephemeral public key. 

When choosing a solution, factors to consider include:
1. Security
2. Computation and scanning cost 
3. Compatibility with #30 

### References

- [Tweet pointing out this issue](https://twitter.com/julianor/status/1266441668913442823)
- [Tweet suggesting XOR and ECB](https://twitter.com/julianor/status/1266456414391803909?s=20)
- [Tweet suggesting ECDH secret](https://twitter.com/nicksdjohnson/status/1266464073576660992?s=20)
- [bips/bip-0047.mediawiki at master · bitcoin/bips](https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki#notification-transaction)

apbendi:
Let's say we used XOR + ECB, but also had a set prefix at the beginning of our plaintext. Doesn't this weaken our security?

seresistvanandras:
[...]

Issue #30, opened 2020-05-29
Add Payload Extension To Encrypted Announcement


apbendi:
### Proposal

Require a standard payload of data be included, at a standard position, in the message that is encrypted and announced on chain. The size of the payload should be such that it result in exactly 32 additional bytes being added to the ciphertext, requiring one additional `bytes32` parameter included in the contract methods & `Announcement` event.

### But Why?

Flexibility for developers building on top of the protocol. It would enable them to identify certain stealth payments as pertaining to their application, and encode some additional metadata about the payment.

Example 1: A developer builds a Tornado-cash like mixer with Umbra. By placing an identifier in payload extension, they can filter stealth payments for a given user to only those relevant for the mixer application.

Example 2: A developer builds a crypto Invoicing application using Umbra, embedding a unique foreign key identifier of the invoice being paid in the payload extension.

mds1:
One consideration here is that this will reduce the size of the anonymity set. For Example 1 this is inevitable since payment amounts would be restricted to a fixed set of values, but this doesn't have to be happen for the app in Example 2.

You'd probably want to set a default payload value that Umbra will use, and let application developers optionally override it with their own app-specific value (and of course make sure they are aware of the tradeoffs)

apbendi:
> One consideration here is that this will reduce the size of the anonymity set.

How so? To clarify, my thinking is this would be concatenated with the secret **prior** to encryption.

mds1:
Ah I misunderstood. I thought you meant the developer would choose an extra parameter to directly emit (unencrypted) with the Announcement event, with part of the intent being to reduce the scan time for their users.

mds1:
Closed by #61 and #58

PR #58, opened 2020-09-25
Update umbra-js encryption

### LINKED ISSUE #41: Specify size of random number to generate
To compute a stealth address for the recipient, the sender must generate a random number. Currently a 32-byte random number is used.

As part of the spec, determine if we should stick with a 32-byte random number or use a different size.

mds1:
Closes #41, #30, #31

Changes:
- We now generate 16 byte random numbers, stored in a 32 byte value. The first 16 bytes can be optionally used as a payload extension, as described in #41 and #30
- Encryption is now done using XOR as described in #31
- Switch from npm to yarn and setup prettier


umbra-js/classes/RandomNumber.js
```diff
@@ -1,40 +1,53 @@
...
    */
-  constructor(length = 32) {
-    this.length = length;
-    this.value = utils.shuffled(utils.randomBytes(length));
-  }
+  constructor(payloadExtension = zeroPrefix) {
+    // Generate default random number without payload extension
+    this.randomLength = 16; // 16 byte random number
+    this.fullLength = 32; // Pad random number to 32 bytes
+    const randomNumberAsBytes = utils.shuffled(utils.randomBytes(this.randomLength));
```
apbendi, line 21: Not a change, just wondering, what does the `shuffled` function do here?

apbendi (approved):
Looks great, tests pass. Nice work! One small question that is unrelated to the change, but I'm just curious. Good to merge :)

umbra-js/classes/RandomNumber.js
mds1: It just shuffles the random bytes array with a Fisher–Yates shuffle. I figured I'd add that in to get a bit more randomness, but in reality it probably doesn't make much difference

apbendi: Gotcha! I'd say that doesn't hurt, but doesn't really help either. Either the bytes coming out of `randomBytes` are secure randomness as is, or they're not. No need to change, though. Should be good to merge 👍

PR #59, opened 2020-09-25
Update To OpenGSNV2 Beta 1.3

apbendi:
Updates the Umbra contracts and tests for compatibility with OpenGSNV2 Beta 1.3.  Also migrates from `npm` to `yarn`.

Run `yarn install` and `yarn test` to ensure everything functions.


mds1:
Why do we use `@openzeppelin/contracts-ethereum-package` and not `@openzeppelin/contracts`? The former is for developing upgradeable contracts, the latter is the same set of contracts but not upgradeable. OpenGSN uses the latter, so this is likely where the conflict comes in. If we switch over to `@openzeppelin/contracts` I think we'd solve the problem without needing to change their contracts.

contracts/.openzeppelin/project.json
```diff
@@ -9,16 +9,18 @@
   "version": "0.0.1",
   "compiler": {
     "compilerSettings": {
+      "evmVersion": "berlin",
```
mds1, line 12: Where was this set to `berlin`? I don't see that setting defined.

Also, why `berlin`? The [Solidity docs](https://solidity.readthedocs.io/en/v0.6.12/using-the-compiler.html#target-options) say it's experimental so I'm not sure if that's a good idea. Is there any reason not to just use the default of `petersburg`?

contracts/README.md
```diff
@@ -9,14 +9,14 @@ On chain components of the [Umbra protocol](../README.md).
...
+* yarn 1.22.0 or later
 
 ### Usage
 
-From the current working directory, run `npm install`. Afterwards run:
+From the current working directory, run `yarn install`. Afterwards run:
 
-* `npm run compile` — To compile all contracts
-* `npm run test` — To run all tests
+* `yarn compile` — To compile all contracts
```
mds1, line 18: I'd recommend updating the `compile` script to specify a solc version and the evm-target, just because it's good to be explicit and it's not clear how the above settings were selected

contracts/contracts/UmbraPaymaster.sol
```diff
@@ -11,29 +13,34 @@ contract UmbraPaymaster is BasePaymaster {
...
+    )
+    external
+    override
+    returns (bytes memory context, bool rejectOnRecipientRevert) {
 
-    return abi.encode(0x0);
-  }
+      require(relayRequest.request.to == umbraAddr, "UmbraPaymaster: Not Target");
+      return (abi.encode(0x0), true);
+    }
```
mds1, line 32: These comments are to silence compiler warnings about unused variables? Cool that this works, didn't know you could do that! Just found the section of the docs [here](https://solidity.readthedocs.io/en/v0.6.12/control-structures.html?highlight=unused#omitted-function-parameter-names)

May as well mention an alternative way of silencing the warnings:

```solidity
  function preRelayedCall(
    GsnTypes.RelayRequest calldata relayRequest,
    bytes calldata signature,
    bytes calldata approvalData,
    uint256 maxPossibleGas
  ) external override returns (bytes memory context, bool rejectOnRecipientRevert) {
    signature; // silence compiler warning by making variable look used
    approvalData; // silence compiler warning by making variable look used
    maxPossibleGas; // silence compiler warning by making variable look used

    require(relayRequest.request.to == umbraAddr, "UmbraPaymaster: Not Target");
    return (abi.encode(0x0), true);
  }

```

mds1 (changes requested):
Looks great, and all tests pass! One main comment about specifying the compiler version and EVM target but that's it

contracts/README.md
apbendi: we've got this in the OZ config file:

https://github.com/ScopeLift/umbra-protocol/pull/59/files#diff-eeaaa081c313f087ef845b5544746432R22

contracts/.openzeppelin/project.json
apbendi: OK good call on this. I had bumped it up because GSN using `chainid` which is only available in `istanbul` or greater, but I was wrong to jump straight to Berlin. Fixed this.

contracts/contracts/UmbraPaymaster.sol
apbendi: Yeah I actually don't love this compiler warning— I don't mind having the named parameters there as it's more explicit. I updated to use the same approach (the one you mentioned) in both method calls, since I was mixing them.

mds1 (approved):

Issue #29, opened 2020-05-29
Separate Keypairs Used For Stealth Address Generation Vs. Announcement Encryption


apbendi:
### Proposal

Using either two ENS Text records, or one longer ENS Text record that represents two concatenated messages, give users the ability to (optionally?) publish TWO public keys representing two distinct keypairs:

* `PKa` - Public Key A, used by the sender for generating a stealth address for the receiver
* `pka` - Private Key A, used by the receiver to generate the stealth addresses public key for withdrawal
* `PKb` - Public Key B, used by the sender for encrypting the announcement messages published on chain using ECDH
* `pkb` - Private Key B, used by the receiver for decrypting announcements published on chain to look for payments to stealth addresses they control

#### Reason 1: Safety

A receiver may choose to run a service that automatically scans announcements and alerts them to any payments to steal addresses they control. This would require keeping a private key "hot" in a service exposed to the internet.

By separating `PKa` and `PKb`, a scanning service would only need to key `pkb` "hot", and expose `pka` only when making withdrawals.

If the user's scanning service was hacked, or `pkb` was otherwise compromised, the user loses privacy, but not their funds.

#### Reason 2: Delegation

A receiver may choose to delegate the service described in `Reason 1` to a trusted third party. Said party would know the identity of receiver, but never have the ability to confiscate the user's funds.

Why would a user who values privacy enough to use Umbra be willing to delegate this service, and thus disclose their supposedly private transactions to a third party? There are many usecases where they wouldn't, but there may be some where they would.

Example: a crypto invoicing SaaS integrated with Umbra. As a business owner, I'm generally comfortable with my invoicing software provider knowing who I'm invoicing and in what amounts, however I'd be decidedly less comfortable with that information being completely public knowledge.

An invoicing service built with Umbra could retain `pkb` for their customers, scan all announcements on behalf of their users, and send email notifications when stealth payments are received. At the same time, the user remains in control of their funds, and the wider world would have no knowledge of the revenue the business was generating, or from whom.

mds1:
Support for this was added in #60. Leaving this open as `protocol-WIP` until we finish using those keys in the rest of the send/receive flow

mds1:
#90 implements this in the `Umbra` helper class of `umbra-js`. The last place to finish implementing this is the [new ENS resolver contract](https://github.com/ensdomains/resolvers/pull/40), so I'll leave this issue open until that aspect is implemented into this repo

mds1:
[...]

PR #60, opened 2020-10-07
Update private key management

### LINKED ISSUE #44: Specify key management approach
Currently, a new private key is randomly generated when a user sets up their Umbra account. This private key is encrypted with a password and saved in local storage, and the user is asked to back up the key by downloading a file of the encrypted key.

A more robust approach is to use a private key that is deterministically generated from a signature of the user's wallet. Loopring does something similar and describes their approach [here](https://medium.com/loopring-protocol/looprings-new-approach-to-generating-layer-2-account-keys-4a16cc334906).

Quoting from the article:

> We have implemented a new approach to generating and managing account keys on the client-side to achieve the following objectives:
> - The web app shall allow users to change the account keys for the same Ethereum addresses, so they can export and share their account keys to third parties (such as market makers) and later revoke the account keys. This means the mapping between Ethereum addresses and account keys has to be one-to-many, not one-to-one.
> - The seed for generating an account key shall contain the entropy of a valid ECDSA signature from the corresponding Ethereum address.
> - When a browser tab is open for the exchange web app, we shall allow users to refresh/reload the web page without re-login or re-signing any messages using the corresponding Ethereum addresses.
> - If the user closes the tab opened for the app, the session should end immediately, and the user will have to log in again.
If the user explicitly logs out, all account-key related session/local data shall be cleaned.
> - No sensitive data shall be sent to the relayer.

We should aim to meet similar objectives.

mds1:
Closes #44 and supports #29 

The `refactor-frontend` branch is a long running branch to rebuild the frontend with the Composition API and Typescript, and will continually be rebased against master. The `app` folder contains the frontend code that exists on master and will ultimately be deleted. The `frontend` folder contains the new frontend code that is being built on the `refactor-frontend` branch. The frontend in the new `frontend` folder is not yet complete and functionality will be added to it as needed.

In this PR we make the following changes to the `frontend` folder
- Add send and receive pages
- Setup linting/formatting on commit (also makes sure app builds on each commit)
- Remove Login button from header. Instead the user is now prompted to connect wallet and sign any required messages when choosing either "Send" or "Receive" from the home page
- Add route guards so you cannot navigate directly to the send or receive pages. You must always start at the home page. This is the simplest solution for now to ensure a user always connects a wallet and signs any required messages
- Visiting the Send page uses the logic/store in `keys.ts` to prompt the user for their signature and save their private keys in memory

To test:
- Checkout this branch
- Run `yarn` to install packages
- Run `yarn dev` to start the app
- Visit the home page, click "Send", sign the message, and make sure both key pairs are logged to the console


apbendi (changes requested):
Running locally, I get the following error in my console when clicking "Send":

```
webpack-internal:///./node_modules/bnc-onboard/dist/esm/wallet-connect-a477efe1.js:20 Uncaught (in promise) Error: A "infuraKey" or a "rpc" object with a parameter of 3 must be included in the WalletConnect initialization object
    at Module.walletConnect [as default] (webpack-internal:///./node_modules/bnc-onboard/dist/esm/wallet-connect-a477efe1.js:20)
    at eval (webpack-internal:///./node_modules/bnc-onboard/dist/esm/onboard-0e872dae.js:7751)
    at async Promise.all (:8080/index 8)
```

I'm guessing this is an env var I need to configure. Is that right? If so, can we add `.env-template` file with instructions in the README. Wasn't immediately obvious from grepping the codebase what I needed to add.

mds1:
Good call, just added that in 1133dc2

apbendi:
All works well with the updated .env vars. Nice work 👍

PR #61, opened 2020-10-09
Modify the contracts to support new encryption scheme

apbendi:
* Remove unneeded parameters from the send method signatures and
  the emitted event
* Update tests to reflect the new structure

This is the contract side of #31


mds1:
Love how this is so much cleaner and more concise than the old encryption scheme. Looks great!

PR #63, opened 2020-11-20
Convert umbra-js library to Typescript

### LINKED ISSUE #49: Refactor library to use Typescript

mds1:
Closes #49


umbra-js/src/utils/utils.ts
```diff
@@ -15,40 +24,42 @@ const constants = require('../constants.json');
...
   // Get transaction data
-  const ethersProvider = new ethers.providers.Web3Provider(provider);
+  const ethersProvider = new Web3Provider(provider);
   const tx = await ethersProvider.getTransaction(txHash);
 
   // Get original signature
-  const splitSignature = {
-    r: tx.r,
+  const splitSignature: SignatureLike = {
+    r: tx.r as string,
```
apbendi, line 58: Why does this have to be cast as a string here? Doesn't Ethers have full type definitions? I tried looking up the `getTransaction` method in the Ethers doc, but couldn't find it. Weird.

umbra-js/test/DomainService.test.ts
```diff
@@ -1,22 +1,26 @@
...
 
+const { CNS_REGISTRY } = constants;
 const { expect } = chai;
 
 const resolution = new Resolution({
   blockchain: {
     cns: {
-      provider: Eip1993Factories.fromWeb3Version0Provider(provider),
+      provider: Eip1993Factories.fromWeb3Version0Provider(
+        (provider as unknown) as Web3Version0Provider
```
apbendi, line 17: This looks a little messy, can you explain why the multiple casts are needed here?

umbra-js/test/DomainService.test.ts
```diff
@@ -1,22 +1,26 @@
...
+      ),
       registry: CNS_REGISTRY,
     },
   },
 });
-const domainService = new DomainService(provider, resolution);
+const domainService = new DomainService((provider as unknown) as ExternalProvider, resolution);
```
apbendi, line 23: Same questions as with the provider above, what's going on here?

apbendi (approved):
Looks good overall! Great cleanup and much nicer with the types; tests past.

In addition to a few small comments/questions, I wanted to mention this: in a few places, we're using  `ethers.js` types in our public interface. I'm wondering the degree to which that locks us into using ethers moving forward. Maybe that's not a problem, or is even preferable, but I just wanted to mention it. In other languages, like Swift, I'd generally shy away from using a sub-dependency's types in the public interface of a package, unless that package was explicitly an extension of the subdependency or something like that. Maybe @wildmolasses can weigh in on what's typical in the TypeScript world?

Another question is on the tests. Doesn't have to be part of this PR, but maybe we can update the tests so the "number of runs" defaults to a lower number, but can be bumped up  with a different yarn command or something? Would be nice if the test suite defaulted to fast, but we also had the option to do more robust runs when desired.

On a less important note, I see a pptx file being included. Why is that here?

Anyway I don't see any of this as a blocker for getting this PR in, mostly just points of discussion. Feel free to merge.

umbra-js/test/DomainService.test.ts
mds1: The providers from OpenZeppelin test environment , Unstoppable Domain's resolution library, and ethers are different enough that TS complains when casting from one to other, so to get around that complaint—since we know the providers all work between the different packages—we cast to unknown first

umbra-js/test/DomainService.test.ts
mds1: Same as above

umbra-js/src/utils/utils.ts
mds1: `tx` is of type [`TransactionResponse`](https://github.com/ethers-io/ethers.js/blob/efeeda5e5c3b22f0c211783a635941dc8035ae25/packages/abstract-provider/src.ts/index.ts#L31-L49), which extends [`Transaction`](https://github.com/ethers-io/ethers.js/blob/efeeda5e5c3b22f0c211783a635941dc8035ae25/packages/transactions/src.ts/index.ts#L31-L48).  The `getTransaction` method is [here](https://github.com/ethers-io/ethers.js/blob/8933467c01b64ead547d7c136f22f3c05c85ca1f/packages/providers/src.ts/base-provider.ts#L1183-L1215) (but like you said, for some reason I can't find it in the docs!)

So as you can see the `r,s,v` components of `Transaction` don't necessarily exist (since that is used to send transactions also), and since `TransactionResponse` extends that, TS thinks the signature components may be undefined we cast to a string. The other two don't need to be casted because they're optional in `SignatureLike`

However, this is a good point. I think the `r,s,v` components should always exist in `TransactionResponse`, so I'll create an issue in ethers to ask about this

mds1:
@apbendi If we remove ethers' types from the interface, here's the main questions to answer / things to consider. There's other parts we users ethers types, but the ones here are the (developer) facing ones:
1. What type do we request when we need a `provider` input? Various libraries all seem to have their own definitions here
2. In a separate PR we are adding an `Umbra` class so the full send/scan/receive functionality is part of this SDK. We wanted to expose the Umbra contract instance for convenience to the user/developer. Instead of exposing an ethers contract instance, would we now just expose an address and ABI?
3. ~~We're also dependent on a `Resolution` type from `@unstoppabledomains/resolution`, which is used for resolving CNS names~~ Added commit f30021b, so we now generate this instance in the constructor

> On a less important note, I see a pptx file being included. Why is that here?

Removed in f30021b. Those were for the recent blog post and I accidentally committed them here.

> Another question is on the tests. Doesn't have to be part of this PR, but maybe we can update the tests so the "number of runs" defaults to a lower number, but can be bumped up with a different yarn command or something? Would be nice if the test suite defaulted to fast, but we also had the option to do more robust runs when desired.

I also reduced the default number of test runs from 1000 to 100. We were using 1000 as a lazy fuzzing approach to find edge cases, but none have cropped up in a while so 100 should be more than enough for now

umbra-js/src/utils/utils.ts
wildmolasses: we could do a check above: `if(!tx.r) throw new Error("Invalid signature data");` then we can avoid casting to string here. up to you guys.

wildmolasses:
Just added a few things:

- disabled esModuleInterop and skipLibCheck in tsconfig.json. These are "leaky" options with consequences to package consumers. With these disabled, users can configure their tsconfig as they'd like.
- moved index.js to /src and made it .ts: this was necessary in order to generate index.d.ts which is our main type export file.
- set package.json such that it knows to build before packing and publishing the module; set files property so that it will only pack what's in /build!

wildmolasses (approved):

PR #67, opened 2020-12-04
Configure monorepo

wildmolasses:
The monorepo config is now ready pending your review. From the top level...

```sh
yarn       # will install all dependencies
yarn clean # deletes build artifacts
yarn build # builds umbra-js and frontend
```

we can move from here to implement additional workflow improvements!

Note: I didn't update the README yet. Do you guys have any suggestions?


apbendi (commented):
Hey @wildmolasses, thanks for taking a crack at this! Can't wait to have it. The top level install, build, and clean functions all seem to work great. Thanks!

Some thoughts:

* Should the build/clean commands also compile/clean the contracts directory?
* Is it possible to add a top level `yarn test` command that runs the tests for all components?
* When I cd into the `umbra-js` directory an attempt to `yarn test`, I now get the following error:

```
$ yarn test
yarn run v1.22.10
$ yarn build && mocha -r ts-node/register ./test/*.test.ts --exit --timeout 0
$ tsc --build

TSError: ⨯ Unable to compile TypeScript:
test/cns.test.ts:4:22 - error TS2307: Cannot find module '../build/utils/cns' or its corresponding type declarations.

4 import * as cns from '../build/utils/cns';
                       ~~~~~~~~~~~~~~~~~~~~
test/cns.test.ts:5:28 - error TS2307: Cannot find module '../build/constants.json' or its corresponding type declarations.

5 import * as constants from '../build/constants.json';
```

* When I cd into the `contracts` directory and run `yarn test`, I now get the following error:

```
yarn test
yarn run v1.22.10
$ npx oz compile && mocha --exit --timeout 0 test/*-test.js
✖ Compiling contracts with solc 0.6.12 (commit.27d51765)
Compilation errors:
contracts/Umbra.sol:13:1: TypeError: Derived contract must override function "_msgData". Two or more base classes define function with same name and parameter types.
contract Umbra is BaseRelayRecipient, IKnowForwarderAddress, Ownable {
^ (Relevant source part starts here and spans across multiple lines).
@opengsn/gsn/contracts/BaseRelayRecipient.sol:49:5: Definition in "BaseRelayRecipient":
    function _msgData() internal override virtual view returns (bytes memory ret) {
    ^ (Relevant source part starts here and spans across multiple lines).
@openzeppelin/contracts/GSN/Context.sol:20:5: Definition in "Context":
    function _msgData() internal view virtual returns (bytes memory) {
    ^ (Relevant source part starts here and spans across multiple lines).

```

This error was new, but resolved after a `yarn install` in the contracts directory. Is that expected?

frontend/package.json
```diff
@@ -1,19 +1,21 @@
...
     "test": "echo \"No test specified\" && exit 0",
-    "dev": "npx quasar dev",
-    "build": "npx quasar build"
+    "dev": "quasar dev",
+    "build": "quasar build",
+    "clean": "rimraf build"
   },
   "dependencies": {
     "@quasar/extras": "^1.0.0",
+    "@umbra/umbra-js": "^0.0.1",
```
mds1, line 18: How does it know to get this package from the `umbra-js` folder instead of npm? Is that from `lerna.json`?

package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "umbra",
+  "version": "0.0.1",
+  "description": "Send and receive stealth payments with the Umbra protocol",
+  "author": "mds <17163988+mds1@users.noreply.github.com>",
```
mds1, line 5: @apbendi Should we update the author in all `package.json` files to something like this? 

```suggestion
  "author": "ScopeLift <contact@scopelift.co>",
```

mds1 (changes requested):
I first ran `yarn` from the root, then was able to run `yarn test` in the `umbra-js` folder without any issues. However, I get the same error as Ben when running tests in the `contracts` folder even after running `yarn` in that directory.

As for the README, we should update the `Development` section to explain / answer:
- The setup of this monorepo and how it works
- How to get setup for development, e.g. all the steps for installing packages, cleaning, building, linting, running prettier, running tests, deploying, etc.
    - Are these things I can do all from the root, or do I need to `cd` into folders at times?
    - Does `yarn clean` remove build artifacts from all 3 folders? What if I only want to remove them from one directory? Same question for `yarn build`
    - For deploying, this might still be a WIP, in which case we can leave a placeholder for it. I just mentioned it since you previously asked about a process for publishing `umbra-js`

Aside from the above, looks great!

frontend/package.json
wildmolasses: [Yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)!

wildmolasses:
Hi guys, ready for review again!

@apbendi 
> Should the build/clean commands also compile/clean the contracts directory?

Done ✅ 

> Is it possible to add a top level yarn test command that runs the tests for all components?

Done ✅  (Can you give me feedback on this? Currently, the tests run in parallel, sacrificing legibility for speed)

> When I cd into the umbra-js directory an attempt to yarn test, I now get the following error...

Fixed ✅ 


---

@mds1 

> As for the README, we should update the Development section
Done ✅ 

> For deploying, this might still be a WIP, in which case we can leave a placeholder for it. I just mentioned it since you previously asked about a process for publishing umbra-js

I did take some time to look into this. Specifically I looked into the `yarn lerna version` command. This updates all package versions (does some hand holding on whether the version change is a patch, minor release, major release, etc), automatically makes a new commit (and pushes it 😬 ), and automatically creates a version tag on github. I believe it could also publish the package to npm. But, the tool seemed to assume some things that in our case don't apply. For example, I'm not sure if we can be specific about publishing *only* umbra-js. I'm happy to keep looking into this. Perhaps it could be a follow up issue.

apbendi (approved):
Everything works great. Very nice @wildmolasses. 👍

PR #69, opened 2020-12-16
migrate contracts folder to @paulrberg solidity template. use hardhat.

### LINKED ISSUE #68: Migrate to Hardhat + TypeScript for contracts folder
Recommend using https://github.com/PaulRBerg/solidity-template as a starting point, and making a few changes:
- Update prettier config file to use our standard configuration
- Update husky config file to use our normal lint-staged setup
- Remove commitlint

wildmolasses:
Resolves #68


apbendi:
Hey @wildmolasses, the master branch is finally in good shape, with all tests passing and everything relatively stable. You should be clear to rebase your branch and move forward with this upgrade without @mds1 or me stomping on your changes. If you have any issues, lmk and we can pair through it.

apbendi:
@wildmolasses after cloning this branch, running `cd contracts` and `yarn` and `yarn test`, adding a mnemonic and infura ID to my `.env` as instructed, then re-running `yarn test`, I get the following errors from the test suite:

```
  1) Umbra GSN
       "before all" hook for "should allow someone to pay with a token":
     Error: could not find unlocked account with sufficient balance; all accounts:
 -
      at CommandsLogic.findWealthyAccount (/Users/bendi/Development/umbra/node_modules/@opengsn/gsn/src/cli/CommandsLogic.ts:99:11)

  2) Umbra
       "before all" hook for "should see the deployed Umbra contract":
     TypeError: Cannot set property 'umbra' of undefined
      at /Users/bendi/Development/umbra/contracts/test/umbra-test.js:56:15
      at step (test/umbra-test.js:33:23)
      at Object.next (test/umbra-test.js:14:53)
      at fulfilled (test/umbra-test.js:5:58)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

Any ideas on what's going on here and how to get past these?

wildmolasses:
Typechain turns contract types into typescript types so they can be imported elsewhere. They are now a hard dependency of umbra-js. These type files are generated and as such are not committed to the repo. But users of umbra-js will have to get them from somewhere. On one hand we can make @umbra/contracts a dependency of @umbra/umbra-js. When @umbra/umbra-js is installed with `yarn install`, @umbra/contracts would be pulled down and typechain would run. This is very easy to configure but requires maybe a couple seconds more time for the umbra-js consumer as well as some extra node_modules folders. The other idea is to commit types directly to the repo. This would mean the consumer's machine would not run typechain as part of its `yarn install`. But it would come at the cost of maintainability, as including generated files in a repo can be messy. What are your thoughts?

apbendi:
hey @wildmolasses, I'm not clear if you're discussing users of `umbra-js` who are using this repo, or users of `umbra-js` who are just pulling it from npm. Can you clarify?

wildmolasses:
> hey @wildmolasses, I'm not clear if you're discussing users of `umbra-js` who are using this repo, or users of `umbra-js` who are just pulling it from npm. Can you clarify?

thanks Ben. talking about users of `umbra-js` who are pulling it from npm

apbendi:
Got it thanks @wildmolasses. So you're saying you could set it up such that installing `@umbra/umbra-js` via npm would also pull down `@umbra/umbra-contracts` and build the types from there? Can you easily pin one version of the former to point to a specific version of the latter? Overall seems OK to me to go that route, but curious what @mds1 thinks.

wildmolasses:
> So you're saying you could set it up such that installing `@umbra/umbra-js` via npm would also pull down `@umbra/umbra-contracts` and build the types from there? Can you easily pin one version of the former to point to a specific version of the latter? 

yes and yes! alternatively (and now that i think about it this is probably the "recommended way") we could make the umbra-js npm package the build and not the src! I'll go ahead with my changes but I'm also looking forward to @mds1 input here!

apbendi (approved):
Works great! Tests feel fast— runs smoothly after fresh `yarn install` and `yarn test` from the root directory. Thanks for also updating the README. Really excited to have the Hardhat tooling at our disposal 🎉

PR #70, opened 2021-01-05
Create Umbra class and test file

mds1:
Creates a helper class called `Umbra.ts`. This class is intended to abstract away contract interactions, random number handling, and cryptographic operations from the frontend.

### To test
1. Go to https://ropsten.etherscan.io/ and copy the latest block number
2. Paste the value in the `startBlock` fields of `contractInfo` in line 43 of `Umbra.ts`
3. Run `yarn test`

The above is necessary because if the block is too old, the ganache fork errors with `could not detect network` when scanning for events. Presumably this is because the ganache fork needs to scan for events emitted by transactions on the real Rospten, and our local fork. If we could start ganache with the deterministic flag we could run these tests on localhost, but it seems OpenZeppelin's `test-environment` doesn't support that, so this is the workaround for now. When we switch to hardhat we can run these tests locally instead of against Ropsten

### Notes
- Withdraw functionality and tests are not yet implemented since it's dependent on #72. These can be implemented in a separate PR since #72 requires a redeploy to Ropsten before we can test against it

### Potential Improvements
1. Regarding the `tsconfig.json` and `typechain` setup, not sure if this is optimal for the monorepo. Specifically, when running `yarn build` from the root, it should compile the `contracts` directory first, then build the `umbra-js` files to ensure Typechain is run with the latest contract version. This will be impacted by @wildmolasses updating the contracts folder, so we can leave this as-is for now


apbendi:
hey @mds1, I'm unable to get this to build. At the top level, I run `yarn install` and get the following error:

```
$ lerna run prepare
lerna notice cli v3.22.1
lerna info versioning independent
lerna info Executing command in 1 package: "yarn run prepare"
@umbra/umbra-js: yarn run v1.22.10
@umbra/umbra-js: $ yarn build
@umbra/umbra-js: $ tsc --build && yarn build:types
@umbra/umbra-js: src/classes/Umbra.ts(15,47): error TS2307: Cannot find module '../../types/contracts' or its corresponding type declarations.
@umbra/umbra-js: error Command failed with exit code 1.
@umbra/umbra-js: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
@umbra/umbra-js: error Command failed with exit code 1.
@umbra/umbra-js: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run prepare exited 1 in '@umbra/umbra-js'
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

I tried blowing away `node_modules` to no avail. Any recommendations?

apbendi:
> Go to https://ropsten.etherscan.io/ and copy the latest block number
> Paste the value in the startBlock fields of contractInfo in line 43 of Umbra.ts
> Run yarn test

I know this is intended to be temporary until we get Hardhat working, but I wonder if it isn't a bit of code smell in itself. Instead of having only hardcoded values by chainId, maybe we should enable the user to optionally pass the chain configuration data into the constructor. After all, they could be using a version they deployed, or using an older version of `umbra-js` with a chain we recently deployed to (e.g. Matic or xDai or whatever).

Off the top of my head, not tested to be correct, this could be something like this:


```typescript
type ChainId = number;

interface ChainConfigData {
  chainId: ChainId;
  umbraAddress: string;
  umbraPaymasterAddress: string;
  startBlock: number;
}

type ChainConfig = ChainId | ChainConfigData;

constructor(readonly provider: EthersProvider, chainConfig: ChainConfig, signer: JsonRpcSigner | null) {
  // If chainId is a `number` pull from hardcoded mapping, otherwise use the data provided
}
```

mds1:
The remaining related TODOs have been pulled out into #75, #78, and #80, so I'm going to merge this PR

Issue #78, opened 2021-01-19
Resolve issues related to CNS name resolution using Ethers provider


apbendi:
In order to support CNS via the updated `umbra-js` package, we need the ability to use an ethers provider with Unstoppable Domains resolution library, however this functionality appears to be broken. Issue on the appropriate repo has been opened here:

https://github.com/unstoppabledomains/resolution/issues/112

Once we have some guidance from the  Unstoppable Team, we need to upgrade our package or take whatever other actions are recommended to resolve the issue and fix CNS resolution.

PR #84, opened 2021-01-19
Add ability to customize chainConfig

### LINKED ISSUE #80: Inject ChainId and other config parameters into the Umbra TypeScript class constructor
Dependent on #70 

Instead of having only hardcoded values by chainId, maybe we should enable the user to optionally pass the chain configuration data into the constructor. After all, they could be using a version they deployed, or using an older version of `umbra-js` with a chain we recently deployed to (e.g. Matic or xDai or whatever).

Off the top of my head, not tested to be correct, this could be something like this:

```typescript
type ChainId = number;

interface ChainConfigData {
  chainId: ChainId;
  umbraAddress: string;
  umbraPaymasterAddress: string;
  startBlock: number;
}

type ChainConfig = ChainId | ChainConfigData;

constructor(readonly provider: EthersProvider, chainConfig: ChainConfig, signer: JsonRpcSigner | null) {
  // If chainConfig is a `number` pull from hardcoded mapping, otherwise use the data provided
}
```

This would also allow us to query and inject the appropriate block number, chainid, and other parameters during testing, rather that requiring other workarounds.

mds1:
Closes #80

Defines new type `ChainConfig` that can be provided when creating an instance of the Umbra class. This defines the Umbra contract address and default block to start scanning from. Umbra instances can now be created with the following options:
- Instantiation: (1) `Umbra.create` with a signer to allow calling methods, or (2) `Umbra.createReadonly` with an node URL to only allow reads
- Chain config: (1)  Don't provide a `chainConfig` to infer it from the signer/provider, or (2) provide a chain ID to use the hardcoded info, or (3) enable it to be fully custom

This results in 3 * 2 = 6 possible ways to create an Umbra class

Other notes:
- The Umbra tests now fetch the latest Ropsten block number and always start scanning from there, so you no longer need to manually update that for tests
- I also made a small update to the `scan` method to allow the start and end block numbers to be overridden.
- All tests should pass, with the exception of the 4 failing ones described in #78

### DIFF
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -15,47 +15,44 @@ import { lookupRecipient } from '../utils/utils';
 import { Umbra as UmbraContract, ERC20 } from '../../types/contracts';
 import * as erc20Abi from '../abi/ERC20.json';
 import type {
+  Announcement,
+  UserAnnouncement,
+  ChainConfig,
   EthersProvider,
   UmbraOverrides,
-  // Announcement,
-  // UserAnnouncementEvent,
+  ScanOverrides,
 } from '../types';
-
-// Additional type definitions
-interface ContactInfo {
-  umbraAddress: string;
-  umbraPaymasterAddress: string;
-  startBlock: number;
-}
-
-interface AnnouncementEvent {
-  receiver: string;
-  amount: BigNumber;
-  token: string;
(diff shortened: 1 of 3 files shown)


apbendi (approved):
Code looks good and all tests pass w/o editing any files, except unrelated CNS failures. Nice work on this. I like the `ScanOverrides` optional param as well.

PR #87, opened 2021-01-20
Pin @unstoppabledomains/resolution to v1.9.0

### LINKED ISSUE #78: Resolve issues related to CNS name resolution using Ethers provider

mds1:
Closes #78

Breaking changes were introduced after this in v1.10.1, so we pin the version to ensure a compatible version is always installed. See the links below for more information:
- https://github.com/unstoppabledomains/resolution/issues/112
- https://github.com/unstoppabledomains/resolution/blob/master/CHANGELOG.md

To test:
1. Update the `startBlock` values in `Umbra.ts` with the latest Ropsten block number
2. Install packages at the root with `yarn`,
3. Run `cd umbra-js` and `yarn test`. All tests should pass with 2 that have always been skipped.

### DIFF
--- a/umbra-js/package.json
+++ b/umbra-js/package.json
@@ -21,7 +21,7 @@
   "author": "Matt Solomon <matt@mattsolomon.dev>, Ben DiFrancesco <ben@scopelift.co>",
   "license": "ISC",
   "dependencies": {
-    "@unstoppabledomains/resolution": "^1.16.0",
+    "@unstoppabledomains/resolution": "1.9.0",
     "bn.js": "^5.1.3",
     "buffer": "^6.0.2",
     "dotenv": "^8.2.0",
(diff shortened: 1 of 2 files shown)


apbendi (approved):
Tests pass after `yarn install`.

I've created https://github.com/ScopeLift/umbra-protocol/issues/88 to make sure we circle back to upgrading the package once we've dropped Ropsten support. Please comment if anything seems off with that issue.

PR #90, opened 2021-02-02
Add umbrajs withdraw methods

### LINKED ISSUE #79: Add withdrawal support the umbra-js class
Pending the merger of #72 and #70, we need to add convenience methods for both token withdrawal methods to the Umbra typescript class.

mds1:
Closes #79 

Also refactors methods to take in exactly what is needed. So we no longer associate keys or a signer with the Umbra class—just a provider—and methods take keys or singers as inputs when needed

### DIFF
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -180,7 +180,7 @@ export class Umbra {
 
   /**
    * @notice Withdraw ETH or tokens to a specified destination address with a regular transaction
-   * @param signer Signer to send transaction from
+   * @param generationPrivateKey Receiver's generation private key
    * @param token Address of token to withdraw
    * @param stealthAddress Stealth address funds were sent to
    * @param destination Address to send funds to
@@ -191,14 +191,15 @@ export class Umbra {
    * @dev This method does not relay meta-transactions
    */
   async withdraw(
-    signer: JsonRpcSigner | Wallet,
+    generationPrivateKey: string,
     token: string,
     stealthAddress: string,
     destination: string,
     overrides: Overrides = {}
   ) {
     // Configure signer
-    const txSigner = this.getSigner(signer);
+    const wallet = new Wallet(generationPrivateKey);
+    const txSigner = this.getSigner(wallet);
 
     if (isETH(token)) {
       // Withdraw ETH
@@ -258,10 +259,15 @@ export class Umbra {
 
   /**
    * @notice Scans Umbra event logs for funds sent to the specified address
-   * @param keyPair KeyPair of user to scan for, must have associated private key
+   * @param generationPublicKey Receiver's generation private key
+   * @param encryptionPrivateKey Receiver's encryption public key
    * @param overrides Override the start and end block used for scanning
    */
-  async scan(keyPair: KeyPair, overrides: ScanOverrides = {}) {
+  async scan(
+    generationPublicKey: string,
+    encryptionPrivateKey: string,
+    overrides: ScanOverrides = {}
+  ) {
     // Get start and end blocks to scan events for
     const startBlock = overrides.startBlock || this.chainConfig.startBlock;
     const endBlock = overrides.endBlock || 'latest';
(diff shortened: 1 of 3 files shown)


umbra-js/src/classes/Umbra.ts
```diff
@@ -59,83 +62,59 @@ const isETH = (token: string) => {
...
+   * @param signer Signer that may or may not have an associated provider
    */
-  static async createReadonly(url: string, chainConfig: ChainConfig | number | null = null) {
-    // If chainConfig is provided, use user-provided info
-    const provider = new JsonRpcProvider(url);
-    if (chainConfig) return new Umbra(provider, null, parseChainConfig(chainConfig));
-    // Otherwise, read chain ID from provider
-    const network = await provider.getNetwork();
-    return new Umbra(provider, null, parseChainConfig(network.chainId));
+  getSigner(signer: JsonRpcSigner | Wallet) {
```
apbendi, line 90: I see what this method is doing, but I wonder about the naming. Typically on an object a method starting with `get` would just be a param getter. I was confused by why I'd have to pass a signer into a method called `getSigner`. That said, I'm not sure what the right alternative is. Maybe `connectSigner` or `getConnectedSigner` or something like that?

umbra-js/src/classes/Umbra.ts
```diff
@@ -175,55 +154,123 @@ export class Umbra {
...
   }
 
   /**
-   * @notice Withdraw funds for a recipient
+   * @notice Withdraw ETH or tokens to a specified destination address with a regular transaction
+   * @param generationPrivateKey Receiver's generation private key
    * @param token Address of token to withdraw
-   * @param recipient Recipient address
+   * @param stealthAddress Stealth address funds were sent to
+   * @param destination Address to send funds to
```
apbendi, line 189: Nit: might reword to "Address where funds will be withdrawn to" to avoid confusion with "send" in previous param's description.

umbra-js/src/classes/Umbra.ts
```diff
@@ -175,55 +154,123 @@ export class Umbra {
...
-    //   // Based on gas price, compute how much ETH to transfer to avoid dust
-    //   // TODO
-    // } else {
-    //   // this.umbra.withdrawToken(...);
-    // }
+  /**
+   * @notice Withdraw tokens by sending a meta-transaction on behalf of a user
+   * @param signer Signer to send transaction from
+   * @param stealthAddr Stealth address funds were sent to
+   * @param destination Address to send funds to
```
apbendi, line 231: Same comment as above.

umbra-js/src/classes/Umbra.ts
```diff
@@ -175,55 +154,123 @@ export class Umbra {
...
+   * @param sponsor Address that receives sponsorFee
+   * @param sponsorFee Fee for relayer
+   * @param v v-component of signature
+   * @param r r-component of signature
+   * @param s s-component of signature
+   * @param overrides
```
apbendi, line 237: Even though they're the same as the other method, probably good to enumerate the available overrides, or at least point devs at the other method where they can be referenced.

apbendi (approved):
Looks great! I think the final public API came out really well— pretty clean and clear how to use the class based on the method params and their types, I think. A few comments, but all pretty minor. Good to merge after addressing as you see fit!

umbra-js/src/classes/Umbra.ts
mds1: > Typically on an object a method starting with get would just be a param getter.

In ethers world this doesn't seem to be the case, as there are methods like `getTransactionReceipt()` and `getDefaultProvider()` which don't just return parameters.

I like `getConnectedSigner()` though, let's go with that

mds1: Changed in ba0e6cbc548a145aa90dc54534b6baef8f192ff5

umbra-js/src/classes/Umbra.ts
mds1: Done in ba0e6cbc548a145aa90dc54534b6baef8f192ff5

umbra-js/src/classes/Umbra.ts
mds1: Done in ba0e6cbc548a145aa90dc54534b6baef8f192ff5

umbra-js/src/classes/Umbra.ts
mds1: Agreed, think I just forgot to add it, so done in ba0e6cbc548a145aa90dc54534b6baef8f192ff5

mds1:
@apbendi All comments addressed in the last commit, feel free to merge if it looks good!

PR #91, opened 2021-02-09
Upgrade to OpenGSN v2.1

### LINKED ISSUE #81: Implement custom GSN Forwarder to avoid double signature checks
Right, we're double checking signatures by using the default Forwarder. We can implement our own dummy forwarder which ignores the signature on the call, because we know it will be checked inside the `withdrawTokenOnBehalf` method.

See also #20 to avoid relaying calls to methods other than `withdrawTokenOnBehalf`. Open question: should this happen in the Paymaster or the Forwarder?

### LINKED ISSUE #86: Upgrade to Production Version of OpenGSN Packages
See branch https://github.com/ScopeLift/umbra-protocol/tree/gsn-21-upgrade

After upgrading, GSN tests fail with:

```
-Failed to relay call. Results:
-Relaying errors (1):
 -http://127.0.0.1:64085 => paymaster rejected in local view call to 'relayCall()' : signature mismatch
```

apbendi:
This PR upgrades OpenGSN to the latest mainnet version, along with the tests for OpenGSN support. It also cleans up the `UmbraForwarder` stub to include the absolute minimum set of functionality needed to comply with the GSN architecture without actually doing the validation inside the forwarder contract.

closes #86 
closes #81

### DIFF
--- a/contracts/contracts/open_gsn/UmbraPaymaster.sol
+++ b/contracts/contracts/open_gsn/UmbraPaymaster.sol
@@ -14,7 +14,7 @@ contract UmbraPaymaster is BasePaymaster {
   }
 
   function versionPaymaster() external view override returns (string memory) {
-    return "1.0.0";
+    return "2.0.0";
   }
 
   function preRelayedCall(
@@ -26,7 +26,7 @@ contract UmbraPaymaster is BasePaymaster {
     (signature, approvalData, maxPossibleGas); // to silence compiler warnings
 
     require(relayRequest.request.to == umbraAddr, "UmbraPaymaster: Not Target");
-    return (abi.encode(0x0), true);
+    return ("", true);
   }
 
   function postRelayedCall(
--- a/contracts/test/umbra-gsn-test.js
+++ b/contracts/test/umbra-gsn-test.js
@@ -3,7 +3,6 @@ const { expectEvent, expectRevert } = require('@openzeppelin/test-helpers');
 const { expect } = require('chai');
 const { RelayProvider } = require('@opengsn/gsn/dist/src/relayclient/RelayProvider');
 const { GsnTestEnvironment } = require('@opengsn/gsn/dist/GsnTestEnvironment');
-const { configureGSN } = require('@opengsn/gsn/dist/src/relayclient/GSNConfigurator');
 const { argumentBytes } = require('./sample-data');
 const { sumTokenAmounts, signMetaWithdrawal } = require('./utils');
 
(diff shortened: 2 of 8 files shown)


contracts/contracts/open_gsn/UmbraPaymaster.sol
```diff
@@ -14,7 +14,7 @@ contract UmbraPaymaster is BasePaymaster {
   }
 
   function versionPaymaster() external view override returns (string memory) {
-    return "1.0.0";
+    return "2.0.0";
```
mds1, line 17: Do we actually need to bump this to 2.0.0 given that it's never been used for anything real? It looks like this method is not used in signatures anywhere so it shouldn't affect replay attacks

contracts/contracts/open_gsn/UmbraPaymaster.sol
```diff
@@ -26,7 +26,7 @@ contract UmbraPaymaster is BasePaymaster {
     (signature, approvalData, maxPossibleGas); // to silence compiler warnings
 
     require(relayRequest.request.to == umbraAddr, "UmbraPaymaster: Not Target");
-    return (abi.encode(0x0), true);
+    return ("", true);
```
mds1, line 29: What's the rationale behind this change?

contracts/test/umbra-gsn-test.js
```diff
@@ -39,24 +38,29 @@ describe('Umbra GSN', () => {
...
     ctx.forwarder = deployedForwarder.address;
 
-    // Deploy the Umbra GSN wrapper & paymaster
+    // Deploy the Umbra GSN wrapper
     ctx.relayRecipient = await UmbraRelayRecipient.new(ctx.umbra.address, ctx.forwarder, {
       from: owner,
     });
+
+    // Deploy the Umbra Paymaster
     ctx.paymaster = await UmbraPaymaster.new(ctx.relayRecipient.address, { from: owner });
```
mds1, line 57: Didn't we migrate to hardhat? It looks like we still use a lot of web3 in the tests. Maybe we didn't migrate to ethers-style and just changed the VM?

mds1 (approved):
Test pass, looks good! Just some small comments

contracts/test/umbra-gsn-test.js
wildmolasses: Yes, we are using hardhat, and the hardhat-web3 plugin. Tests could be updated to use ethers, but it wasn't strictly necessary!

contracts/contracts/open_gsn/UmbraPaymaster.sol
apbendi: Good question. I'm not exactly sure what was causing it, but something was throwing an error until I updated this value. I think perhaps their JS library (used in the tests) for interacting with the contracts will throw if the versions don't agree.

contracts/contracts/open_gsn/UmbraPaymaster.sol
apbendi: I just pulled this from their updated Paymaster reference implementation for the mainnet v2.0. Not sure it actually matters.

PR #92, opened 2021-02-12
Create deploy script

### LINKED ISSUE #77: Create A Hardhat Deploy Script
Deploy script spec:

1. Deploy `Umbra` where the three constructor params are configurable via `.env` or other deployment config file
2. Deploy `UmbraForwarder`
3. Deploy `UmbraRelayRecipient` with the `Umbra` address and the `UmbraForwarder` addresses from prior steps as constructor arguments
4. Deploy `UmbraPaymaster` with the `UmbraRelayRecipient` address from the prior step as constructor argument
5. Call the `setRelayHub` [method](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/test/umbra-gsn-test.js#L81) on the `UmbraPaymaster` contract instance deployed in prior step with the address of the [public relayer](https://docs.opengsn.org/contracts/addresses.html#ethereum) for the appropriate network. Since we're planning to deploy to Rinkeby, it looks like the correct address is `0x53C88539C65E0350408a2294C4A85eB3d8ce8789`
6. Fund the `UmbraPaymaster` by sending at least 1 testnet ETH to the address of the contract deployed in step 4. This is a normal ETH [value transfer](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/test/umbra-gsn-test.js#L85)
7. TBD: There may be work related to the [ENS Resolver](https://github.com/ensdomains/resolvers/pull/40) work, but this is not quite ready.

As usual, we want to log all the addresses produced out to the filesystem, along with any errors.

Since we're depending on the existence of a public RelayHub contract, we might want to consider using the [forking feature](https://hardhat.org/guides/mainnet-forking.html) of hardhat network to fork Rinkeby and test during development. Alternatively, we could use the OpenGSN test package [used in the tests](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/test/umbra-gsn-test.js#L5) to deploy a RelayHub to Hardhat network first. The former feels easier to me, but I could be wrong.

We probably won't be using OpenGSN in our official mainnet deployment, though we need it now for testing, so don't spend crazy amounts of time making the script perfect and super configurable. It just have to work well enough to get us to Rinkeby.

jferas:
Closes #77 ... Does every thing described in the steps listed in the issue, but is getting an odd transaction error on step 6... funding the paymaster... it is configurable via a JSON file.


contracts/package.json
```diff
@@ -73,7 +73,8 @@
     "prettier:list-different": "prettier --list-different \"**/*.{js,json,md,sol,ts}\"",
     "test": "concurrently --success first --kill-others \"hardhat node &>/dev/null\" \"hardhat test --network localhost\"",
     "typechain": "hardhat typechain",
-    "prepare": "yarn run typechain"
+    "prepare": "yarn run typechain",
+    "deploy:umbra": "hardhat run scripts/deploy.js"
```
mds1, line 77: Nit, but maybe rename the deploy script? Since there's only one we don't need the `:umbra` suffix. Also good to explicitly compile beforehand
```suggestion
    "deploy": "yarn build && hardhat run scripts/deploy.js"
```

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+
+const RINKEBY_PAYMASTER_PUBLIC_RELAYER_ADDRESS = '0x53C88539C65E0350408a2294C4A85eB3d8ce8789';
+
+const deployParams = require('./deployParams.json');
+
+// Initialize object that will hold all deploy info. We'll continually update this and save it to
+// a file using the save() method below
+const parameters = {
+  // TODO consider changing layout so a contract's constructor parameters live with its address
+  admin: null,
```
mds1, line 22: Seems this is never used so we can delete this

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+  contracts: {}, // will be populated with all contract addresses
+  actions: {}, // will be populated with deployment actins
```
mds1, line 24:
```suggestion
  actions: {}, // will be populated with deployment actions
```

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+ * rather than restart the full deployment
+ * @dev To initialize a deploy:
+ *   `yarn deploy:umbra --network <network>`   (where network specifies a netwrk found in the hardhat.config.ts file)
+ * If deploying to a local node (--network localhost), first start Hardhat using `yarn hardhat node`
+ */
+const fs = require('fs');
+const hre = require('hardhat');
+const { ethers } = hre;
+
+const RINKEBY_PAYMASTER_PUBLIC_RELAYER_ADDRESS = '0x53C88539C65E0350408a2294C4A85eB3d8ce8789';
```
mds1, line 14: I think we should move this into `deployParams.json`, then modify `deployParams.json` as explained in the comment at that file

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+
+const deployParams = require('./deployParams.json');
```
mds1, line 16: For me this has a red squiggly line underneath that says

```
File 'umbra/contracts/scripts/deployParams.json' is not listed within the file list of project 'umbra/contracts/tsconfig.json'. Projects must list all files or use an 'include' pattern.ts(6307)
```

If you add `"scripts/**/*.json",` as an entry to `contracts/tsconfig.json` that should resolve this. For some reason the `"scripts/**/*"` doesn't cover this, even though you'd expect it to

contracts/scripts/deployParams.json
```diff
@@ -0,0 +1,5 @@
+{
+  "toll": 1,
+  "tollCollector": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789",
+  "tollReceiver": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789"
+}
```
mds1, line 5: A structure like this might be better since it's more flexible and explicit, and allows the deploy script to throw if the supported network doesn't match a key listed here. Then we can also specify the paymaster addresses here instead of in the deploy script

```json
{
  "localhost": {
    "toll": 1,
    "tollCollector": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789",
    "tollReceiver": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789"
  },
  "rinkeby": {
    "toll": 1,
    "tollCollector": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789",
    "tollReceiver": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789"
  }
}
```

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+});
+
+// method to save the deploy info to a JSON file
+const save = (value, field, subfield = undefined) => {
+  if (subfield) {
+    parameters[field][subfield] = value;
+  } else {
+    parameters[field] = value;
+  }
+  fs.writeFileSync(fileName, JSON.stringify(parameters));
```
mds1, line 42: We should modify the output file name to include the network name, then modify the `.gitignore` to exclude all localhost deploy files but commit the rest. That way the most recent deploy on any network always lives in the repo

mds1 (changes requested):
Love the error handing that outputs to the JSON file! A few other comments:

------------------------

1. When I run start hardhat locally and run `yarn deploy:umbra --network localhost` it errors with `Deployment Error:  Error: invalid ENS name (argument="name", value=undefined, code=INVALID_ARGUMENT, version=providers/5.0.22)`. Is there a deploy setup step I'm missing?
2.  Have you tested this on Rinkeby? I don't see any Rinkeby specific sections that require a private key to be provided in the `.env` file
3. Can you add "Closes #77" as the very first text in the PR description? This way it automatically closes that issue when merged

mds1:
> When I run start hardhat locally and run yarn deploy:umbra --network localhost it errors with Deployment Error: Error: invalid ENS name (argument="name", value=undefined, code=INVALID_ARGUMENT, version=providers/5.0.22). Is there a deploy setup step I'm missing?

Turns out this was my fault—I had `deployParams.json` modified to match my suggestion in [this comment](https://github.com/ScopeLift/umbra-protocol/pull/92#discussion_r576231373), which caused the issue

contracts/scripts/deploy.js
jferas: Done

contracts/scripts/deployParams.json
apbendi: I like the approach suggested by @mds1.

contracts/scripts/deploy.js
apbendi: Agreed that it'd be nice to move this into the json config.

contracts/package.json
jferas: Good idea.. done

contracts/scripts/deployParams.json
jferas: Another good idea.. done

contracts/scripts/deploy.js
```diff
@@ -0,0 +1,106 @@
...
+    console.log('UmbraRelayRecipient contract deployed to address: ', umbraRelayRecipient.address);
+
+    const UmbraPaymaster = await ethers.getContractFactory('UmbraPaymaster', adminWallet);
+    const umbraPaymaster = await UmbraPaymaster.deploy(umbraRelayRecipient.address);
+    await umbraPaymaster.deployed();
+    save(umbraPaymaster.address, 'contracts', 'UmbraPaymaster');
+    console.log('UmbraPaymaster contract deployed to address: ', umbraPaymaster.address);
+
+    // set the relayer address on the Paymaster contract
+    await umbraPaymaster.setRelayHub(RINKEBY_PAYMASTER_PUBLIC_RELAYER_ADDRESS);
```
apbendi: Maybe it'd be worth logging the tx id where this took place?

apbendi (commented):
Overall this looks great @jferas! Left some minor comments, in addition to @mds1 suggestions which all make sense to me.

One last thing we'll need to do here is handle the ENS resolver deployment, once that work is done. I think we're getting close to having that wrapped up, but not sure if we should wait to add a commit to this PR or merge, then revisit once the resolver stuff is ready. Thoughts @mds1?

contracts/scripts/deploy.js
jferas: Good idea.. moved

contracts/scripts/deploy.js
jferas: Another good idea.. done

mds1:
> I think we're getting close to having that wrapped up, but not sure if we should wait to add a commit to this PR or merge, then revisit once the resolver stuff is ready. Thoughts @mds1?

I think either is fine. If this one is ready to be merged I'd say merge it and we can always open another PR

PR #96, opened 2021-02-22
Frontend updates

### LINKED ISSUE #48: Refactor app to support Typescript

mds1:
Closes #48 

Changes:
- Updates `umbra-js` so private key generation happens in that library
- Updates `umbra-js` so ENS supports publishing and using two public keys
- Updates `umbra-js` to add tests verifying our public key compression and decompression processes
- Core frontend functionality is complete (CNS integration and support for recipient IDs other than ENS are not yet implemented)

To test:
- Run `yarn` from project root
- `cd frontend && yarn run dev`

Right before merging, I'll also update the prettier config to use a line width of 120 and run it against the codebase. Doing that last to keep the diff more manageable

closes #48


apbendi:
When attempting to connect MetaMask before password is entered, I get the following error:

Screen Shot 2021-02-22 at 3 20 38 PM

After unlocking the wallet and clicking setup, I then receive the following:

Screen Shot 2021-02-22 at 3 20 58 PM

Refreshing afterwards resolves this, but can we handle the not-yet-unlocked-wallet case more gracefully?

apbendi:
In setup step 1, could we possibly link directly to the ENS app?

After completing setup step 2, the app does not display a confirmation or advance me to the next step. The button is pressable but gives this notification when pressed.

Screen Shot 2021-02-22 at 3 31 24 PM

Would it be easy to advance the user to step 3 automatically after signing?

Similarly, when step 3's tx is mined, no indication is given. Pressing the button again brings up MM to sign the tx again. I know we haven't quite polished all this yet, but would it be difficult to just redirect the user to the main page after the tx is mined, for example?

Screen Shot 2021-02-22 at 3 31 48 PM

apbendi:
On sending: after I minted USDC and then went to send some, I got a console error saying my balance was insufficient. This resolved after refresh. I assume we check only when the page loads? Would it be possible to at least show a toast for this error, and include something like "If your balance has changed, try refreshing the page."

apbendi:
On receiving, after the withdrawal tx is mined, nothing is updated in the UX (except for the little alert box— which I really like by the way— disappearing). How hard would it be to update the row to show the withdrawal as completed after the tx is mined?

mds1:
Checklist of changes to make from above:
- [ ] Fix MetaMask permissions when not on localhost
- [ ] In Setup step 1, link user directly to ENS app
- [ ] In Setup step 2, automatically advance user to step 3 after signing
- [ ] In Setup step 3, change screen to show a "success" with a link to the home page
- [ ] Move Blocknative notify UI to top-center for better visibility
- [ ] Re-check token balances right before send
- [ ] Update Receive table UI live right after withdrawals

apbendi (approved):
Awesome work 🎉 🎉 🎉

PR #103, opened 2021-03-05
Stricter input validation in umbra-js, other cleanup/ehancements

mds1:
Changes:
- To assist users who integrate with JavaScript instead of TypeScript, we add stricter input validation to methods in `umbra-js`.
- Bump minimum elliptic version to 6.5.4 since older versions had vulnerabilities
- Remove unused class getters to clean up `KeyPair.ts`
- We also expand the list of supported ENS domains

One current issue is that on commit I get the below error. I tried to make a change to the `tsconfig.json` file, but this didn't seem to work, cc @wildmolasses 

image

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -220,7 +220,6 @@ function useReceivedFundsTable(announcements: UserAnnouncement[]) {
         tx = await umbra.value.withdraw(
           spendingPrivateKey,
           token.address,
-          announcement.receiver,
           destinationAddress.value
         );
       } else {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,10 @@
 {
-    "files": [],
-    "references": [{ "path": "frontend" }, { "path": "umbra-js" }, { "path": "contracts" }],
-    "exclude": ["node_modules"]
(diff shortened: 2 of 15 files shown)


wildmolasses:
Hey Matt! You shouldn't need to specify the `typeRoots` property in the config. See the [docs](https://www.typescriptlang.org/tsconfig#typeRoots) on what `typeRoots` does. By default `node_modules/@types/**` should be included. But by providing a `typeRoots` entry we're saying use types ONLY from `node_modules/@types`, which means we can't use types from @umbra/contracts or elsewhere. I'm thinking this is how we blew up. But I'm guessing you made the change because some type seemed to be missing! If this is the case, can you give a little detail on what type issues you ran into before you updated the `typeRoots` property?

mds1:
So the main issue I was trying to fix was that first error in the screenshot: `Invalid type "any" of template literal expression` on line 37 of `KeyPair.ts`. Without the `typeRoots` property that error was still present, with it those others got added

The types for lines 36-37 are defined in `@types/elliptic`. Line 36 doesn't have a type (even though it should) and that carries over to line 37

apbendi (approved):
Looks great. Try as I might, could not find anything to complain about. Nice work 👍

PR #122, opened 2021-03-16
CNS setup

### LINKED ISSUE #88: Upgrade to Latest Unstoppable Library After Deprecating Ropsten
Per https://github.com/unstoppabledomains/resolution/issues/112, we've pinned ourselves to an older version of UD library due to a breaking change in v1.10.

It seems that once we've deployed the Phase 1 upgrade to Rinkeby, and deprecated our support for the original Ropsten testnet version, we should be able to upgrade to the latest version without issue.

mds1:
Closes #88 and adds back full CNS support

- CNS domains owned by user are fetched from The Graph and we use the first one returned
- During account setup, if user has a CNS and ENS name, we ask them which they are setting up
- CNS stores the 33 byte compressed public key
- Updates to the new contracts from #120 (so your receive table will be cleared)

### Testing

Switch to this branch and run `yarn && cd umbra-js && yarn build && cd ../frontend && yarn dev`

Flow 1:
- Register a test CNS name on Rinkeby [here](https://unstoppabledomains.com/developers) using an address that does NOT have a Rinkeby ENS name
- Go through the setup process with your CNS name, then send funds to yourself and make sure you can receive/withdraw them

Flow 2:
- Register a test CNS name on Rinkeby [here](https://unstoppabledomains.com/developers) using an address that does have a Rinkeby ENS name
- Go through the setup process with your CNS and/or ENS name, then send funds to yourself. If you've set up both ENS and CNS someone should be able to send to either name and you'd see the funds in both cases when scanning


frontend/src/pages/AccountSetup.vue
```diff
@@ -114,55 +122,84 @@
...
   const carouselStep = ref('1');
-  const isWaitingForUser = ref(false);
+  const isWaiting = ref(false);
+
+  onMounted(() => checkAndSetName());
+
+  function checkAndSetName() {
+    if (selectedName.value) return;
+    if (userEns.value && !userCns.value) selectedName.value = userEns.value;
+    else if (!userEns.value && userCns.value) selectedName.value = userCns.value;
```
apbendi, line 153: I find this segment really hard to read as formatted. I'd strongly favor expanding this out to add braces and newlines. My personal opinion is the readability outweighs the terseness, and also reduces likelihood of an error if the logic is expand at some point in the future.

umbra-js/src/utils/cns.ts
```diff
@@ -1,57 +1,100 @@
...
+  }
+  if (!isCnsDomain(name)) {
+    throw new Error(`Name does not end with supported suffix: ${supportedCnsDomains.join(', ')}`);
+  }
   return resolution.namehash(name);
 }
 
 /**
- * @notice For a given CNS domain, recovers and returns the public key from its signature
+ * @notice For a given ENS domain, returns the public keys, or throws if they don't exist
```
apbendi, line 47: This says ENS when it should say CNS, right? Maybe copy pasted from the other class?

umbra-js/src/utils/cns.ts
```diff
@@ -1,57 +1,100 @@
...
+    throw new Error(`Public keys not found for ${name}. User must setup their Umbra account`);
+  }
+  // Return uncompressed public keys
+  const spendingPublicKey = computePublicKey(compressedSpendingPublicKey);
+  const viewingPublicKey = computePublicKey(compressedViewingPublicKey);
   return { spendingPublicKey, viewingPublicKey };
 }
 
 /**
  * @notice For a given CNS domain, sets the associated umbra signature
```
apbendi, line 71: This say "sets the associated umbra _signature_" but I believe that's outdated, as we're just setting the two public keys directly now.

apbendi (commented):
Hey, this looks great overall! A couple of code/comment nits. Additionally, here's some things I ran into testing the frontend:

When I receive a payment from a CNS name, it doesn't show said name in the 'From' column on the Receive screen:

Screen Shot 2021-03-16 at 5 41 19 PM

When I try to setup my keys for CNS on a name that already has ENS, I simply see this screen. How do I set my CNS keys for name that already has ENS?

Screen Shot 2021-03-16 at 5 49 00 PM

frontend/src/pages/AccountSetup.vue
mds1: Heh, I do think there is value in terseness because one you grok that line, you benefit from fitting more on screen > can fit more context in my head at once, which I like. But, added braces and comments in 1723291!

umbra-js/src/utils/cns.ts
mds1: Yes! Fixed in 1723291

umbra-js/src/utils/cns.ts
mds1: Fixed in 1723291

mds1:
> When I receive a payment from a CNS name, it doesn't show said name in the 'From' column on the Receive screen:

This is an interesting point. With ENS, we just do a reverse lookup from the address and show the corresponding ENS name. But we can't do that for CNS—the best we can do is lookup _all_ CNS names owned by that address, and show the first one. Unlike ENS, it seems Etherscan does not let you see who owns a CNS name, so the only way you'd be able to see the actual sender is either (1) via a The Graph query like below, or (2) clicking the CNS name in the table to copy the address, then searching on Etherscan

What are your thoughts here? I'm on the fence about whether showing CNS names is better UX

> When I try to setup my keys for CNS on a name that already has ENS, I simply see this screen. How do I set my CNS keys for name that already has ENS?

It seems the reason is because that wallet address does not own a CNS name on Rinkeby, which is why it didn't show up. Here's how I checked:
1. Look up that `umbra-demo.eth` resolves to 0x91C727F6E9feA86A8DC968333c16f615c5603c8a
2. Open the CNS Graph Rinkeby registry [here](https://thegraph.com/explorer/subgraph/unstoppable-domains-integrations/dot-crypto-rinkeby-registry)
3. Use the following query to search for CNS names owned by `0x91c727f6e9fea86a8dc968333c16f615c5603c8a` (address must be lowercase):

```
{
  domains(where: {owner: "0x91c727f6e9fea86a8dc968333c16f615c5603c8a"}) {
    name
  }
}
```

And that returns: 

```
{
  "data": {
    "domains": []
  }
}
```

mds1:
- [ ] Show CNS names in receive table
- [ ] Support withdrawing ETH to CNS names (need to manually resolve)
- [ ] Support withdrawing tokens to CNS names (need to manually resolve)
- [ ] Fix Hide button in tables

mds1:
Ok, all things above are done. All address lookups and handling are handled in a new `address.ts` file. I didn't actually test the token withdraws, but we now convert any names entered to an address, so since ETH withdraws work token withdraws should also work

When withdrawing, if you enter an ENS or CNS name that doesn't resolve to an address, an error will be thrown. You can test this as follows:
- `udtestdev-test123.crypto` resolves to my address, so withdraw should work
- `udtestdev-msolomon.crypto` does not resolve to an address, so withdraw should error
- `msolomon.eth` resolves to my address, so should work
- `umbra-is-AWESOME.eth` probably does not resolve to an address, so withdraw should error

PR #125, opened 2021-03-17
Run prettier on contracts directory

apbendi:
Would love to understand why this doesn't seem to run on the contract directory in the commit hook? cc @wildmolasses


mds1 (approved):
Looks good! @apbendi up to you if we want to merge before @wildmolasses takes a look at why prettier config isn't running on commit, or if fixing that should be a separate PR

PR #131, opened 2021-03-19
Warn about bad privacy

### LINKED ISSUE #107: Warn Users About Poor Privacy Decisions
Allow the user to do the following, but only after confirming via a warning about privacy implications

#### Withdrawal:

- [ ] Withdrawal to the address they're logged in yet
- [ ] Withdrawal to ENS/CNS name
- [ ] Withdrawal to address that reverse resolves via ENS
- [ ] Withdrawal to address that owns a POAP??
- [ ] Withdrawal to address that has interacted with Gitcoin??

#### Send:

- [ ] Warn users it's a bad idea to send to themselves through Umbra— that it does not increase privacy— on the send screen

Warning should link to an FAQ/cheat sheet page to help.

mds1:
Closes #107

For now, we show a modal that gives two options: Learn More, or Acknowledge the Risks. The latter continues with withdrawal, and the former links to the FAQ page. Once we update the FAQ / write more docs, we can change where that links to if we want

- [x]  Withdrawal to the address they're logged in yet
- [x]  Withdrawal to ENS/CNS name
- [x]  Withdrawal to address that reverse resolves via ENS
- [x]  Withdrawal to address that owns CNS names
- [ ] Withdrawal to address that owns ENS names: They have a subgraph, but not sure how to write the proper query—create issue to implement this later
- [x]  Withdrawal to address that owns a POAP??
- [ ]  Withdrawal to address that has interacted with Gitcoin: There's an endpoint that returns all address that have donated to a grant, but it's a big file and CORS is not supported. Best option is probably to have a simple backend endpoint that stores all the addresses in a JSON file (which we'd update regularly). Then the endpoint returns true if the supplied address is in the file, false otherwise

### DIFF
--- a/frontend/src/utils/address.ts
+++ b/frontend/src/utils/address.ts
@@ -87,3 +87,46 @@ export const lookupOrFormatAddresses = async (addresses: string[], provider: Pro
   const promises = addresses.map((address) => lookupOrFormatAddress(address, provider));
   return Promise.all(promises);
 };
+
+// ================================================== Privacy Checks ===================================================
+
+// Checks for any potential risks of withdrawing to the provided name or address, returns object containing
+// a true/false judgement about risk, and a short description string
+export const isAddressSafe = async (name: string, userAddress: string, domainService: DomainService) => {
+  userAddress = getAddress(userAddress);
+
+  // Check if we're withdrawing to an ENS or CNS name
+  if (ens.isEnsDomain(name)) return { safe: false, reason: `${name}, which is a publicly viewable ENS name` };
+  if (cns.isCnsDomain(name)) return { safe: false, reason: `${name}, which is a publicly viewable CNS name` };
+
+  // We aren't withdrawing to a domain, so let's get the checksummed address.
+  const address = getAddress(name);
+  const provider = domainService.provider as Provider;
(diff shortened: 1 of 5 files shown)


frontend/src/utils/address.ts
```diff
@@ -87,3 +87,46 @@ export const lookupOrFormatAddresses = async (addresses: string[], provider: Pro
...
+
+  // Check if address owns a CNS name
+  const cnsName = await lookupCnsName(address, provider);
+  if (cnsName) return { safe: false, reason: `an address that resolves to the publicly viewable CNS name ${cnsName}` };
+
+  // Check if address is the wallet user is logged in with
+  if (address === userAddress) return { safe: false, reason: 'the same address as the connected wallet' };
+
+  // Check if address owns any POAPs
+  if (await hasPOAPs(userAddress)) return { safe: false, reason: 'an address that has POAP tokens' };
```
apbendi, line 118: Looks like you're passing the wrong address to the `hasPOAPs` method. This is the logged in user, not the address the user has put in for withdrawal.

mds1:
Feedback addressed in last commit, so going to merge!

PR #133, opened 2021-03-19
Advanced UI features

### LINKED ISSUE #106: Advanced UI Features
Features that should be flagged behind some kind of "advanced" only toggle that adds friction and with disclaimer to only use if you know what you're doing.

- [ ] Export Private Key Of Receiving Stealth Address
- [ ] Define start and end blocks for receive Announcement scanning
- [ ] Send to a general ETH address using recovered TX pub key, transaction hash, or specific Public Key
- [ ] Include a Payload Extension when sending
- [ ] Display Payload Extension on receives (should this be there even in non-advanced mode?)
- [ ] After withdraw, clicking on the word "Withdraw" in the receive table should take the user to the withdrawal tx on Etherscan

### LINKED ISSUE #124: Send Fields Should Not Be Editable While TX is Pending
When a user is in the process of sending funds, and the UI is waiting for transaction signature/confirmation, the fields for recipient, token, and amount should be greyed out and otherwise not editable.

Screen Shot 2021-03-16 at 5 53 10 PM

mds1:
Closes #124, closes #106 (moves one task from #106 into #135 since I want to think of the cleanest way to do that and there's more pressing functionality first). Changes:

1. Export Private Key Of Receiving Stealth Address
2. Define start and end blocks for receive Announcement scanning
3. Send to a general ETH address using recovered TX pub key, transaction hash, or specific Public Key
4. Include a Payload Extension when sending
5. Display Payload Extension on receives (should this be there even in non-advanced mode?)

Notes: 
- The advanced toggle is in the footer
- Items 1, 2, and 3 are advanced-mode only feature. Items 4 and 5 are always present
- When advanced mode is on, the 🧙 emoji is shown in the top right with tooltip on hover. 
- Settings such as dark mode, start and end blocks, and advanced mode toggle are stored in local storage in a new Settings store (custom private key for scanning is intentionally not stored)

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -14,6 +14,7 @@
       <div class="text-center text-italic">Scanning for funds...</div>
     </div>
 
+    <!-- Received funds table -->
     <q-table
       v-else
       :columns="mainTableColumns"
@@ -27,9 +28,7 @@
       <!-- Header labels -->
       <template v-slot:header="props">
         <q-tr :props="props">
-          <q-th v-for="col in props.cols" :key="col.name" :props="props">
-            {{ col.label }}
-          </q-th>
+          <q-th v-for="col in props.cols" :key="col.name" :props="props"> {{ col.label }} </q-th>
           <q-th auto-width />
         </q-tr>
       </template>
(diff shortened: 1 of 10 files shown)


apbendi (approved):
Overall this looks and works great! Really nice work on this— the features are all helpful. I like the memo functionality a lot and think it's good to leave it in by default.

A couple of very minor nits:

1) When exporting a private key from MetaMask, it defaults to not having the `0x` prefix. Given this, I wonder if it would be better UX not to require it:

Screen Shot 2021-03-30 at 5 22 17 PM

2) When typing in a memo, it's unclear how much more I will be able to type, or how much I have to delete, to fit in the space available. Due to code points in Unicode, their won't be any easy way to give a "character" limit, but I wonder if we could have some kind of progress bar or little circle-pie-chart type thing that fills up or shrinks based on the bytes used as the user types?

mds1:
Added support for both 1 and 2 in the last two commits

apbendi:
Looks and works great. Nice 👍

Issue #104, opened 2021-03-08
Convert umbra-js tests to HardHat and add Coverage Report


apbendi:
Currently the tests use OZ tests environment and don't produce coverage.

mds1:
Hey @wildmolasses, it seems right now running `yarn test` from the root causes a failure in an umbra-js test, but running `yarn test` from within the umbra-js folder does not. See https://github.com/ScopeLift/umbra-protocol/pull/145#pullrequestreview-626370648

Do you mind see if that's still the case after doing the Hardhat refactor as part of this issue?

PR #145, opened 2021-04-01
Switch to noble-secp256k1

mds1:
(Built off the branch in #133, so that needs to be merged first, then we'll rebase this against the master)

- Implements noble-secp256k1 for all ECC operations
- A bit of cleanup (remove unneeded padHex method)
- Need to double check that we're not using elliptic behind the scenes from an ethers method
- We also change the shared secret computation, such that the shared secret is now the hash of the resulting point, instead of the point itself. This is preferred as in theory it should make the output symmetric keys more uniformly distributed. Consequently, **this PR is a breaking change** and once merged, past sends will no longer display in the UI when building locally

### DIFF
--- a/umbra-js/package.json
+++ b/umbra-js/package.json
@@ -25,16 +25,14 @@
     "bn.js": "^5.1.3",
     "buffer": "^6.0.2",
     "dotenv": "^8.2.0",
-    "elliptic": "^6.5.4",
     "eth-ens-namehash": "^2.0.8",
     "ethers": "^5.0.25",
-    "js-sha3": "^0.8.0"
+    "noble-secp256k1": "^1.1.2"
   },
   "devDependencies": {
     "@openzeppelin/test-environment": "^0.1.6",
     "@types/bn.js": "^5.1.0",
     "@types/chai": "^4.2.14",
-    "@types/elliptic": "^6.4.12",
     "@types/mocha": "^8.0.4",
     "@umbra/contracts": "^0.0.1",
     "chai": "^4.2.0",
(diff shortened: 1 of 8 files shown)


apbendi (commented):
Looks great! Running the tests I got the following error:

```
1) Utilities
@umbra/umbra-js:        Recipient identifier lookups
@umbra/umbra-js:          looks up recipients by address:
@umbra/umbra-js:      Error: Transaction not found. Are the provider and transaction hash on the same network?
@umbra/umbra-js:       at /Users/bendi/Development/umbra/umbra-js/src/utils/utils.ts:46:11
@umbra/umbra-js:       at step (src/utils/utils.ts:36:23)
@umbra/umbra-js:       at Object.next (src/utils/utils.ts:17:53)
@umbra/umbra-js:       at fulfilled (src/utils/utils.ts:8:58)
@umbra/umbra-js:       at processTicksAndRejections (internal/process/task_queues.js:97:5)
@umbra/umbra-js: error Command failed with exit code 1.
```

which is possibly related to this warning?

```
@umbra/umbra-js: ========= NOTICE =========
@umbra/umbra-js: Request-Rate Exceeded  (this message will not be repeated)
@umbra/umbra-js: The default API keys for each service are provided as a highly-throttled,
@umbra/umbra-js: community resource for low-traffic projects and early prototyping.
@umbra/umbra-js: While your application will continue to function, we highly recommended
@umbra/umbra-js: signing up for your own API keys to improve performance, increase your
@umbra/umbra-js: request rate/limit and enable other perks, such as metrics and advanced APIs.
@umbra/umbra-js: For more details: https://docs.ethers.io/api-keys/
@umbra/umbra-js: ==========================
```

Can we add an API key as a required env var setting to prevent this?

mds1:
Added info about the above test failure to https://github.com/ScopeLift/umbra-protocol/issues/104#issuecomment-811981267 and will track it there

PR #146, opened 2021-04-01
Make sure the y18n dependency resolves to non-vulnerable version

apbendi:


mds1:
Reference: https://github.com/advisories/GHSA-c4w7-xm78-47vh

mds1:
Just rebased against the latest master. All tests pass, so merging!

PR #150, opened 2021-04-02
Remove elliptic

mds1:
- Replaces `ethers.wallet.createRandom()` with noble-secp256k1's `randomPrivateKey()`, since `createRandom()` ultimately passes through elliptic at one point
- Removes the `eth-ens-namehash` package for getting the namehash of ENS names and we now use the ethers.js method instead

### DIFF
--- a/umbra-js/src/classes/KeyPair.ts
+++ b/umbra-js/src/classes/KeyPair.ts
@@ -2,10 +2,9 @@
  * @dev Class for managing secp256k1 keys and performing operations with them
  */
 
-import { getSharedSecret as secpGetSharedSecret, getPublicKey, Point, CURVE } from 'noble-secp256k1';
-import { Wallet } from 'ethers';
+import { getSharedSecret as secpGetSharedSecret, utils, getPublicKey, Point, CURVE } from 'noble-secp256k1';
 import { BigNumber } from '@ethersproject/bignumber';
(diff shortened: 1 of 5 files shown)


umbra-js/src/classes/KeyPair.ts
```diff
@@ -92,16 +91,14 @@ export class KeyPair {
       throw new Error('Must provide instance of RandomNumber');
     }
     // Get shared secret to use as encryption key
-    const ephemeralWallet = Wallet.createRandom();
-    const sharedSecret = getSharedSecret(ephemeralWallet.privateKey, this.publicKeyHex);
+    const ephemeralPrivateKey = hexlify(utils.randomPrivateKey()); // private key as hex with 0x prefix
+    const ephemeralPublicKey = `0x${getPublicKey(ephemeralPrivateKey.slice(2))}`; // public key as hex with 0x prefix
```
apbendi, line 95: Why use hexlify in the line above but string interpolation here? Is it doing the same thing?

apbendi (approved):
Looks good and tests pass. Note one small question. Feel free to merge after resolving or responding.

umbra-js/src/classes/KeyPair.ts
mds1: Hexlify is taking in a Uint8Array of bytes in that case and converting it to a hex string

PR #151, opened 2021-04-03
ENS resolver upgrade implementation

mds1:
Closes #113 

Everything should be working, but admittedly there's a bit of complexity and lots of checks to this flow, so there may still be some bugs for edge cases—please test thoroughly!

PR #153, opened 2021-04-05
Remove unused dependencies and methods

### LINKED ISSUE #152: Remove more umbra-js dependencies
- `bn.js` is not used anywhere
- `buffer` is used in one place, can be replaced with `Uint8Array`
- `dotenv` is used in once place, seems like it should be devDependency

mds1:
Closes #152 

(branched off #151, so figured we can just merge this into 151's branch first since this will be a much quicker PR to review/merge)

### DIFF
--- a/umbra-js/package.json
+++ b/umbra-js/package.json
@@ -22,9 +22,6 @@
   "license": "ISC",
   "dependencies": {
     "@unstoppabledomains/resolution": "1.20.0",
-    "bn.js": "^5.1.3",
-    "buffer": "^6.0.2",
-    "dotenv": "^8.2.0",
     "ethers": "^5.0.25",
     "noble-secp256k1": "^1.1.2"
   },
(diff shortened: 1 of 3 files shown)


paulmillr:
seems like this doesn't help much judging from lockfile.

Well, since ethers.js wants to include noble in the next version (ethers-io/ethers.js#727, [twitter](https://twitter.com/ricmoo/status/1378102168281038849)), I think you'll see the real pkg size difference right after that.

mds1:
Yea ended up being a small change. Since it's a monorepo, a good portion of the lockfile comes from dependencies of the fronted + contracts workspaces too, so elliptic and it's dependencies likely won't be removed from lockfile anytime soon. But they still _should_ be tree-shaken out of the production umbra-js bundle, though I haven't looked into that yet

apbendi (approved):
lgtm!

PR #161, opened 2021-04-15
umbra-js coverage and enhancements

mds1:
- Improves coverage of umbra-js files. Most code not covered by testing is either (1) setting public keys, or (2) `throw` statements from error conditions that are tricky / not clear how to get into 
- To help prevent stuck funds:
    - Adds list of addresses that the Umbra class `send` method will not send to
    - Adds list of keys that the KeyPair class will not initialize with
- Updates umbra-js dependencies to latest version
- Coverage script now removes builds folder before running

@wildmolasses You'll notice I also added `"skipLibCheck": true,` to the `frontend` and `umbra-js` tsconfig files to handle complaints that TS threw after upgrading dependencies. Let me know if you prefer a different way of handling this

### DIFF
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -2,12 +2,13 @@
   // "extends": "@quasar/app/tsconfig-preset",
   "extends": "../tsconfig.settings.json",
   "compilerOptions": {
-    "composite": true,
-    "outDir": "build",
-    "rootDir": "src",
     "baseUrl": ".",
+    "composite": true,
     "esModuleInterop": true,
+    "outDir": "build",
     "resolveJsonModule": true,
+    "rootDir": "src",
+    "skipLibCheck": true,
     "paths": {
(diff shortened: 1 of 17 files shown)


wildmolasses:
Thanks for the heads up! as we've discussed before, `skipLibCheck` is a leaky config property -- if we use it here, we'll also have to use it in any ts project that depends on @umbra/umbra-js. IMO it's worth a little investigation to see if we can't fix the underlying issue.

In this case I noticed our yarn.lock is making use of @sinonjs/fake-timers@7.0.5 as well as @types/sinon@9.0.11. I checked out the [@sinonjs/fake-timers package README](https://www.npmjs.com/package/@sinonjs/fake-timers) and they say there that their typescript types are not yet production ready. They suggest a workaround in the tsconfig file. I implemented the workaround, removed the skipLibCheck property, and confirmed that packages build.

Issue #140, opened 2021-03-30
Copy for website FAQ Page


apbendi:

mds1:
## Introduction

- What is Umbra?
- Can you walk me through an example?
- How does it work?
- How does Umbra compare to Tornado Cash and Aztec?

## Account Setup

- Is this step required?
- Should I use my own ENS/CNS name or an umbra.eth subdomain?
- Can my subdomain be revoked?

## Sending

- Why are only certain tokens available?
- When will the recipient receive their funds?
- Why is the maximum memo length so short?

## Receving

- What address are safe for withdrawing funds to?
- Why does it take so long to scan for my funds?
- When are my funds available to withdraw?

## Technical Details

- How does it work? (more details)
- What are spending and viewing keys?

## For Developers

- How can I build on top of Umbra?
- How can I receive a user's viewing key, but not their spending key?
- What are Hooks and how do I use them?
- What is the payload extension and how do I use it?

## Advanced Mode

- What is Advanced Mode?
- How do I send funds to a user by their address or public key?
- How can I access funds sent to an address I own?
- How can I scan just certain range of blocks?

mds1:
Writing up answers here before adding them to the site: https://hackmd.io/@scopelift/BkWfbt_HO

PR #164, opened 2021-04-16
Add faq

mds1:
Closes https://github.com/ScopeLift/umbra-protocol/issues/140

- Adds FAQ page (link in header). Each FAQ question and section header can be clicked to copy a URL to the clipboard. Navigating to the URL will open that FAQ question on page load
- Adds Terms of Service (link in footer). Don't love the location of the Terms of Service link so open to alternative locations for that link / alternative Footer layouts


apbendi (changes requested):
Looks good, thanks for banging this out! A few things (and sorry in advance to nitpick on some of these!)

1. 
>  Each FAQ question and section header can be clicked to copy a URL to the clipboard. Navigating to the URL will open that FAQ question on page load

This sounds awesome, but doesn't seem to work for me. Clicking the headers does nothing. Clicking a question opens it, but does not change the URL or copy anything to the clipboard. Am I missing something? This is in Chrome, btw.

2. I sent the Privacy Policy separately, do you mind adding that as well?

3. Layout wise, here's a quick mockup of something that could possibly work a little better? I dunno, let me know what you think. Obviously it's done sloppily and not even properly aligned/centered, but I think you can see what I was going for.

Screen Shot 2021-04-20 at 7 31 14 AM

The RSS icon here is for a link to Umbra tagged ScopeLift blogposts, i.e. https://www.scopelift.co/blog/tag/umbra

I'd be ok with using something other than the RSS icon here (a generic web globe icon, for example? Or a document, or whatever makes sense). It was just an easy one to grab in the mockup tool.

The advanced mode toggle could live right about the horizontal divider line on the left. Or even better, I think putting that stuff up in the top right might be ideal. On my rather large 4K monitor, on slightly boomer-eyes resolution, those controls are well below the fold. For a lot of people, they might not see them. I'll leave it up to you whether it's worth moving, though.

mds1:
> This sounds awesome, but doesn't seem to work for me. Clicking the headers does nothing. Clicking a question opens it, but does not change the URL or copy anything to the clipboard. Am I missing something? This is in Chrome, btw.

- When I float the mouse over a section header, a link icon shows up. Clicking copies the section URL to the clipboard and shows a success notification
- When I click a question, the section expands and it copies the header URL to the clipboard with no visual feedback (but maybe it should update the URL shown, hmm 🤔 )

Neither of these happen for you? I tested in Brave and Chrome and both work

> I sent the Privacy Policy separately, do you mind adding that as well?

Will do that now

> Layout wise...

I do like having the advanced in the footer to add friction to enabling it, since it can be risky and you really should know what you're doing. There's also the dark mode toggle, and those don't fit too well in the above mockup. It's also confusing whether the icons link to ScopeLift things or Umbra things. 

I think maybe 3 vertical columns would look good. Like a "Settings" column with the two toggles stacked vertically, an "About" column in the middle with links to ScopeLift, ToS, Privacy, then a "Links" column on the right for GitHub, Telegram, Twitter, etc. I'll search for some inspiration

mds1:
I also just ran Prettier on the full repo, since there was on unformatted file (`Umbra.sol`)

mds1:
Last todo before merging: Update URL when clicking a question

PR #165, opened 2021-04-21
Wallet and code cleanup

### LINKED ISSUE #132: Cleanup and maintenance
Just some small tasks that will clean things up a bit / make things more readable/maintainable:

- With ethers, we import using `import { something } from @ethersproject/<package>`. We do this because it should reduce bundle sizes, but it results in lots of import lines and is hard to read. I'll create a file that just directly exports all the ethers methods we use, that way files get a bit easier to read since all the imports are centralized
- `alerts.ts` is setup as a composition function, but since it's stateless that's not really necessary. Split this into regular exported methods

mds1:
Closes #132 

- Reduces set of supported wallets to only those known to use deterministic signatures (i.e. those used by zkSync)
- Cleans up imports so it's easier to track which ethers packages we use
- Remove unnecessary composition function in alerts.ts 
- Add a few clarifying comments
- Update docs in `umbra-js/README.md`, including info on how to generate a full API reference with typedoc
- Bumps noble-secp256k1 version to v1.2.0 (first release after its audit)


apbendi (approved):
LGTM 👍

PR #168, opened 2021-04-26
Umbrajs input validation + other cleanup

mds1:
- Adds a few `getAddress` checks to validate addresses + ensure we always use checksum addresses
- Add script to `package.json` to simplify umbra-js doc generation
- Add files/folders to eslintignore—the umbra-js `yarn lint` was failing and this fixes that
- In `Umbra.sol`, `ETH_TOKEN_PLACHOLDER` did not have it's visibility specified, so made it public
- Ran Prettier which formatted the README

### DIFF
--- a/contracts/contracts/Umbra.sol
+++ b/contracts/contracts/Umbra.sol
@@ -29,7 +29,7 @@ contract Umbra is Ownable {
   // ======================================= State variables =======================================
 
   /// @dev Placeholder address used to identify transfer of native ETH
-  address public constant ETH_TOKEN_PLACHOLDER = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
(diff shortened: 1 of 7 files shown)


contracts/contracts/Umbra.sol
```diff
@@ -29,7 +29,7 @@ contract Umbra is Ownable {
   // ======================================= State variables =======================================
 
   /// @dev Placeholder address used to identify transfer of native ETH
-  address constant ETH_TOKEN_PLACHOLDER = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+  address public constant ETH_TOKEN_PLACHOLDER = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
apbendi, line 32: Curious why we'd need to make this public?

apbendi (commented):
Looks good, but added one comment/question: what's the point of making the placeholder address public?

PR #171, opened 2021-04-30
Final Rinkeby and Mainnet Contract Deployments

apbendi:


contracts/scripts/deployParams.json
```diff
@@ -10,5 +10,11 @@
     "tollCollector": "0x5d11E5f17eCaFc43A8D966E4F37809C07208C80E",
     "tollReceiver": "0x91C727F6E9feA86A8DC968333c16f615c5603c8a",
     "payMasterPublicRelayer": "0x53C88539C65E0350408a2294C4A85eB3d8ce8789"
+  },
+  "mainnet": {
+    "toll": 0,
+    "tollCollector": "0x12aE679061C58e9f7660A2f7e6f06573014d2615",
+    "tollReceiver": "0xb4435399AB53D6136C9AEEBb77a0120620b117F9",
+    "payMasterPublicRelayer": ""
```
mds1, line 18: I think we can delete this since it's no longer used

mds1 (approved):
Just one small comment, feel free to merge after otherwise

contracts/scripts/deployParams.json
mds1: `payMasterPublicRelayer` is used if GSN is deployed, which is optional, so it's there as a placeholder

PR #176, opened 2021-05-05
Frontend polish

### LINKED ISSUE #148: Design / UX improvements
- [x] Make things look better (in light mode / dark mode)
- [x] Hide "Account Setup" tab once complete
- [x] Show loading / UX feedback when connecting wallet
- [x] Div that can be clicked for "Connect wallet" is full width, not just the button
- [x] Improve the currently confusing UX of the account setup page shown in the screenshot below
- [x] Show pop up modal on withdraw confirming the action
- [x] Link warning at the top to security page
- [x] Auto-connect last used wallet
- [ ] Test other wallets

image

### LINKED ISSUE #175: Improve Account Setup UX

### LINKED ISSUE #177: Handle when a user changes accounts
Currently we do nothing and the user needs to refresh, which can results in weird states, mismatched Umbra keys, etc.

mds1:
Closes #148, closes #177, closes #175

- [x] Tweaks to make things look better (in light mode / dark mode)
- [x] Hide "Account Setup" tab once complete
- [x] Show loading / UX feedback when connecting wallet
- [x] Div that can be clicked for "Connect wallet" is full width, not just the button
- [x] Improve the currently confusing UX of the account setup page shown in the screenshot below (kind of....)
- [x] Show pop up modal to confirm before withdrawing
- [x] Link the warning at the top to security page
- [x] Auto-connect last used wallet


apbendi:
Navigating to the site, clicking MM in wallet connect, this is what the screen looks like after connecting. The indicator is active but the screen doesn't change.

Screen Shot 2021-05-06 at 10 31 19 AM

mds1:
TODO: Add a check when user connects their wallet that they are on either mainnet or Rinkeby, since those are the only supported networks

apbendi:
After completing setup with a subdomain and returning home, the UI does not show the "Setup" button for about 500 ms, then it blinks back into place. It shouldn't be there, right?

Additionally, after setting up my umbra subdomain, the app still shows my normal ENS name in the top right. I can guess that's happening because we're just doing reverse resolution, but I can imagine it being pretty confusing to users to set up an ENS subdomain for the app then still have their regular ENS name showing.

apbendi:
While setting up my main ENS name on step 3, I was able to click the fourth dot while waiting for the first 2 txs to complete. 

Screen Shot 2021-05-06 at 10 48 23 AM

This moved me to the next screen which said I was done and didn't give me a way to go back. Then a couple seconds later, I was asked to sign another tx:

Screen Shot 2021-05-06 at 10 48 31 AM

Is there any reason to let users click on these dots at all? There useful as an indicator of steps, but I don't think we want them used as navigation

apbendi:
Is there a specific section of the FAQ we should be pushing users in this "Learn More" button. Maybe faq#receiving-funds ?

Screen Shot 2021-05-06 at 10 53 29 AM

apbendi:
When I put my umbra subdomain as the withdrawal address, I get this error:

Screen Shot 2021-05-06 at 10 53 37 AM

This makes sense, because we're not setting an address for the user when they register this, only their stealth keys. This brings up the question: should we be setting their address? Do we want people to use their `umbra.eth` subdomain in generic ENS contexts? Or is it better to have user's think of it as a special recieve-via-Umbra-only address? Honestly not sure.

If the latter, should we catch and specifically acknowledge the case where a user tries to withdraw to a subdomain? Perhaps show a message like "`umbra.eth` subdomains are only useful for people sending you funds" or something?

apbendi:
Switching account in MM updates the address in the top right corner, but does not seem to change which keys in memory are used. Receive page still scans prior account keys.

apbendi:
Any reason for the 24-hr time on the withdrawal screen (instead of just my local time format?)

Screen Shot 2021-05-06 at 1 41 58 PM

apbendi:
When sending a token, this message is only sent *AFTER* I've done an `approve` tx:

Screen Shot 2021-05-06 at 2 04 19 PM

apbendi:
Maybe change the warning copy to "This is beta software. Use at your own risk."

apbendi:
I like the relayer fee confirm flow a lot! Thinking out loud:

I wonder what it would look like if the final total was bold and/or slightly larger? Also, I wonder how it would look if the fee part was in red and/or had a minus sign in front?

Screen Shot 2021-05-06 at 2 07 18 PM

PR #178, opened 2021-05-11
umbra-js security improvements

mds1:
Contingent on https://github.com/ScopeLift/umbra-protocol/pull/176 being merged first, since this builds off that PR (hence all the extra commits)

**Random Number Fixes** 
- `RandomNumber` class ignores payload extension input (for now) and always generates 32 byte random numbers
- Memo field was removed from the send form

**Point validation**
- Adds a new utility method in umbra-js called `assertValidPoint()` which verifies that a point is on the secp256k1 curve

**Other**
- I noticed wrongly named function inputs in `DomainService.ts`, so fixed that also

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -86,9 +86,6 @@
                   <span>{{ col.value.from }}</span>
                   <q-icon class="copy-icon" name="far fa-copy" right />
                 </div>
-                <div v-if="hasPayloadExtension(props.row.randomNumber)" class="text-caption text-grey">
-                  {{ formatPayloadExtensionText(props.row.randomNumber) }}
-                </div>
               </div>
 
               <!-- Default -->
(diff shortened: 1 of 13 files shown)


apbendi (commented):
Haven't completed a full review, but here are a couple things I'm seeing so far:

1. I'm still seeing memos on the Receive page for old sends. Shouldn't that be removed?

2. I connected my wallet, went to Receive, and saw old payments. Then I went to Send, sent 0.1 ETH to myself, went back to Receive, and saw this:

Screen Shot 2021-05-11 at 11 05 48 AM

Screen Shot 2021-05-11 at 11 06 40 AM

It happens shortly after the app starts scanning, and after it occurs, the Receive screen is blank except for the header.

mds1:
Both should be fixed now. There were some remnants of payload extension logic which was the reason for (1), and removing that seems to have resolved (2) as a I couldn't reproduce it—this makes sense since that error message is thrown by the strings package of ethers.js

mds1:
Replaced by https://github.com/ScopeLift/umbra-protocol/pull/179

Issue #189, opened 2021-05-21
Feedback thread


mds1:
### Account Setup

- [ ] 1. Confusion around why their ENS name is not detected (this happens when the user does not have the reverse resolver set). Maybe we use a Graph query to show all names owned by the user
     - TBD: Either leave it and tell users to set reverse resolver OR show all names you own
- [x] 2. In Setup, Step 3: Publish Keys, the user's name is missing. It says "...by sending funds to ." but should say "...by sending funds to yourname.eth"
     - Fix this
     - Done in 3aab2b7712a1de6c521d18ca62e6c3e4a71ed2f6. [Before](https://user-images.githubusercontent.com/17163988/119672959-db5bec00-bdef-11eb-95fc-ee002edd2dca.png). [After](https://user-images.githubusercontent.com/17163988/119672676-a51e6c80-bdef-11eb-9045-bb2aabc63925.png)
- [x] 3. For subdomain setup, copy should indicate that there's two transactions (similar to how regular ENS indicates that there's three)
     - Fix this
     - Done in 6ee296e72872a4c31d8a9d2cebc6f4c456237a48. [Before](https://user-images.githubusercontent.com/17163988/119673640-5f15d880-bdf0-11eb-96f3-7abd3e865888.png). [After](https://user-images.githubusercontent.com/17163988/119673322-1f4ef100-bdf0-11eb-91f0-88d6e723e159.png)
- [ ] 4. Names that use an old version of the public resolver say "your name is not using the public resolver"
     - Non-blocker: Tell user they need to migrate to new Public Resolver. Link to ENS app. Question: What is address of old public resolver?
- [x] 5. You can click "next" on step 1 without selecting a name. (Nothing bad would happen, since you wouldn't be able to set keys, but still should prevent this)
     - Fix this
     - Done in 7b9e31f

### Sending

- [x] 1. Confusion that you can't just send to `name.eth` if that user has not setup their account
    - Fix: Make error message more clear that `name.eth` needs to setup their Umbra account
    - Done in c25fab7

### Receive Table / Withdrawing

- [ ] 1. On Receive Table, the toggle from "Withdraw" to "Hide" is confusing. Maybe change "Hide" to "Hide withdraw form"?
    - Non-blocker
    - Options: (1) "Withdraw \/" and "Withdraw /\" OR (2) "Withdraw \/" and "Hide /\" OR (3) "Withdraw" and an "X" button?
[...]

PR #190, opened 2021-05-26
Implement user feedback

mds1:
Resolves all blockers described in https://github.com/ScopeLift/umbra-protocol/issues/189

Note that once deployed this will also update the signature used to access your account, and everyone who has completed setup so far will need to repeat the setup process


apbendi (approved):
Looks good!

I've create #193, #192, and #191 to track minor feedback from this issue. Otherwise, good to go for merge and deploy!

PR #196, opened 2021-06-04
Enforce minimum send amount to prevent "stuck" funds

mds1:


frontend/src/pages/AccountSend.vue
```diff
@@ -80,12 +80,19 @@ function useSendForm() {
...
   }
 
+  const isEth = (address: string) => getAddress(address) === '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+
   function isValidTokenAmount(val: string) {
     if (!val || !(Number(val) > 0)) return 'Please enter an amount';
     if (!token.value) return 'Please select a token';
+
     const { address: tokenAddress, decimals } = token.value;
+    if (Number(val) < 0.01 && isEth(tokenAddress)) return 'Please send at least 0.1 ETH';
```
apbendi, line 90: Looks like the error message is off. Should read 0.01 ETH, no?

apbendi (changes requested):

frontend/src/pages/AccountSend.vue
mds1: Just fixed!

apbendi (approved):

PR #203, opened 2021-07-27
Disable withdrawal button if fee is zero/missing

davidlaprade:
Adds the following functionality:

Tests were all manual.

I manually verified that [DAI](https://rinkeby.etherscan.io/tx/0xec206e938e9c40808d5ed9088d67eb5ca4518b7c158900f2af41dac49716a02d) and [ETH withdrawals](https://rinkeby.etherscan.io/tx/0x9194981a1393fe2a41511a1332656de5f0bad12f362605996f6b4b0c41e73a50) still work

PR #206, opened 2021-07-30
Fronted features and fixes

### LINKED ISSUE #192: Prevent Scary Errors If User Closes Metamask
Upon returning to the site, the user is asked to log into metamask. If they dismiss metamask, they get a scary error. Suppress these errors and instead just ask the user to log in again the next time they click one of the buttons. This occurs when MetaMask is locked (i.e. requires a password)

IMAGE 2021-05-27 17:33:43

mds1:
Changes in this PR:

1. Lets user choose gas price when withdrawing ETH: The default gas price to use is queried from the node, and if the user wants to change it, they can. See #202 and #203 for more info / video demos

2. Adds payment links: On the send form, fill in any combination of fields and click the new "Generate Payment Link" button. This puts a URL on the clipboard that, when visited, pre-populates the form with the info in the link. It should work even if you visit the send page without a wallet connected

3. Prevents the "Unknown account #0" error when user rejects wallet connect prompt (resolves #192). More info in #205 

4. Fixes small UI error where modal is not on top of the page. More info in #205 

Thanks to @davidlaprade for helping with most of these!

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -380,8 +380,8 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
   }
 
   type ExecuteWithdrawalOptions = {
-    gasPrice?: BigNumber // fee in wei
-  }
+    gasPrice?: BigNumber; // fee in wei
+  };
 
   /**
    * @notice Executes the withdraw process
@@ -404,7 +404,7 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
       isWithdrawInProgress.value = true;
       let tx: TransactionResponse;
       if (token.symbol === 'ETH') {
-        const {gasPrice} = options;
+        const { gasPrice } = options;
         tx = await umbra.value.withdraw(spendingPrivateKey, token.address, acceptor, { gasPrice });
         txHashIfEth.value = tx.hash;
         txNotify(tx.hash);
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -379,10 +379,14 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
     showConfirmationModal.value = true;
   }
 
+  type ExecuteWithdrawalOptions = {
+    gasPrice?: BigNumber // fee in wei
+  }
+
   /**
    * @notice Executes the withdraw process
    */
-  async function executeWithdraw() {
+  async function executeWithdraw(options: ExecuteWithdrawalOptions) {
     if (!umbra.value) throw new Error('Umbra instance not found');
     if (!provider.value) throw new Error('Provider not found');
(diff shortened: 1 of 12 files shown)


apbendi (commented):
Overall looks and works great. Nice job @mds1 and @davidlaprade 🎉 

I'm seeing a bunch of test failures of `umbra-js` on this branch. Can you reproduce that? Those tests do not fail on master for me.

Here are some things I noticed in my testing:

* When the "Confirm Withdrawal" screen first pops up, it does a weird resizing thing where it shows up big then shrinks down:

umbra-withdrawal-flash

* Is there an underscore after this edit icon?

Screen Shot 2021-07-29 at 10 37 58 PM

* Instead of "Generate payment link", how about "Copy Shareable Payment Link" with the little clipboard icon for the copy?
* After I click confirm, the gas price I've selected is still editable, when it really shouldn't be:

Screen Shot 2021-07-29 at 10 42 35 PM

mds1:
@apbendi Addressed all changes. Changed the button text to "Copy payment link" to keep a bit more concise, and added an FAQ question explaining the button

mds1:
> I'm seeing a bunch of test failures of umbra-js on this branch. Can you reproduce that? Those tests do not fail on master for me.

I just ran `yarn` at the root and `yarn test` in the umbra-js folder and everything passed, so I'm not able to reproduce this. Can you try again and post the specific failures you're seeing?

apbendi:
Overall the changes look good! Great idea on the FAQ item. Copy looks good and good call out on the privacy implications.

One thing I noticed while testing— I'm not sure if it changed or if I missed it the first time— is that the fee does not "stick" when closing the form to edit it. When re-opening the form, the custom fee is retained, but when closing it, it reverts:

Umbra-100-Gwei-Flip-Flop-Submission

mds1:
> One thing I noticed while testing— I'm not sure if it changed or if I missed it the first time— is that the fee does not "stick" when closing the form to edit it. When re-opening the form, the custom fee is retained, but when closing it, it reverts:

Ah, this was intentional. @davidlaprade and I thought that this approach made sense to handle the use case where a user might change their mind about a custom fee and just revert back to the default fee. If we wanted the fee to "stick" after clicking the X, then you'd either need a "reset" button or they'd have to close and re-open the modal to do so.

A use case for resetting is you accidentally enter a value that's too large or too small, and want to reset it because you don't know how check gas prices / or it's faster to do it this way / or you trust the default more than yourself

apbendi:
> Ah, this was intentional. @davidlaprade and I thought that this approach made sense to handle the use case where a user might change their mind about a custom fee and just revert back to the default fee. If we wanted the fee to "stick" after clicking the X, then you'd either need a "reset" button or they'd have to close and re-open the modal to do so.

> A use case for resetting is you accidentally enter a value that's too large or too small, and want to reset it because you don't know how check gas prices / or it's faster to do it this way / or you trust the default more than yourself

That makes sense, but was pretty un-intuitive, at least for my cave man brain. I clicked the "x" expecting to have that fee reflected. Perhaps we can make it more explicit that closing the edit form will throw away the custom fee— for example, what if we change the "x" to a button that said "Cancel"?

PR #219, opened 2021-08-13
Improve notifications

### LINKED ISSUE #213: Remove Blocknative Toasts
Replace them with a homegrown system that simply waits for data from the chain to show the tx has been mined. Can be a toast or other indicator of some kind. This PR is needed for L2 deployments, but can be merged to `master` and deployed out of band with the Polygon project.

mds1:
Closes #213

I ended up sticking with Blocknative's notify.js because they have have a "UI Only" mode where you can control it manually like a generic notification library. I think their toasts look nice, they're likely familiar to many users, and it's a lot less work than building our own notification system, so using this seems like a great option.

As a result, there's two changes to this PR:
1. Transaction notifications: These use the same Blocknative notify.js library but now in "UI Only" mode. This looks and feels identical to users, and we handle the notification status manually based on the connected provider within the `txNotify` method. You can also click a notification and it will open the Etherscan link
2. Other notifications: All other notifications now use the same notify.js library, instead of the Quasar notify library, and have the same look and feel as transaction notifications.

To test:
1. Send any types of transaction you like to see the "pending" and "success" notifications
2. Reject a transaction to see the "error" notification
3. In Account Setup, sign the message, then go back a step and sign again to see a "hint" style notification

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -407,7 +407,7 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
         const { gasPrice } = options;
         tx = await umbra.value.withdraw(spendingPrivateKey, token.address, acceptor, { gasPrice });
         txHashIfEth.value = tx.hash;
-        txNotify(tx.hash);
+        txNotify(tx.hash, provider.value);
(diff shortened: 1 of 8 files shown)


apbendi (approved):
Looks and works great. Nice 👏 👏 👏

PR #220, opened 2021-08-15
Improve Mobile Browser Responsiveness

radchukd:
https://gitcoin.co/issue/ScopeLift/umbra-protocol/218/100026261

### DIFF
--- a/frontend/src/components/WithdrawForm.vue
+++ b/frontend/src/components/WithdrawForm.vue
@@ -50,8 +50,12 @@
       <div @click="emit('togglePrivateKey')" class="text-caption hyperlink q-mt-lg">
         {{ spendingPrivateKey ? 'Hide' : 'Show' }} withdrawal private key
       </div>
-      <div v-if="spendingPrivateKey" @click="emit('copyPrivateKey')" class="cursor-pointer copy-icon-parent q-mt-sm">
-        <span class="text-caption">{{ spendingPrivateKey }}</span>
+      <div
+        v-if="spendingPrivateKey"
+        @click="emit('copyPrivateKey')"
+        class="text-caption text-break-word cursor-pointer copy-icon-parent q-mt-sm"
+      >
+        <span>{{ spendingPrivateKey }}</span>
         <q-icon class="copy-icon" name="far fa-copy" right />
       </div>
     </div>
--- a/frontend/src/components/WithdrawForm.vue
+++ b/frontend/src/components/WithdrawForm.vue
@@ -1,11 +1,12 @@
 <template>
-  <q-form @submit="this.$emit('initializeWithdraw')" class="form-wide q-pa-md" style="white-space: normal">
+  <q-form @submit="emit('initializeWithdraw')" class="form-wide q-pa-md" style="white-space: normal">
     <!-- Withdrawal form -->
     <div v-if="!isWithdrawInProgress">
       <div>Enter address to withdraw funds to</div>
       <base-input
-        v-model="destinationAddress"
-        @click="this.$emit('initializeWithdraw')"
+        v-model="content"
+        @input="emit('updateDestinationAddress', content)"
+        @click="emit('initializeWithdraw')"
         appendButtonLabel="Withdraw"
         :appendButtonDisable="isWithdrawInProgress || isFeeLoading"
         :appendButtonLoading="isWithdrawInProgress"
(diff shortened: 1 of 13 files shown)


apbendi (changes requested):
Hey @radchukd, overall this is looking great! Thanks so much for picking this up and turning it around so quickly. Really happy with the way it's looking :)

Just did a QA run through the app and found a few regressions that seem to have come along with the changes:

### Receive

Withdrawal seems to have been broken. When I click on "Withdraw" I get the following error toast, with the corresponding console error.

Screen Shot 2021-08-16 at 10 52 28 PM

Screen Shot 2021-08-16 at 10 52 35 PM

### Setup

After I select a name, it does not advance to the next screen, and the copy says "click the arrow" which is not visible in mobile mode. One fix would be a copy change: "click the arrow or swipe left". A better solution would be to advance to the next step automatically, but I don't know why we don't do this already. Do you remember @mds1?

Screen Shot 2021-08-16 at 10 46 48 PM

Similarly, after I sign in step 2, it does not advance  me to the next step. This only becomes evident when I tap the sign button again and get the following toast:

Screen Shot 2021-08-16 at 10 46 27 PM

If I swipe to continue at this point, things do work in the next step, so I think we just need to advance to the next step after the signature is collected.

apbendi:
A couple other minor pieces of feedback:

I like the menu, but I wonder if we can afford to make the font a *little bit* bigger and give the links a little more padding.

Also, I'm not sure how I feel about the wallet indicator being in the menu, but I also don't have a better idea. Any thoughts @mds1?

Screen Shot 2021-08-16 at 11 01 20 PM

mds1:
Like @apbendi said, this is looking great overall, so thanks @radchukd! Below are my replies to Ben, then some additional thoughts from me. Also I did not look at the code yet but will check that out today

------------------------

> Withdrawal seems to have been broken. When I click on "Withdraw" I get the following error toast, with the corresponding console error.

Related to this, there's some errors in the console when I scan for funds

image


>  A better solution would be to advance to the next step automatically, but I don't know why we don't do this already. Do you remember @mds1?

We do currently do this. It must have broke somehow. I haven't looked at the code changes yet though so not sure offhand how it happened

> Also, I'm not sure how I feel about the wallet indicator being in the menu, but I also don't have a better idea. Any thoughts @mds1?

The alternative would be to hide the "Umbra" text in the header to give more space for the address. I think I'd prefer doing that so it's easy to make sure you are connected to the right wallet

------------------------

A few other UI thoughts:
1. Can we left-align the footer on mobile? Centering looks a little awkward
2. Mobile cards for the receive table look awesome. Two small thoughts:
    1. On mobile we should use the `formatAddress` helper to the address can be on the same line as "From"
    2. The "Received" text has nothing associated with it. I'm guessing "Received" and "From" were each supposed to be rows?

image

mds1:
Also, in case you didn't get yourself setup to test sending and receiving, the easiest way is just connect your wallet to Rinkeby, go through the Account Setup steps, then send some ETH to yourself and withdraw it back your wallet

frontend/src/components/WithdrawForm.vue
```diff
@@ -0,0 +1,107 @@
+<template>
```
mds1, line 1: Ah a new component! Totally fine pulling things out into this (that component was getting too big anyway), but it seems something isn't being passed right here cause the receive/withdrawal issues mentioned above

frontend/src/layouts/AddressSettings.vue
```diff
@@ -0,0 +1,39 @@
...
+import { defineComponent } from '@vue/composition-api';
+import ConnectWallet from 'components/ConnectWallet.vue';
+
+export default defineComponent({
+  name: 'AddressSettings',
+  components: { ConnectWallet },
+  props: {
+    userDisplayAddress: {
+      type: String,
+      required: false,
```
mds1, line 27: I think this should be required, but with `null` or `undefined` as acceptable types

frontend/src/layouts/HeaderLinks.vue
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api';
+
+export default defineComponent({
+  name: 'HeaderLinks',
+
+  props: {},
```
mds1, line 21: Nit, but we can remove these two lines

mds1:
Just finished looking over the code and it looks really good! Just added a few small comments above

radchukd:
Hi, @apbendi @mds1. I am glad to hear that you like it. Thanks for all the feedback and for the advice on setting up Rinkeby.
I have just pushed a commit, which addresses all of the issues you have mentioned.
Please, let me know if it's all good.
Thanks.

frontend/src/layouts/AddressSettings.vue
radchukd: But, what if the user has not connected their wallet yet?

mds1: You'll see in `wallet.ts` where `userAddress` is initialized/defined, and it defaults to `undefined` if a user has not connected their wallet yet. So the prop would still be passed, just with a value of undefined

mds1 (approved):
Everything is looking and working great now, thank you @radchukd for the quick fixes! I'll wait for @apbendi to approve also before rebasing + merging

apbendi (approved):
Great job @radchukd. This looks good to me. @mds1 can you handle the rebase and merge? Eager to get this deployed!

mds1:
Just pushed one small commit here to handle private key wrapping in advanced mode, then merged!

@radchukd Thank you for the great work here 🙌

Issue #211, opened 2021-08-03
Add Network Selector


apbendi:
Add a network selector similar in nature to the one used by Tornado Cash. It should live next to the user's wallet address and display the current network. It should have some UI caret or other indicator to show it can be expanded. When clicked, it should have a drop down or modal that lists the available networks, including which is currently selected.

<img width="273" alt="Screen Shot 2021-08-02 at 10 00 14 PM" src="https://user-images.githubusercontent.com/1479802/127946672-2546b2b7-76e2-476e-8680-8293429ee436.png">

<img width="323" alt="Screen Shot 2021-08-02 at 10 00 22 PM" src="https://user-images.githubusercontent.com/1479802/127946677-563d4458-9148-4a79-adcd-d64421f4341e.png">

The networks listed should include, for now:

* Ethereum Mainnet
* Polygon Mainnet
* Rinkeby Testnet
* Mumbai Testnet

When the user selects a network, it should prompt them via their wallet to add/switch to the appropriate network, and reset the rest of the frontend state as needed.

mds1:
Based on the above, I think the full spec of this issue would be:
- For now, support Mainnet and Rinkeby only
- Clicking a network should use EIP-3085’s `wallet_addEthereumChain` (when it’s available) to prompt the user to switch the network
- Should interop with the SupportedChainIds type/variable defined in `frontend/src/components/models.ts` so we can add new supported networks by defining them in one place (or two if we need to define a type + variable separately)

radchukd:
Addressed in #224

PR #224, opened 2021-08-20
Add Network Selector

radchukd:
Addresses #211

### DIFF
--- a/frontend/src/components/models.ts
+++ b/frontend/src/components/models.ts
@@ -13,9 +13,47 @@ export interface MulticallResponse {
   returnData: string[];
 }
 
+export type Chain = {
+  chainId: string;
+  chainName: string;
+  nativeCurrency: {
+    name: string;
+    symbol: string;
+    decimals: 18;
+  };
+  rpcUrls: string[];
+  blockExplorerUrls?: string[];
+  iconUrls?: string[];
+};
+export const supportedChains: Array<Chain> = [
+  {
+    chainId: '0x1',
+    chainName: 'Ethereum Mainnet',
+    nativeCurrency: {
+      name: 'Ethereum',
+      symbol: 'ETH',
+      decimals: 18,
+    },
+    rpcUrls: ['https://mainnet.infura.io/v3/INFURA_API_KEY'],
+    blockExplorerUrls: [],
+    iconUrls: [],
+  },
+  {
+    chainId: '0x4',
+    chainName: 'Ethereum Testnet',
+    nativeCurrency: {
+      name: 'Ethereum',
+      symbol: 'ETH',
+      decimals: 18,
+    },
+    rpcUrls: ['https://rinkeby.infura.io/v3/INFURA_API_KEY'],
+    blockExplorerUrls: [],
+    iconUrls: [],
+  },
+];
 // Set comprised of intersection of Chain IDs present for all contracts in src/contracts, supported by umbra-js, and by relayer
 export type SupportedChainIds = '1' | '4'; // strings for indexing into JSON files
-export const supportedChainIds = [1, 4]; // numbers for verifying the chainId user is connected to
+export const supportedChainIds = supportedChains.map((chain) => Number(chain.chainId)); // numbers for verifying the chainId user is connected to
 
 // CNS names owned by wallet are queried from The Graph, so these types help parse the response
 type CnsName = { name: string };
--- a/frontend/src/components/models.ts
+++ b/frontend/src/components/models.ts
@@ -13,6 +13,7 @@ export interface MulticallResponse {
   returnData: string[];
 }
 
+// Spec: https://eips.ethereum.org/EIPS/eip-3085#specification
 export type Chain = {
   chainId: string;
   chainName: string;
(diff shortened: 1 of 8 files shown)


frontend/src/components/models.ts
```diff
@@ -13,9 +13,47 @@ export interface MulticallResponse {
...
+  nativeCurrency: {
+    name: string;
+    symbol: string;
+    decimals: 18;
+  };
+  rpcUrls: string[];
+  blockExplorerUrls?: string[];
+  iconUrls?: string[];
+};
+export const supportedChains: Array<Chain> = [
```
mds1, line 28: I like this layout for defining support chains. A few comments though:
- Can we make `rpcUrls`, `blockExplorerUrls`, and `iconUrls` from plural/arrays into singular/strings? Not sure I see the reasoning for making them arrays at the moment
- Can we add a `logoURI: string;` field to the `nativeCurrency` object? This way we have the token symbol of the native currency also. For ETH we can use `'/tokens/eth.svg'` as the URI

frontend/src/layouts/NetworkDropdown.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <q-btn-dropdown
+    color="primary"
+    no-caps
```
mds1, line 4: Perhaps make this button a little more subtle
```suggestion
    color="primary"
    no-caps
    outline
    rounded
```

frontend/src/components/models.ts
```diff
@@ -13,9 +13,47 @@ export interface MulticallResponse {
...
+  {
+    chainId: '0x1',
+    chainName: 'Ethereum Mainnet',
```
mds1, line 31:
```suggestion
    chainName: 'Mainnet',
```

frontend/src/components/models.ts
```diff
@@ -13,9 +13,47 @@ export interface MulticallResponse {
...
+      symbol: 'ETH',
+      decimals: 18,
+    },
+    rpcUrls: ['https://mainnet.infura.io/v3/INFURA_API_KEY'],
+    blockExplorerUrls: [],
+    iconUrls: [],
+  },
+  {
+    chainId: '0x4',
+    chainName: 'Ethereum Testnet',
```
mds1, line 43:
```suggestion
    chainName: 'Rinkeby',
```

frontend/src/components/models.ts
```diff
@@ -13,9 +13,47 @@ export interface MulticallResponse {
...
+export type Chain = {
+  chainId: string;
+  chainName: string;
+  nativeCurrency: {
+    name: string;
+    symbol: string;
+    decimals: 18;
+  };
+  rpcUrls: string[];
+  blockExplorerUrls?: string[];
```
mds1, line 25: Would be great if the `blockExplorerUrl: string` field was directly used within the `getEtherscanUrl()` method in the `utils.ts` file! That way when adding a network we can just do it all in this file

frontend/src/utils/utils.ts
```diff
@@ -15,6 +15,20 @@ export const getEtherscanUrl = (txHash: string, chainId: number) => {
   return `https://${networkPrefix}/tx/${txHash}`;
 };
 
+/**
+ * @notice Generates short names based on the given `chainId`
+ */
+export const getChainShortName = (chainId?: number) => {
```
mds1, line 21: Can this method just return the `chainName` from the `supportedChains` list? Similar to the Etherscan comment above—that way when adding a network we can just do it all in the one file, and these methods will pull from there

Also maybe rename to just getChainName since we'll pull directly from that list

frontend/src/layouts/NetworkDropdown.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <q-btn-dropdown
```
mds1, line 2: This format of `<q-btn>` and `<q-list>` works, but you could also use our own `<base-select>` component which might be a little cleaner / less code here.

mds1 (changes requested):
Loving this, thank you @radchukd!

------------------------

Main thing to update (as mentioned in the comments) is to have the `getEtherscanUrl` and `getChainShortName` methods pull form the `supportedChains` array, so everything can be defined in one spot. Currently adding a new network would require adding something in a few spots

It also seems the dropdown items aren't centered vertically as shown below. One easy way to fix this would be to use the `BaseSelect` component we have instead of implementing the dropdown yourself

image

frontend/src/layouts/NetworkDropdown.vue
mds1: Or if you switch to the `BaseSelect` component `:outline="true"` and `:rounded="true"`, but it seems we'd have to add a prop for `rounded` as it's not currently in the `BaseSelect` component

frontend/src/components/models.ts
radchukd: I took this interface out of [EIP](https://eips.ethereum.org/EIPS/eip-3085#specification).

mds1: Oh! Didn't realize that. Ok, fine with me to leave this as-is then 👌

mds1: Maybe throw in a comment with a link to that EIP if you don't mind!

frontend/src/utils/utils.ts
radchukd: On [Tornado](https://app.tornado.cash/) they have short name in the selected field and long names in dropdown.
If we keep the names short, there wouldn't be a need for this method at all.

mds1: Good data point. I was looking at https://gnosis-safe.io/app/#/ which only uses short names. 

Personally I think the short names are sufficiently descriptive and I'd be ok removing this

mds1: Maybe we wait to see what @apbendi thinks

radchukd:
I have remade this with BaseSelect and width shouldn't be a problem now, so I have removed getChainShortName.

mds1:
Thanks! Just pushed two small commits to cleanup some styling on the network selector, and make sure the chainId is properly converted from decimal to hex. LGTM, just waiting on @apbendi to review before merging

mds1 (approved):

apbendi (changes requested):
Overall this looks awesome @radchukd! Great work, and thanks for tackling it so quickly. Here are a few things that came up in my QAing:

1. When testing in a mobile-size (and, weirdly, only in mobile size) I got the following error when switching between networks:

Screen Shot 2021-08-20 at 1 35 15 PM

Screen Shot 2021-08-20 at 1 35 26 PM

2. When the browser is in a slightly smaller size, the selector is truncated. In general, the selector seems a bit wide, even in browser sizes where it is not truncated.

Screen Shot 2021-08-20 at 1 33 12 PM

3. Right now when the user is on an unsupported network, the selector is not shown. Not sure how hard this would be, but I think a better flow would be to show the selector with "⚠️ Unsupported" shown, then allow the user to select the drop down and select a valid network from the list. Thoughts?

Screen Shot 2021-08-20 at 1 34 26 PM

radchukd:
Everything should be fine now.
I have also added isLoading ref to the wallet store, so the network wouldn't be unsupported(undefined) at first, instead, there would be a spinner.

apbendi (approved):
Looking and working great @radchukd. Awesome work. Thanks!

PR #225, opened 2021-08-20
Implement deployment script for StealthKeyRegistry

apbendi:


contracts/hardhat.config.ts
```diff
@@ -78,7 +78,7 @@ const config: HardhatUserConfig = {
         count: 10,
         initialIndex: 0,
         mnemonic,
-        path: "m/44'/60'/1'/0",
+        path: "m/44'/60'/0'/0",
```
mds1, line 81: Didn't we use the old path of `"m/44'/60'/0'/0"` for the original deploys? If so this feels like something that should be conditional (env var?) based on which deploy script you're using

contracts/scripts/deploy-registry.js
```diff
@@ -0,0 +1,95 @@
...
+ * @dev To initialize a deploy:
+ *   `yarn deploy:registry --network <network>`   (where network specifies a netwrk found in the hardhat.config.ts file)
+ * If deploying to a local node (--network localhost), first start Hardhat using `yarn hardhat node`
+ */
+const fs = require('fs');
+const hre = require('hardhat');
+const { exit } = require('process');
+const { ethers } = hre;
+
+const network = process.env.HARDHAT_NETWORK;
```
mds1, line 15: Is this identical to `const network = hre.network.name`? I think so, but not as familiar with this approach so just confirming

PR #227, opened 2021-09-06
Add connect wallet button

radchukd:
I fixed wallet loading and added the "Connect a wallet" button along with some refactoring.

### DIFF
--- a/frontend/src/layouts/BaseLayout.vue
+++ b/frontend/src/layouts/BaseLayout.vue
@@ -20,15 +20,24 @@
                     <span class="primary header-black q-ml-md">Umbra</span>
                   </div>
                   <div v-else-if="$q.screen.xs">
+                    <div v-if="isLoading" class="q-ml-md row">
+                      <q-spinner color="primary" size="1em" />
+                    </div>
                     <address-settings
-                      v-if="!isLoading"
+                      v-else-if="!isLoading && userDisplayAddress"
                       :userDisplayAddress="userDisplayAddress"
                       :advancedMode="advancedMode"
                       class="q-ml-md row"
                     />
-                    <div v-else class="q-ml-md row">
-                      <q-spinner color="primary" size="1em" />
-                    </div>
+                    <connect-wallet v-else-if="!isLoading">
+                      <base-button
+                        class="q-ml-md cursor-pointer"
+                        color="primary"
+                        :dense="true"
+                        label="Connect a wallet"
+                        :outline="true"
+                      />
+                    </connect-wallet>
                   </div>
                 </div>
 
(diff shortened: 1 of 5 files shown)


apbendi (commented):
Hey @radchukd, looking good! Nice work.

One thing I did notice is clicking on my address after connecting, which used to allow the user to change wallets, now produces an error:

Screen Shot 2021-09-07 at 7 28 21 AM

Screen Shot 2021-09-07 at 7 28 35 AM

frontend/src/store/wallet.ts
```diff
@@ -136,9 +159,42 @@ export default function useWalletStore() {
     rawProvider.value = p;
   }
 
+  function setLoading(l: boolean) {
+    isLoading.value = l;
```
mds1, line 163: A nit, but would prefer a more meaningful parameter name than `l`, maybe `status`?

frontend/src/store/wallet.ts
```diff
@@ -219,18 +278,23 @@ export default function useWalletStore() {
...
   async function setNetwork(chain: Chain) {
+    setLoading(true);
+
     try {
       await provider.value?.send('wallet_switchEthereumChain', [{ chainId: chain.chainId }]);
     } catch (error) {
       // This error code indicates that the chain has not been added to MetaMask.
-      if (error.code === 4902) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (((error as never) as any).code === 4902) {
```
mds1, line 288: Why do we need this new casting?

mds1: Maybe better to do `} catch (error as {code: string}) {`

frontend/src/layouts/BaseLayout.vue
```diff
@@ -20,15 +20,15 @@
                     <span class="primary header-black q-ml-md">Umbra</span>
                   </div>
                   <div v-else-if="$q.screen.xs">
+                    <div v-if="isLoading" class="q-ml-md row">
+                      <q-spinner color="primary" size="1em" />
+                    </div>
```
mds1, line 25: My localStorage has `__q_strn|MetaMask` as the value for the `last-wallet` key, but when testing out this PR it gets stuck with `isLoading = true`, and there's a few (possibly related) errors in the console. Some screenshots below (might need to open the image in a new tab to see the full resolution)

image

image

image

radchukd: What exactly did you do?

frontend/src/store/wallet.ts
radchukd: New casting is needed because of `Object is of type 'unknown'.ts(2571)`

`error as {code: string}` This doesn't work because of `Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)` https://github.com/microsoft/TypeScript/issues/20024.

I fixed this casting like this:
```
 catch (switchError) {
      const { code } = switchError as { code: number };
      if (code === 4902) {}
}
```

frontend/src/layouts/BaseLayout.vue
mds1: I just built the app locally like normal. Also just noticed it seems my screenshot of the header was poorly timed as its from the small portion of the spinner cycle when there is no spinner 🙃

> Try clearing site data.

I thought about this, but didn't want to do it, and don't think we should merge this PR if it requires clearing site data—the reason is because if you need to clear site data for it to work properly, then the live app will break for users once deployed and we don't want to make them clear their data

radchukd: Yep, you're right. I think I got it now.

mds1: Yes, seems to be working now, thank you!

mds1 (approved):
LGTM, thanks @radchukd!

@apbendi I'll wait for your final review before merging

PR #230, opened 2021-09-21
fix from format in card

### LINKED ISSUE #232: Network Selector Never Loads When On Unsupported Network
If the user's wallet is on an unsupported network, the network selector no longer loads. Instead, the loading spinner continues forever. This appears to be a regression. In previous versions, the network selector would load and say "Unsupported Network", allowing the user to choose a valid network and switch. We should restore this functionality.

Screen Shot 2021-09-23 at 11 48 57 AM

radchukd:
Closes #232

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -58,7 +58,7 @@
                 </div>
               </q-card-section>
               <q-card-section class="column justify-center items-center">
-                <div>From: {{ formatAddress(props.row.tx.from) }}</div>
+                <div>From: {{ props.row.receipt.from }}</div>
                 <div class="text-caption text-grey">
                   Received: {{ formatDate(props.row.block.timestamp * 1000) }}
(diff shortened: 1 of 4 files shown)


mds1 (approved):

apbendi (changes requested):
Hey @radchukd, looking good in terms of addresses with ENS names now showing in mobile mode 👍 

However, it looks like there's been a regression in the case of showing a normal address. It is no longer using the truncated form (i.e. 0x1234...5678), leading to normal addresses overflowing the screen in mobile:

Screen Shot 2021-09-23 at 11 56 44 AM

radchukd:
Also fixes #232

mds1:
Good catch @apbendi 

Changes look good, re-approved by me! Confirmed this fixes #232 also so @radchukd I edited your PR description so the issue will be linked and closed automatically once it's merged

apbendi (approved):
Working well. Thanks @radchukd 👏

PR #233, opened 2021-09-24
Add receiver column

### LINKED ISSUE #231: Add "Receiver" Column to the Receive Screen
To help clarify how the stealth address mechanism works to our users, we want to expose a bit more of what's going on underneath the hood. One idea is to actually show the automatically generated stealth address that has received the funds (or holds the ability to withdraw them from the contract) on the receive screen.

Below is a rough mockup of what this can look like. (Please ignore fonts and spacings in the mockup).

Screen Shot 2021-09-21 at 7 15 24 AM

Requirements:

- [ ] Change the name of the "From" column to "Sender"
- [ ] Add a column on the receive screen named "Receiver"
- [ ] Add an info icon next to "Receiver", when the user hovers/taps on the icon, they should see small modal with a short explanation (text below)

> The stealth address which received these funds. It looks like an empty address, and it can't be tied to your account, but only you have the ability to generate its private key.

- [ ] Show the truncated address of the receiver, which can be sourced from the [announcements](https://github.com/ScopeLift/umbra-protocol/blob/master/umbra-js/src/classes/Umbra.ts#L321)

radchukd:
closes #231 
Add receiver column
Add clickable tooltip component
Change positioning in card
Add copying to card
Rename 'From' to 'Sender'


radchukd:
I had to add a custom tooltip component because quasar's native auto hides when the mouse leaves the emitter. With mine, you can click and select text inside of it.
I also changed the card layout a little bit, and added copying to it.

apbendi (changes requested):
Hey @radchukd, looking great overall! Nice work so far. Here's a few things I noticed:

* Sender addresses no longer seem to be truncated (i.e. 0x1234...5678) and as a result, the table scrolls horizontally. We should truncate these addresses so the full table can fit in a reasonable size screen.
* Sorting the table seems to have been enabled (when it wasn't before). This should be turned off. In particular it's not possible to click the tooltip without causing the table to sort. But we don't want the sorting anyway.

umbra-table-issues

* Also, in mobile mode, I think we want an extra space between "Stealth Receiver" and the question mark icon for the tool tip. It looks too close.

Screen Shot 2021-09-24 at 5 19 55 PM

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -284,7 +341,8 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
   const mainTableColumns = [
     { align: 'left', field: 'block', label: 'Date Received', name: 'date', sortable: true, sort: sortByTime },
     { align: 'left', field: 'amount', label: 'Amount', name: 'amount', sortable: true, format: toString },
-    { align: 'left', field: 'receipt', label: 'From', name: 'from', sortable: true },
+    { align: 'left', field: 'receipt', label: 'Sender', name: 'from', sortable: true },
+    { align: 'left', field: 'receiver', label: 'Stealth Receiver', name: 'receiver', sortable: true },
```
mds1, line 345: Like @apbendi mentioned, let's flip `sortable` to `false` here. Sorting on randomly generated stealth addresses doesn't add value so that will resolve the "clicking tooltip also sorts" problem

mds1:
> Sender addresses no longer seem to be truncated (i.e. 0x1234...5678) and as a result, the table scrolls horizontally. We should truncate these addresses so the full table can fit in a reasonable size screen.

I think might be because we need to rebase this branch against the latest `master`. You'll also notice #232 is broken on this branch, and I believe that's why

@radchukd Not sure if Ben's ever mentioned it, but we like to use rebases in stead of merging. So to rebase this branch against master you'd do:

```
git checkout master
git pull origin master
git checkout add-receiver-column
git rebase master
```

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -107,7 +138,24 @@
         <!-- Header labels -->
         <template v-slot:header="props">
           <q-tr :props="props">
-            <q-th v-for="col in props.cols" :key="col.name" :props="props"> {{ col.label }} </q-th>
+            <q-th v-for="col in props.cols" :key="col.name" :props="props">
+              {{ col.label }}
+
+              <!-- Question mark with tooltip for receriver column --->
+              <clickable-tooltip v-if="col.name === 'receiver'" icon="fas fa-question-circle">
```
mds1, line 145: What is this commented out tooltip section?

frontend/src/components/ClickableTooltip.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <q-btn-dropdown
```
mds1, line 2: Can we make the tooltip also show up on hover? I feel like hover for desktop and tap for mobile is the standard approach.

Also, if we could replace other uses of `q-tooltip` with this custom component in this PR, that way everything is consistent, that would be awesome

frontend/src/layouts/BaseLayout.vue
```diff
@@ -142,7 +142,11 @@
             <span class="dark-toggle text-caption q-ml-md">Advanced mode {{ advancedMode ? 'on' : 'off' }}</span>
             <span>
               <q-icon class="dark-toggle" right name="fas fa-question-circle">
-                <q-tooltip content-class="bg-muted dark-toggle shadow-2 q-pa-md" max-width="14rem">
+                <q-tooltip
+                  content-class="bg-muted dark-toggle shadow-2 q-pa-md"
+                  :hide-delay="$q.screen.lt.lg ? 1000 : 0"
```
mds1, line 147: Why did you add a hide delay only on small screens? I'd guess it has to do with something related to clicking vs hovering to show the tooltip

radchukd: On mobile, you had to press and hold to see the tooltip, with this tweak you can click it and it hides after delay (Pressing and holding still work).

radchukd:
I rebased the branch, but for some reason, the diff shows changes, that are already on the master.

mds1 (approved):
I just rebased to resolve conflicts. I'm not sure how, but the "fix from format in card" commit on this branch is a bit different than the one on master. It's probably ok to just leave it unless @apbendi has any ideas / wants to fix it

Either way, PR looks good to me pending a final QA review from @apbendi

apbendi (approved):
Not sure about the commit, but the changes look good. Thanks @radchukd 👍

Issue #214, opened 2021-08-04
Simplified Stealth Key Resolution Approach


apbendi:
The publishing of public keys—  those used by senders to create stealth addresses for a given sender— is not part of the core Umbra protocol. Rather, it is a periphery concern that can be handled by some convention followed by any app(s) leveraging Umbra.

### Current Approach

Currently, the Umbra app publishes these keys by associating them directly with the user's ENS or CNS name. It does this using a custom resolver (for ENS) or a text record (for CNS).

### Challenges

Through the process of development, beta testing, and early adoption, we've recognized a number of challenges with this approach:

* User Experience
    * Setup requires several transactions and is expensive
    * Sending is different than in normal dApps, in that it *requires* the sender use a domain name, when most dApps allow the user to provide a domain name OR a regular Ethereum address.
    * It's confusing: user's seem to think that by setting up Umbra, all transactions sent through that domain name become stealth protected. The relationship between the name service and Umbra is not clear.
* Development
    * There are lots of code paths and corner cases associated with supporting name services (which name service to use, what resolver the user is currently using, whether they're using their own name, or an umbra.eth subdomain, etc...)
    * There are edge cases associated with tying keys to the the name directly (what happens if the user transfers or sells the name, what happens if the owner of the name is not the same as the resolution address, etc...)
* Layer 2 / Sidechain Integrations
    * We want to deploy Umbra across multiple networks, but it's unclear how to do stealth key resolution when doing so. Presumably we could resolve strictly on the Ethereum L1, regardless of which network the user is doing the sending/receiving from, but this introduces other challenges:
        * It would be confusing for users, and force them to switch to the Ethereum L1 to setup the app before they could use it on L2
        * It creates an issue when deriving the stealth keys, which is derived by having the user sign over a human readable message *which includes the chainId*.

### Alternative Approach

Instead of attaching the user's stealth keys directly to their ENS/CNS name, we can instead deploy a "StealthKeyRegistry" contract. The contract would:

* Provide a mapping between an address and the address' stealth/viewing keys
* Only allow the owner of said address to update the keys

An app supporting Umbra could then:

* Allow a user to send via Umbra by providing a simple Ethereum address, rather than an ENS/CNS name. The app would look up the address in the registry to find its stealth keys, and notify the sender if the address had not yet been configured.
[...]

PR #234, opened 2021-09-29
Stealth key registry

### LINKED ISSUE #226: Implement app / umbra-js changes to support new StealthKeyRegistry
This issue will require changes to both the Umbra frontend and `umbra-js` to support the new StealthKeyRegistry system. The registry has been deployed to [rinkeby](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/deploy-history/registry-rinkeby-latest.json) and [mainnet](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/deploy-history/registry-mainnet-latest.json).

#### Legacy User Detection

- [x] Legacy users, i.e. user's who have already gone through the current ENS/CNS stealth key resolution process, should be detected and shown a banner or message across the top of the page.
- [x] It should show a message like, "We've upgraded our key resolution system. Please submit one transaction to update your configuration. You won't be able to receive Umbra transactions until you've done so." The message should be a link which takes them to the `/setup`. It can also have a "Learn More" link afterwards which takes them to a yet-to-be-written FAQ question about the upgrade.

#### Determining Prior Setup & Fetching Public Stealth Keys

- [x] Call the Registry's [stealthKeys](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/contracts/StealthKeyRegistry.sol#L169) view method, with the connected address, and check if the `spendingKey` and the `viewingKey` returned are non-zero. If either (or both, as is the case of normal usage) are zero, then assume the user has not completed setup
- [x] As long as both keys are non-zero, use the keys + prefixes returned as a given address' stealth keys

#### Setup Flow

- [x] The updated setup flow should have only two steps:
     1. Sign the message to generate public/private keypair
     2. Submit transaction to the Registry's [setStealthKeys](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/contracts/StealthKeyRegistry.sol#L70) method

#### Sending

- [x] The "Send" page should now accept a regular Ethereum address OR an ENS/CNS name
    - [x] Perform validation on the address (checksum)
- [x] If the user provides an ENS/CNS name, first resolve the name to an address
- [x] Take the address (directly provided or resolved) and use it to determine if this address has completed setup (as described above)
    - [x] If the user has not registered their stealth keys, notify the sender that this address has not yet configured Umbra
    - [x] If the user has registered their stealth keys, use them as dictated by the protocol to generate and encrypt the stealth address for this transaction
- [x] If the user has toggled "Advanced Mode" on, there should be a checkbox below the recipient input which says "Send using recipient's Ethereum public key".
     - [x] When selected, this should map to the existing advanced mode functionality that occurs when a user has advanced mode and supplies a normal Ethereum address. Namely, it should use the recovered public key from a past transaction from the supplied address to server as the spending & viewing keys for the stealth transaction.
     - [x] It might be nice to also include a "learn more" link next to this checkbox which takes the user to an appropriately modified version of [this faq](https://app.umbra.cash/faq#how-do-i-send-funds-to-a-user-by-their-address-or-public-key) question explaining what they're doing

#### umbra-js

- [x] The entire DomainService class can probably be deprecated and ultimately remove, but for the current version, we probably want to keep it in for the sake of detecting legacy users, as described above
- [ ] All the ens & cns util functions can be deprecated and ultimately removed in the same fashion
- [ ] The ForwardingStealthKeyResolver, ENSRegistry, and CNSRegistry ABI's can all be deprecated and ultimately removed.
- [x] The StealthKeyRegistry ABI will have to be added
- [x] A new class for interacting with the `StealthKeyRegistry` is probably warranted. This can be a pretty simple convenience wrapper around the ethers contract instance, with an escape hatch to access the underlying instance
- [x] The `lookupRecipient` function will have to be updated to remove usage of DomainService and switch to an instance of the new StealthKeyRegistry convenience class

mds1:
Closes #226

See commit messages for details


frontend/src/layouts/BaseLayout.vue
```diff
@@ -148,13 +155,6 @@
               </base-tooltip>
             </span>
           </p>
-
-          <p v-if="isAccountSetup">
-            <q-icon class="dark-toggle icon" name="fas fa-user q-mr-xs" />
-            <span class="dark-toggle text-caption q-ml-sm">
-              Repeat <router-link class="hyperlink" :to="{ name: 'setup' }">account setup</router-link>
```
mds1, line 155: Now that the stealth key registration is much simpler and you can no longer configure multiple ENS names, there's no reason to repeat account setup once completed, so I removed this from the footer

apbendi:
When we show the user a toast about the receiver not having their account set up, it overflows the toast:

Screen Shot 2021-10-05 at 7 28 35 AM

A similar thing happens when the user enters an address with a bad checksum:

Screen Shot 2021-10-05 at 7 28 59 AM

When the user enters something that is too short to be an address, we show the error inline immediately after

Screen Shot 2021-10-05 at 7 29 11 AM

I'm guessing we don't do this with stealth key resolution or with checksumming because those are async checks? Still, I wonder how hard it would be to run the async checks after the user stops input for (say) 250 ms, then show the error message inline like we do in the third case?

If that's not a quick fix, we can consider it as a future enhancement. For now, at a minimum, we should make the toast error readable.

frontend/src/pages/AccountSend.vue
```diff
@@ -15,10 +15,40 @@
...
+      <!-- Identifier, advanced mode tooltip -->
+      <q-checkbox
+        v-if="advancedMode"
+        v-model="useNormalPubKey"
+        class="text-caption q-pt-sm q-pb-lg"
+        dense
+        :style="!recipientId || isValidId(recipientId) === true ? 'margin-top:-2em' : ''"
+      >
+        Send using recipient's standard public key
+        <base-tooltip icon="fas fa-question-circle">
```
apbendi, line 31: The copy in this tooltip is thorough, but I'm afraid no one will ever read it.

Screen Shot 2021-10-05 at 7 29 49 AM

What do you think of going with something like: "When checked, the public key used will be the one underlying the user's Ethereum address. The receiver will have to manipulate their account's private key to withdrawal the funds. **Don't use this feature unless you know what you're doing**. Learn More.

Also, I wonder if the checkbox should simply be checked and disabled (i.e. greyed out and not clickable) when the user enters a tx hash or public key? Alternatively, I wonder if we should just remove that feature even in advanced mode?

Finally, if I tap the tooltip icon on mobile, which is required to show the contents, it toggles the check mark. It should not do this.

frontend/src/pages/AccountSend.vue
```diff
@@ -36,7 +66,13 @@
...
-        <base-button :disable="isSending" :full-width="true" label="Send" :loading="isSending" type="submit" />
+        <base-button
+          :disable="!isValidForm || isSending"
+          :full-width="true"
+          label="Send"
+          :loading="isSending"
+          type="submit"
+        />
         <base-button
           @click="generatePaymentLink({ to: recipientId, token, amount: humanAmount })"
```
apbendi, line 77: Payment links with amounts seem to be broken. Try visiting http://localhost:8080/send?to=benrink.eth&token=ETH&amount=1 as an example. The console shows an error message:

Screen Shot 2021-10-05 at 1 22 56 PM

frontend/src/layouts/BaseLayout.vue
```diff
@@ -76,10 +82,17 @@
...
-          This is beta software. Use at your own risk.
-          <router-link class="hyperlink" :to="{ path: '/faq#security' }">Learn more</router-link>
+        <!-- Legacy warning -->
+        <div
+          v-if="!isAccountSetup && isAccountSetupLegacy"
+          class="dark-toggle text-center text-bold q-my-md q-pa-md"
+          style="border-radius: 15px"
+          :style="isDark ? 'color: #FFFAEB; background-color: #7c5e10' : 'color: #513C06; background-color: #fcefc7'"
+        >
+          We've upgraded our name resolution system. Use the
```
apbendi, line 92: This looks good, but I'm afraid it's easy to ignore.

Screen Shot 2021-10-05 at 1 48 16 PM

I wonder if we should use a red background and text, and add some 🚨🚨🚨 to it for good measure 😅

frontend/src/pages/AccountSend.vue
```diff
@@ -15,10 +15,40 @@
     <!-- Send form -->
     <q-form v-else @submit="onFormSubmit" class="form" ref="sendFormRef">
       <!-- Identifier -->
-      <div v-if="!advancedMode">Recipient's ENS or CNS name</div>
+      <div v-if="!advancedMode">Recipient's ENS name or address</div>
       <div v-else>Recipient's ENS name, CNS name, transaction hash, address, or uncompressed public key</div>
       <base-input v-model="recipientId" :disable="isSending" placeholder="vitalik.eth" lazy-rules :rules="isValidId" />
 
+      <!-- Identifier, advanced mode tooltip -->
+      <q-checkbox
```
apbendi, line 23: It doesn't seem to be possible to use this feature unless the receiving address has set up Umbra, which goes counter to the purpose of this power user feature. I think we'd want to disable the check when the user has this checked.

Screen Shot 2021-10-05 at 1 54 56 PM

Screen Shot 2021-10-05 at 1 54 09 PM

apbendi: Actually, this feature seems not to work at all. I sent 201 DAI to benrink.eth with the checkbox checked, and they went to the stealth registry key instead.

frontend/src/layouts/BaseLayout.vue
```diff
@@ -76,10 +82,17 @@
...
+          <router-link class="hyperlink" :to="{ name: 'setup' }">Setup</router-link> page to submit one transaction
+          which migrates you to the updated system. You won't be able to receive Umbra transactions until you've done
+          so. <router-link class="hyperlink" :to="{ path: '/TODO' }">Learn more</router-link>.
```
apbendi, line 95: Copy for the new question:

"Why do I need to setup my account again?"

If you configured your Umbra account while the service was still in Beta— before October 2021— you need to redo setup. This requires one signature and one transaction.

This additional setup step is due to an upgrade to Umbra's stealth key registry system. The old system relied on records placed on the resolver of the ENS record associated with a user's address. The new system is much simpler, and uses a simple mapping between the user's address and their stealth keys.

The new system was developed in response to user feedback during Umbra's Beta period. The old system required multiple transactions, and also required senders use the receiver's ENS name, which caused confusion. The new system requires only a single transaction, and senders can provide either the receiver's ENS name or their regular Ethereum address. Overall, the new system is simpler both to configure and use.

For an even more in depth and technical explanation of this upgrade, see [this issue](https://github.com/ScopeLift/umbra-protocol/issues/214) on Umbra's Github repository.

frontend/src/pages/AccountSend.vue
mds1: Yea we can shorten the copy to your version.

> Also, I wonder if the checkbox should simply be checked and disabled (i.e. greyed out and not clickable) when the user enters a tx hash or public key? 

 I agree it should be disabled when user enters a public key, but entering a transaction hash raises a good question—when a transaction hash is entered, we just derive the public key directly from the transaction parameters, but alternatively we could take the from address and look it up in the registry if this box is unchecked, so should we? I learn towards no, because that seems like a useless use case, i.e. transaction hashes _always_ use the standard pubkey of the sender

> Alternatively, I wonder if we should just remove that feature even in advanced mode?

I have a strong preference towards keeping this feature, because we want the "anonymity set" of recipients to be as large as possible. Removing this makes is significantly more likely that only users in the StealthKeyRegistry are using Umbra, which shrinks the anonymity set

> Finally, if I tap the tooltip icon on mobile, which is required to show the contents, it toggles the check mark. It should not do this.

Fixed in 9ce8cbe

frontend/src/layouts/AddressSettings.vue
```diff
@@ -1,10 +1,11 @@
...
     <connect-wallet>
-      <span v-if="userDisplayAddress" class="text-caption cursor-pointer dark-toggle">
-        {{ userDisplayAddress }}
+      <span v-if="userDisplayName" class="text-caption cursor-pointer dark-toggle">
+        {{ userDisplayName }}
       </span>
     </connect-wallet>
+    <Jazzicon v-if="userAddress" :address="userAddress" class="q-ml-sm" />
     <span v-if="advancedMode" class="q-ml-md">
       <base-tooltip label="🧙">Advanced mode is on</base-tooltip>
```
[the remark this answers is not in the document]
mds1: Ah yea it did get smaller when we migrated to our custom tooltip component. Moved to the left and made it bigger in 9d6bf15eb3bd8e6c37281838c3f36b7051a45431

frontend/src/pages/AccountSend.vue
mds1: Whoops, big bug! Should be fixed in e9828e6d918af440d680c05edda38f5ff6636ecc, let me know if it's still not working as expected

mds1: Whoops, big bug! Should be fixed in e9828e6d918af440d680c05edda38f5ff6636ecc, let me know if it's still not working as expected

frontend/src/layouts/BaseLayout.vue
mds1: Added in 1b76a6082a9698be7b2d4695c301c5c899509aa4

mds1:
> When we show the user a toast about the receiver not having their account set up, it overflows the toast:
>
> A similar thing happens when the user enters an address with a bad checksum:

Hmm, I guess the easiest solution here might be to make the toasts wider? Alternatively we could truncate the addresses, e.g. `0x1234....abcd` but that might be inconvenient, plus it seems like other non-address parts are cut off too

> When the user enters something that is too short to be an address, we show the error inline immediately after
> 
> I'm guessing we don't do this with stealth key resolution or with checksumming because those are async checks? Still, I wonder how hard it would be to run the async checks after the user stops input for (say) 250 ms, then show the error message inline like we do in the third case?
> 
> If that's not a quick fix, we can consider it as a future enhancement. For now, at a minimum, we should make the toast error readable.

Checksumming is not explicitly checked, it's handled behind the scenes by ethers when you submit the form. Ethers will only throw with an invalid checksum, and will not throw if you enter all lowercase. In practice I'd guess invalid checksums rarely happen anyway so personally I'm not worried about explicitly handling bad checksums, but it would be easy to add so maybe we should

You're right that stealth key resolution is async which is why it's validated on form submit. It seems Quasar inputs [support async rules](https://v1.quasar.dev/vue-components/input#async-rules), so it should be easy to implement. But, I didn't know it supported async rules, so it might be a newer feature that requires upgrading our quasar version so this might not be as easy as I think. I'll give it a shot tomorrow, but since this PR is already big, if it becomes a hassle we should pull it out into a separate issue

apbendi:
Ok @mds1, re-reviewed and resolved issues that looked fixed. Responded to your questions. Shared on additional bug I discovered. We're getting there! Thanks :)

mds1:
Notification alerts now are wider + dynamically check whether or not you're in dark mode, as of 61e1eeb069d7ac73b096fd70fb778d1429580c7b

mds1:
d5466b595935a94fb3a37df29c4b024fe3d87b78 should add support for async validation of recipient ID before submitting the form, but could use a bit more testing, cc @apbendi

PR #242, opened 2021-10-10
Remove connect modal for send form

### LINKED ISSUE #235: feat: don't require user to connect wallet on the send form until they click the "send" button
This way people can generate payment links without connecting a wallet, and it's just a better UX in general

radchukd:
Closes #235

### DIFF
--- a/frontend/src/components/HomeActionCard.vue
+++ b/frontend/src/components/HomeActionCard.vue
@@ -1,12 +1,12 @@
 <template>
-  <connect-wallet :to="to" style="max-width: 325px">
+  <router-link :to="to" class="no-text-decoration" style="max-width: 325px">
     <q-card class="card-border cursor-pointer q-py-md">
       <q-card-section class="text-center text-primary text-h6 header-black q-pb-none">
         {{ text }}
       </q-card-section>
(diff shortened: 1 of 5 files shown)


radchukd:
@apbendi 
Thanks, that works. Only, the first commit got lost and I recommitted the changes.

apbendi (commented):
I'm still seeing some weird states emerge here when I combine payment link + wrong network. Check out the example here:

payment-link-issues

I'm not exactly sure what the issue is but I'm starting to wonder if we need to rethink this feature? Seems like it could lead to weird edge cases, especially as we add support for other "real" networks.

It feels like we may need to:

* Keep a connected wallet as a requirement to view the send screen
* Include network in payment links
* When a user views a payment link, make sure they're connected to the right network

If we want to enable payment link generation w/o connecting a wallet, it almost seems like that might require a separate UI, one that requires the user to also select the network, then does the appropriate validation on that network.

apbendi:
Thanks for the update @radchukd— this looks reasonable, but after some internal discussion and consideration, we're going to hold off on this feature for now. Worried there still might be edgecases, and it's more of a "nice to have" rather than a "must have".

PR #251, opened 2021-11-03
Polygon support

### LINKED ISSUE #212: Network Selection Driven App Adjustments
This issue is contingent on #211. It details the changes to App display/behavior dependent on the network selection. It can be separated into multiple issues if desired.

### Receive Screen

- [x] Scan for transactions using the connected network _— already implemented_
- [x] Link to the correct blockchain explorer for txs where appropriate _— mostly implemented, FAQ page is the only place where we manually link to contract addresses_— 
- [ ] Display the correct icon / asset names based on token mapping
     - [ ] On Polygon the base asset is MATIC and should show the Polygon logo
- [ ] When withdrawing tokens, ~send to the correct relayer endpoint~ include the chainId as a parameter in the relayer request to ensure it's relayed on the correct network (note: requires an update to the relayer) _— **blocked by API updates**_

### Send Screen

- [ ] Adjust the way names are resolved so that
     - [x] When connected to Ethereum Mainnet, resolve by Mainnet ENS/CNS _— already implemented_
     - [x] When connected to Ethereum Rinkeby, resolve by Rinkeby ENS/CNS _— already implemented_
     - [ ] When connected to Polygon, resolve by *Mainnet* ENS/CNS (will check with ENS for guidance here)
- [ ] Display the correct icon / asset names, using the correct addresses
     - [x] On Polygon, the base asset is MATIC and should show the Polygon logo
     - [ ] Fetch the token list from the backend by including the chainId as a parameter, to get the network specific token list returned _— **blocked by API updates**_
     - [ ] ~~Which other tokens should we support on Polygon? Just start with the same stables already on mainnet?~~ Add support for the same stablecoins, plus MATIC, and WETH instead of ETH _— **blocked by API updates**_
     - [x] ~~Should we also support WETH on Polygon? Should we display it as WETH or ETH if we do— is there a convention here for other Polygon Dapps?~~ _— will be covered by the above item_
     - [x] ~How does token bridging on Mumbai work? Does Mumbai have the same addresses for bridged assets, or totally different ones? If the latter, we should probably just pick 1 or 2 to supporting, the way Rinkeby supports "Dai" and ETH.~ _—No need to support Mumbai because Polygon is cheap enough_

### Setup Screen

- [ ] On Polygon the presence of the "Setup" button should be driven by whether the connected address is configured *on Polygon*, i.e. whether they've set stealth keys in the current network's registry

### LINKED ISSUE #247: umbra-js: tests currently fail because Hardhat doesn't support personal_sign
In https://github.com/ScopeLift/umbra-protocol/pull/240, we removed `eth_sign` and now only use `personal_sign` for generating the signatures used to derive private keys. However, Hardhat does not yet support `personal_sign` so these tests currently fail with `MethodNotFoundError: Method personal_sign not found`

TBD how to handle this depending on the timeline for Hardhat to support `personal_sign`: https://github.com/nomiclabs/hardhat/issues/1981

mds1:
Closes #212, closes #247

Based on the network you're connected to:
- Etherscan URLs are properly generated
- Native token name and symbol update appropriately
- Toll and payment summary is shown when toll is non-zero (currently only on Polygon)

Features:
- Send page now always defaults to the native token (mirrors Uniswap's behavior)
- Send page now has smarter balance checks: if sending native token, ensure amount + toll <= wallet balance, if sending other tokens, check each separately

Style changes:
- Adds logos to the network selector dropdown
- Update ETH logo to match Uniswap's for consistency across apps (before we used a black diamond one instead of this purple circle one)

Other
- Commit 37000c4 bumps hardhat version and adds a fix for #247

TODO:
- [x] Sometimes the payment link amounts are parsed but then the amount becomes undefined, and there's an error in the console. @radchukd I know you were recently looking at payment link code recently so let me know if you see what the issue is here
- [x] FAQ item on the toll/fee (this should be linked to in the Umbra Fee tooltip). @apbendi do you mind writing this one up?
- [x] Proper ENS resolution, which is a bit tedious since it will require some API changes to umbra-js, so that will come later (possibly in this PR depending on how fast it gets merged). Currently our umbra-js helper methods take a `provider` input and use that for everything (ENS lookup, CNS lookup, reading the chain, etc). But now they need to take two providers: one for mainnet (always mainnet, unless we're on Rinkeby, to facilitate dev/testing) and one for everything else

All of these TODOs can arguably be done in a separate PR, which is why this is not marked as draft


radchukd:
Hi, I have found what's causing the values to reset.
Basically, `tokenOptions` aren't empty, when the relayer is not loaded.

To fix:

Add `if (!relayer.value) return [];` before this line:
https://github.com/ScopeLift/umbra-protocol/blob/4290cc10bf7564baffebcf8b692c552881586c2e/frontend/src/store/wallet.ts#L356

And `tokenOptions.value.length &&` to this line:
https://github.com/ScopeLift/umbra-protocol/blob/a535725c90241d5e77ab3116d6f07c8ec6ce54cb/frontend/src/pages/AccountSend.vue#L228

mds1:
> And `tokenOptions.value.length &&` to this line:

Ah yep, this check seems sufficient to resolve it. Even if we don't have a valid relayer, we still want to support the native token so I don't think we want (or need?) that first change of `if (!relayer.value) return [];`

mds1:
Converting this to draft until we get a Polygon subgraph deployed, since directly querying events is already no longer viable

image

apbendi (commented):
Hey @mds1, really looking and working great overall. A few things that came up in my first wave of testing:

1. Minor, but I wonder if we can think of a different icon to use for Rinkeby. I found it a bit weird/confusing to see two Ethereum icons. Have to noodle on options 🤔 

Screen Shot 2021-11-17 at 1 31 14 PM

2. There seem to be floating point precision issues with calculating totals after fee. This didn't happen with every value, so try different amounts to reproduce.

Screen Shot 2021-11-17 at 1 35 32 PM

Screen Shot 2021-11-17 at 1 36 07 PM

3. MATIC withdrawals seem to be using mainnet gas estimates.

Screen Shot 2021-11-17 at 1 41 50 PM

4. Reverse resolution in the Receive table's "Sender" column still seems to be using Rinkeby when connected to Rinkeby, as in the screenshot below:

Screen Shot 2021-11-17 at 1 46 42 PM

mds1:
> Minor, but I wonder if we can think of a different icon to use for Rinkeby. I found it a bit weird/confusing to see two Ethereum icons. Have to noodle on options 🤔

Fixed in 90347e4 by stealing Uniswap's idea of only showing testnet options if you're already connected to them

> There seem to be floating point precision issues with calculating totals after fee. This didn't happen with every value, so try different amounts to reproduce.

Fixed in dc34d0d

------------------------

other fixes to come....

mds1:
Pushed commit 838beed which adds subgraph and starts the work for #253 

> MATIC withdrawals seem to be using mainnet gas estimates.

Fix pushed in f760d97

mds1:
> Reverse resolution in the Receive table's "Sender" column still seems to be using Rinkeby when connected to Rinkeby, as in the screenshot below:

Fixed in e6d358e

------------------------

Remaining items:
- Warn if user is not on L1, and the address the ENS/CNS destination resolves to is a contract
- Add FAQ item on the toll/fee, and link to it from the Umbra Fee tooltip on the Send page

mds1:
Both of the above remaining items are done in the last commit, so I'd say this PR is ready for review and some thorough QA testing!

contracts/package.json
```diff
@@ -33,7 +33,7 @@
     "ethereum-waffle": "^3.2.0",
     "ethers": "^5.0.17",
     "fs-extra": "^9.0.1",
-    "hardhat": "^2.6.0",
+    "hardhat": "^2.6.8",
```
apbendi, line 36: (arbitrary code location to allow for threading)

If the user does not already have Polygon configured in their wallet, attempting to switch to it seems to fail silently. I think most apps will ask to add the network for the user if it doesn't exist. Can we easily do this as well?

umbra-polygon-no-network-switch

apbendi, line 36: This warning occurs when using Rinkeby, and the address in question is only registered as `umbra-demo.eth` on Rinkeby's ENS deployment. This implies to me this part of the code is still resolving ENS based on the Rinkeby deployment.
 
Screen Shot 2021-12-08 at 4 06 44 PM

apbendi, line 36: The warning message is missing a space before the address

Screen Shot 2021-12-08 at 4 14 54 PM

mds1: We had this feature, and I accidentally broke it when adding the network selector. Details + fix in 1203fca

mds1: 1. Oddly I wasn't able to reproduce this, and it told me `umbra-demo.eth` was an invalid address when on Rinkeby. However I think I see the cause, so I pushed ef043e4 which should fix it. Let me know if that works

2. This made realize that for CNS, we still use their rinkeby registry if we're on Rinkeby. Do we want to keep that, or switch it to mainnet for consistency? My first reaction is to stick with mainnet for CNS too, for consistency

mds1: Fixed in 4c3370c

apbendi, line 36: If I change network, I see an error iff:

1. I change using Metamask, not our selector
2. I'm on the Send page

umbra-network-change-send

apbendi: @mds1

1. Your fix seems to have worked!
2. I agree, always resolving CNS on mainnet is probably best for consistency. Is it a quick change you can tack on to this PR, or should we create a separate issue?

apbendi, line 36: Weird issue, but block explorer links seem to be malformed, with an `https//` (missing colon).

```
https://https//polygonscan.com/tx/0x0ed14cd2610a764d9728492b55981af68e0792ef695e7f4305b6daf37cd99398
```

This is happening with blocknative links and also on the receive screen to historical txs.

mds1: Done in 1233ddc

mds1: broke earlier in this PR in 1203fca, just fixed in 8efe97a

mds1: fixed in 480da6e

PR #261, opened 2021-12-17
feat: stealth key validation

### LINKED ISSUE #259: Check Stealth Keys Match Registry
On the receive page, when a user signs the appropriate message to generate their stealth keys, we should also pull the stealth keys from the registry for their account and compare them.

- [ ] If the user's stealth keys match, we can show a small indicator or icon of some kind on the page, perhaps like a small message above the table that says "✓ Stealth Keys validated". Also add a question mark icon with a hover message that says: "This means your account is properly configured to receive payments via Umbra"
- [ ] If the user has not set their stealth keys yet, we should show a message above the table indicating this and direct them to the setup page "You won't be able to receive funds until you've configured Umbra. Please navigate to the [setup page](https://app.umbra.cash/setup) and do so". Despite this warning, we should still scan for txs and show the table, just in case.
- [ ] If the stealth keys generated by the user by signing the message do not match the keys in the registry, we should show a red alert warning with the following copy:

"It appears the stealth keys published for this account do not match those generated by it. Please navigate to the [setup page](https://app.umbra.cash/setup) and complete the setup flow to ensure the keys published match this account. Failure to do so may result in payments sent to your account being difficult, or even impossible, to withdraw. [Read more]()."

(Read More is a link to the FAQ).

- [ ] Note that this will require a change to our setup screen which allows users to repeat setup if their keys don't match
- [ ] None of this validation on the receive screen should take place if the user is in advanced mode and using a custom public key

mds1:
closes #259

Flows to QA test:
- normal user flow remains the same
- new user flow remains the same
- user with wrong stealth keys visits receive page > visits setup page > sets keys > back to receive page—I copied keys set from one account and used them to set another (i.e. used valid but wrong pubkey points), but I did not test points not on the curve (i.e. did not test random points) because that seems like an unlikely scenario for a user to end up in
- user with no stealth keys visits receive page > same as above from there
- above flows with advanced mode


apbendi (approved):
Looking and working really well. Very nice 👍

PR #269, opened 2022-01-11
Prevent UI Blocking During Scanning and Show Progress Bar

apbendi:
* Refactor umbra-js to separate fetching from scanning announcements
* Chunk the scanning work in the frontend to prevent the UI from locking
* Show a progress bar while scanning for better UX

One weird issue: the UI flashes back to showing the loading indicator briefly after scanning is completed, but it's not clear to me why that is.

### DIFF
--- a/frontend/src/boot/components.ts
+++ b/frontend/src/boot/components.ts
@@ -4,6 +4,7 @@ import BaseButton from 'components/BaseButton.vue';
 import BaseInput from 'components/BaseInput.vue';
 import BaseSelect from 'components/BaseSelect.vue';
 import LoadingSpinner from 'components/LoadingSpinner.vue';
+import ProgressIndicator from 'components/ProgressIndicator.vue';
 
 // more info on params: https://quasar.dev/quasar-cli/cli-documentation/boot-files#Anatomy-of-a-boot-file
 export default boot(({ Vue }) => {
@@ -11,4 +12,5 @@ export default boot(({ Vue }) => {
   Vue.component('base-input', BaseInput);
   Vue.component('base-select', BaseSelect);
   Vue.component('loading-spinner', LoadingSpinner);
+  Vue.component('progress-indicator', ProgressIndicator);
 });
--- a/frontend/src/worker/worker.ts
+++ b/frontend/src/worker/worker.ts
@@ -26,7 +26,7 @@ export const filterUserAnnouncements = (
 
     if (index < announcements.length) {
       progress((100 * index) / announcements.length);
-      setTimeout(doChunk, 1);
+      setTimeout(doChunk, 0);
     } else {
       completion(userAnnouncements);
     }
--- a/frontend/src/worker/worker.ts
+++ b/frontend/src/worker/worker.ts
@@ -1,0 +1,36 @@
+import { AnnouncementDetail, UserAnnouncement, Umbra } from '@umbra/umbra-js';
+import { getAddress } from 'src/utils/ethers';
+
+export const filterUserAnnouncements = (
+  spendingPublicKey: string,
+  viewingPrivateKey: string,
+  announcements: AnnouncementDetail[],
+  progress: (percentage: number) => void,
+  completion: (userAnnouncements: UserAnnouncement[]) => void
+) => {
+  const userAnnouncements: UserAnnouncement[] = [];
+  let index = 0;
+  const chunk = 10;
+
+  const doChunk = () => {
+    for (let count = chunk; count > 0 && index < announcements.length; count--, index++) {
+      const ann = announcements[index];
+      const { amount, from, receiver, timestamp, token: tokenAddr, txHash } = ann;
+      const { isForUser, randomNumber } = Umbra.isAnnouncementForUser(spendingPublicKey, viewingPrivateKey, ann);
+      const token = getAddress(tokenAddr); // ensure checksummed address
+      const isWithdrawn = false; // we always assume not withdrawn and leave it to the caller to check
+      if (isForUser) {
+        userAnnouncements.push({ randomNumber, receiver, amount, token, from, txHash, timestamp, isWithdrawn });
+      }
+    }
+
+    if (index < announcements.length) {
+      progress((100 * index) / announcements.length);
+      setTimeout(doChunk, 1);
+    } else {
+      completion(userAnnouncements);
+    }
+  };
+
+  doChunk();
+};
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -313,8 +313,8 @@ export class Umbra {
     // that window is defined first to avoid trying to use fetch in node environments
     if (typeof window !== 'undefined' && this.chainConfig.subgraphUrl) {
       try {
-        // Map the subgraph amount field from string to BigNumber
         const subgraphAnnouncements = await this.fetchAllAnnouncementsFromSubgraph(startBlock, endBlock);
+        // Map the subgraph amount field from string to BigNumber
         return subgraphAnnouncements.map((x) => ({ ...x, amount: BigNumber.from(x.amount) }));
       } catch (err) {
         return this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
@@ -380,8 +380,8 @@ export class Umbra {
     startBlock: string | number,
     endBlock: string | number
   ): Promise<AnnouncementDetail[]> {
-    // Graph query failed, try requesting logs directly IF we are not on polygon. If we are on
-    // Polygon, there isn't much we can do, so for now just show a warning and return an empty array
+     // Fetching announcements from logs is not possible on Polygon because the network produces too
+     // many blocks too quickly. If this is attempted, we throw an error.
     if (this.chainConfig.chainId === 137) {
       throw new Error('Cannot fetch Announcements from logs on Polygon, please try again later');
     }
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -26,7 +26,7 @@ import { RandomNumber } from './RandomNumber';
 import { blockedStealthAddresses, lookupRecipient } from '../utils/utils';
 import { Umbra as UmbraContract, Erc20 as ERC20 } from '@umbra/contracts/typechain';
 import { ERC20_ABI } from '../utils/constants';
-import type { Announcement, ChainConfig, EthersProvider, ScanOverrides, SendOverrides, SubgraphAnnouncement, UserAnnouncement } from '../types'; // prettier-ignore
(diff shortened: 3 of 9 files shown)


frontend/src/worker/worker.ts
```diff
@@ -0,0 +1,36 @@
...
+      const token = getAddress(tokenAddr); // ensure checksummed address
+      const isWithdrawn = false; // we always assume not withdrawn and leave it to the caller to check
+      if (isForUser) {
+        userAnnouncements.push({ randomNumber, receiver, amount, token, from, txHash, timestamp, isWithdrawn });
+      }
+    }
+
+    if (index < announcements.length) {
+      progress((100 * index) / announcements.length);
+      setTimeout(doChunk, 1);
```
mds1, line 29: you could set this to zero though in practice it won't matter (it's the usage of `setTimeout` that helps us, not the delay, more [here](https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful))

umbra-js/src/classes/Umbra.ts
```diff
@@ -26,7 +26,7 @@ import { RandomNumber } from './RandomNumber';
 import { blockedStealthAddresses, lookupRecipient } from '../utils/utils';
 import { Umbra as UmbraContract, Erc20 as ERC20 } from '@umbra/contracts/typechain';
 import { ERC20_ABI } from '../utils/constants';
-import type { Announcement, ChainConfig, EthersProvider, ScanOverrides, SendOverrides, SubgraphAnnouncement, UserAnnouncement } from '../types'; // prettier-ignore
+import type { Announcement, ChainConfig, EthersProvider, ScanOverrides, SendOverrides, SubgraphAnnouncement, UserAnnouncement, AnnouncementDetail } from '../types'; // prettier-ignore
```
mds1, line 29: (arbitrary comment location)

scanning seems to make no progress when the Umbra tab doesn't have focus, so I can't switch tabs and let it scan in the background—any idea what's going on here?

umbra-js/src/classes/Umbra.ts
```diff
@@ -313,107 +313,161 @@ export class Umbra {
...
-          const isWithdrawn = false; // we always assume not withdrawn and leave it to the caller to check
-          if (isForUser) userAnns.push({ randomNumber, receiver, amount, token, from, txHash, timestamp, isWithdrawn });
-          return userAnns;
-        }, [] as UserAnnouncement[]);
-
-        // Filtering and parsing done, return announcements
-        return { userAnnouncements };
+        // Map the subgraph amount field from string to BigNumber
+        const subgraphAnnouncements = await this.fetchAllAnnouncementsFromSubgraph(startBlock, endBlock);
+        return subgraphAnnouncements.map((x) => ({ ...x, amount: BigNumber.from(x.amount) }));
```
mds1, line 318: comment is off by a line, maybe just inline it
```suggestion
        const subgraphAnnouncements = await this.fetchAllAnnouncementsFromSubgraph(startBlock, endBlock);
        return subgraphAnnouncements.map((x) => ({ ...x, amount: BigNumber.from(x.amount) })); // map subgraph amount field from string to BigNumber
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -313,107 +313,161 @@ export class Umbra {
...
+    endBlock: string | number
+  ): Promise<SubgraphAnnouncement[]> {
+    if (!this.chainConfig.subgraphUrl) {
+      throw new Error('Subgraph URL must be defined to fetch via subgraph');
+    }
+
+    // TODO: We're ignoring these overrides for The Graph. Is this intentional? Should we remove the parameters,
+    // or update so it uses them?
+    startBlock;
+    endBlock;
```
mds1, line 347: ah these were being ignored originally before the refactor, right? I think that's just an oversight on my part when implementing the subgraph query

we probably should respect those values to save the user scanning time, but it's advanced-mode only so likely not a common path, so I'm ok punting on it and tracking it in an issue (unless you want to change it in this PR, then of course feel free)

umbra-js/src/classes/Umbra.ts
```diff
@@ -313,107 +313,161 @@ export class Umbra {
...
-  async userAnnouncementsFromLogs(
-    spendingPublicKey: string,
-    viewingPrivateKey: string,
+  async fetchAllAnnouncementFromLogs(
     startBlock: string | number,
     endBlock: string | number
-  ): Promise<UserAnnouncement[]> {
+  ): Promise<AnnouncementDetail[]> {
+    // Graph query failed, try requesting logs directly IF we are not on polygon. If we are on
+    // Polygon, there isn't much we can do, so for now just show a warning and return an empty array
```
mds1, line 384: two outdated comments here:
1. since this is a separate method, `Graph query failed` is outdated
2. comment says show a warning and return an empty array, but actually we throw an error

umbra-js/src/classes/Umbra.ts
```diff
@@ -313,107 +313,161 @@ export class Umbra {
...
+   * @notice Scans all Umbra event logs for funds sent to the specified address
+   * @dev This is a convenience method that first runs `fetchAllAnnouncements`, then filters them through
+   * the static helper `isAnnouncementForUser`. The latter is CPU intensive and this method will block while
+   * all announcements are processed. To avoid performance issues, you may need to run fetching and filtering
+   * steps separately, and use chunking, web workers, or other threading strategies.
+   * @param spendingPublicKey Receiver's spending private key
+   * @param viewingPrivateKey Receiver's viewing public key
+   * @param overrides Override the start and end block used for scanning
+   */
+  async scan(spendingPublicKey: string, viewingPrivateKey: string, overrides: ScanOverrides = {}) {
```
mds1, line 427: just to confirm—this is the same behavior as the original `scan` method, just refactored to use the new methods?

umbra-js/src/types.ts
```diff
@@ -55,6 +55,19 @@ export interface Announcement {
   token: string;
 }
 
+// Type definition for Announcement event with extra data from the chain
+export interface AnnouncementDetail {
```
mds1, line 59: just curious what the context is around this new type—was it implicitly used before and now just explicitly defined?

frontend/src/boot/components.ts
```diff
@@ -4,11 +4,13 @@ import BaseButton from 'components/BaseButton.vue';
 import BaseInput from 'components/BaseInput.vue';
 import BaseSelect from 'components/BaseSelect.vue';
 import LoadingSpinner from 'components/LoadingSpinner.vue';
+import ProgressIndicator from 'components/ProgressIndicator.vue';
```
mds1, line 7: (arbitrary location)

> the UI flashes back to showing the loading indicator briefly after scanning is completed, but it's not clear to me why that is.

what's happening here is:
- we set `scanStatus` to complete
- this triggers the `AccountReceiveTable` component to be mounted [here](https://github.com/ScopeLift/umbra-protocol/blob/0a06f0ca72d473c66747a4a7116d76ab23c254f6/frontend/src/pages/AccountReceive.vue#L79), and that component is passed an array of announcements returned from scanning
- the `AccountReceiveTable` component then shows _it's own loading spinner_, because it does async post-processing of the announcements [here](https://github.com/ScopeLift/umbra-protocol/blob/0a06f0ca72d473c66747a4a7116d76ab23c254f6/frontend/src/components/AccountReceiveTable.vue#L409-L429)

mds1 (commented):
Awesome job on this!! The percentage loading spinner is 🔥🔥🔥, such a better UX

frontend/src/boot/components.ts
mds1: maybe in `AccountReceiveTable` we change the message from "Scanning for funds..." to "Almost done...." or something a bit different/distinguished from the previous "scanning for funds" message

apbendi: Another option would be to set it to the progress indicator with 100% so it doesn't appear to flash at all. Feels a little hacky but might better UX. What do you think? Gonna try it out and see how it feels.

frontend/src/worker/worker.ts
```diff
@@ -0,0 +1,36 @@
...
+export const filterUserAnnouncements = (
+  spendingPublicKey: string,
+  viewingPrivateKey: string,
+  announcements: AnnouncementDetail[],
+  progress: (percentage: number) => void,
+  completion: (userAnnouncements: UserAnnouncement[]) => void
+) => {
+  const userAnnouncements: UserAnnouncement[] = [];
+  let index = 0;
+  const chunk = 10;
```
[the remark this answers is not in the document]
apbendi: I chose 10 because that was the point at which it started feeling smoothest. If we wanted to really tune this, we would test it across different devices, but I don't think it's a big enough deal.

I don't think we need to make this an ENV var because it's not something that will need to be changed based on changing external conditions. I.E. the experience is the same regardless of the number of events we need to scan.

umbra-js/src/classes/Umbra.ts
apbendi: Good catch! I did not realize that. Let me do some investigation. If it's not a quick fix I'll make an issue.

umbra-js/src/classes/Umbra.ts
apbendi: That's right! Behaves the same as the original method— I didn't even change the tests. It just leverages the new methods rather than implementing the behavior internally.

frontend/src/boot/components.ts
mds1: Hmm, there's a good chance in the future that we no longer can reliably show a progress indicator (e.g. if scanning is done in a web worker or wasm), so we'd be removing this whole thing and make sure to remove it there too. I think my preference would be to change the copy to:

- Step 1: Fetching annoucements...
- Step 2: Checking for funds...
- Step 3: Processing result...

And that way it carries over nicely to a future version where we can't show a progress bar in step 2

apbendi: I think a webworker and/or wasm version will still be able to report status updates so we can show completion percentage. That said, after trying it out, I actually don't like my own suggestion. It makes it seem like it's "stuck" at 100%. Went with your suggestion and updated to say "Processing results..."

frontend/src/worker/worker.ts
apbendi: updated.

umbra-js/src/classes/Umbra.ts
apbendi: fixed

umbra-js/src/classes/Umbra.ts
apbendi: fixed

umbra-js/src/classes/Umbra.ts
apbendi: Yep these were ignored prior to the refactor, it just became more obvious when they were split into stand alone methods. I created an issue to track here: https://github.com/ScopeLift/umbra-protocol/issues/275

umbra-js/src/classes/Umbra.ts
mds1: huh, that's annoying, but sounds good thanks!

mds1 (approved):

Issue #265, opened 2022-01-05
Amount Error Status Fails To Update When Currency Changed


apbendi:
On the send screen, when the currency changes, the error status from the previous currency is not updated. Note the status of the "Send" button itself *does* seem to change as it should

![asset-min-bug-umbra](https://user-images.githubusercontent.com/1479802/148292038-0dcfe2fc-2542-46ec-85ba-c510f40df421.gif)
.

PR #272, opened 2022-01-14
bug: ensure form validation happens when values are present

thelostone-mc:
#### Description 

Updates condition when form validation happens to ensure validation run if both token value and amount are present/updated.

The issue here was the `isValidTokenAmount` is mapped to the token amount field and the error message wasn't getting triggered when the token name was being updated 

#### Demo

Untitled

#### Fixes
fixes https://github.com/ScopeLift/umbra-protocol/issues/265

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -245,8 +245,14 @@ function useSendForm() {
         humanAmount.value = undefined;
       }
 
-      // Revalidates form on network change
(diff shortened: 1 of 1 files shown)


apbendi:
Hey @thelostone-mc, this is awesome! Will review this sometime soon. Thanks for the surprise contribution 🎉

mds1 (approved):
LGTM, thanks a lot! Will let our QA expert @apbendi review also before merging 😛

apbendi (approved):
Works great!  Thanks @thelostone-mc

PR #273, opened 2022-01-14
doc: updates FAQ grammar

thelostone-mc:
#### Description

Updates FAQ grammar. 
 
 Most users will use the Umbra stealth key registry to send funds, so the recipient most likely keys 
 **IS UPDATED TO** 
 Most users will use the Umbra stealth key registry to send funds, so the recipient most likely has keys


mds1 (approved):
Thanks!

PR #279, opened 2022-01-20
Frontend tweaks to support WETH on Polygon

apbendi:
Makes updates to the frontend in preparation for supporting non-stablecoin
tokens on Polygon, notably WETH.

* Defines minimum amount for Polygon WETH as 0.001 on the send page
* Rounds WETH amounts, which can be well less than 0 on Polygon, to
  display 6 decimals when showing fees and received amounts

Note: before the backend update for WETH support is deployed, there should be no noticeable changes in the UI from this PR. This PR can and should be reviewed, merged, and deployed ahead of the rollout of WETH support on the backend to ensure it works seamlessly on the frontend immediately.

### DIFF
--- a/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
+++ b/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
@@ -108,7 +108,7 @@ import { UserAnnouncement } from '@umbra/umbra-js';
 import { FeeEstimate } from 'components/models';
 import { formatAddress } from 'src/utils/address';
 import { BigNumber, formatUnits } from 'src/utils/ethers';
-import { getEtherscanUrl, getGasPrice, round } from 'src/utils/utils';
+import { getEtherscanUrl, getGasPrice, round, roundTokenAmount } from 'src/utils/utils';
 import useWalletStore from 'src/store/wallet';
 
 export default defineComponent({
@@ -160,7 +160,12 @@ export default defineComponent({
     const etherscanUrl = computed(() => getEtherscanUrl(props.txHash, props.chainId)); // withdrawal tx hash URL
     const ethDisplayDecimals = 6;
     const numDecimals = isNativeToken ? ethDisplayDecimals : 2; // maximum number of decimals to show for numbers the UI (ETH fee will be small, hence the larger value)
-    const formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
+    let formattedAmount: string;
+    if (isNativeToken) {
+      formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
+    } else {
+      formattedAmount = roundTokenAmount(amount, props.activeFee.token);
+    }
     function isValidFeeAmount(val: string) {
       if (!val || !(Number(val) > 0)) return 'Please enter an amount';
       if (BigNumber.from(amount).lte(customFeeInWei.value)) {
@@ -205,8 +210,21 @@ export default defineComponent({
     // Define computed properties dependent on the fee (must be computed to react to ETH gas price updates by user).
     // Variables prefixed with `formatted*` are inteded for display in the U)
     const amountReceived = computed(() => amount.sub(useCustomFee.value ? customFeeInWei.value : fee.value)); // amount user will receive
-    const formattedFee = computed(() => round(formatUnits(fee.value, decimals), numDecimals)); // relayer fee, rounded
-    const formattedAmountReceived = computed(() => round(formatUnits(amountReceived.value, decimals), numDecimals)); // amount user will receive, rounded
+    const formattedFee = computed(() => {
+      if (isNativeToken) {
+        round(formatUnits(fee.value, decimals), ethDisplayDecimals);
+      } else {
+        return roundTokenAmount(props.activeFee.fee, props.activeFee.token);
+      }
+      return round(formatUnits(fee.value, decimals), numDecimals);
+    }); // relayer fee, rounded
+    const formattedAmountReceived = computed(() => {
+      if (isNativeToken) {
+        return round(formatUnits(amountReceived.value, decimals), numDecimals);
+      } else {
+        return roundTokenAmount(amountReceived.value, props.activeFee.token);
+      }
+    }); // amount user will receive, rounded
     // prevent withdraw attempts if fee is larger than amount
     const canWithdraw = computed(() => {
       if (!loaded.value) return true; // assume true until finished loading, to prevent resize issue shown here: https://github.com/ScopeLift/umbra-protocol/pull/206#pullrequestreview-718683599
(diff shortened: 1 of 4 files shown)


frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -108,7 +108,7 @@ import { UserAnnouncement } from '@umbra/umbra-js';
 import { FeeEstimate } from 'components/models';
 import { formatAddress } from 'src/utils/address';
 import { BigNumber, formatUnits } from 'src/utils/ethers';
-import { getEtherscanUrl, getGasPrice, round } from 'src/utils/utils';
+import { getEtherscanUrl, getGasPrice, round, roundTokenAmount } from 'src/utils/utils';
```
mds1, line 111: _(arbitrary comment location to start a thread)_

Just documenting that we said the scope of this PR is not limited to Polygon, but in practice Polygon is where we'll be added support for more tokens first

frontend/src/utils/utils.ts
```diff
@@ -28,6 +28,22 @@ export const round = (value: number | string, decimals = 2) => {
   });
 };
 
+/**
+ * @notice Rounds to appropriate human readable decimals for the token
+ * @dev This is very fragile and should be replace with something robust, i.e. something that walks
+ * value and rounds considering sig-figs or similar
+ */
+export const roundTokenAmount = (amount: BigNumberish, token: { decimals: number; address: string }): string => {
```
mds1, line 36: Per our offline discussion, we might change this to something like this (untested) snippet:

```typescript
// Show two sig figs if number is <= 1, and show two decimals otherwise
export function formatAmount(amount: BigNumberish, decimals: number) {
  const formattedAmount = parseFloat(formatUnits(amount, decimals))
  // I think toLocaleString with maximumFractionDigits=2 should also work instead of toPrecision
  if (formattedAmount <= 1) return formattedAmount.toPrecision(2) 
  return formattedAmount.toLocaleString(undefined, {
      minimumFractionDigits: 0,
      maximumFractionDigits: 10,
    })
}
```

Would we still need the existing `round` and `formatAmount` methods used in various places, or could we use this everywhere?

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -160,7 +160,12 @@ export default defineComponent({
     const etherscanUrl = computed(() => getEtherscanUrl(props.txHash, props.chainId)); // withdrawal tx hash URL
     const ethDisplayDecimals = 6;
     const numDecimals = isNativeToken ? ethDisplayDecimals : 2; // maximum number of decimals to show for numbers the UI (ETH fee will be small, hence the larger value)
-    const formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
+    let formattedAmount: string;
+    if (isNativeToken) {
+      formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
+    } else {
+      formattedAmount = roundTokenAmount(amount, props.activeFee.token);
+    }
```
mds1, line 168: Then with the updated formatting methods we could then remove these if/else blocks to just `const formattedAmount = formatAmount(amount, tokenDecimals)`?

frontend/src/pages/AccountSend.vue
```diff
@@ -296,8 +296,13 @@ function useSendForm() {
...
     if (chainId === 137) {
-      if (isNativeToken(tokenAddress)) return 1.0;
-      else return 3; // stablecoin token minimum
+      if (isNativeToken(tokenAddress)) {
+        return 1.0;
+      } else if (tokenAddress === '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619') {
+        return 0.001; // weth token minimum
+      } else {
+        return 3; // stablecoin token minimum
+      }
```
mds1, line 305: Hmm, we should probably think of a better way to minimum send amounts, since my naive approach here won't scale very well.

Having the relayer's `/tokens?chainId=x` endpoint utilize the `extensions` field of the `TokenInfo` type to return an `extensions.minimumSendAmount` field seems like a good solution

Then for native tokens we'd want specify the minAmount to send in the frontend, perhaps the `nativeCurrency` fields in `models.ts`

apbendi: I agree this probably belongs on the backend. Including it in the tokens endpoint could work. We could also add a field to that endpoint with the the native currency. I think it'd be better to have it all on the backend, rather than split across frontend/backend.

frontend/src/utils/utils.ts
apbendi: Agree with all of this. Mind if we pull into a separate issue and address it all at once?

mds1: Do you mean leave this PR as-is and implement that method in the separate issue, or implement it now but pull out just the "use the new method everywhere else" part into a separate issue?

frontend/src/pages/AccountSend.vue
mds1: The reason I disagree with having the native token also on the backend is because you don't need the relayer/API to send the native token. So if the relayer is ever down, then sending the native token also goes down, which is not really necessary. 

A compromise there is to have the relayer return native token info, but the frontend has its own to fallback on if the relayer is down.

But for the scope of this issue I think this section is fine and we can move this discussion somewhere else!

frontend/src/utils/utils.ts
apbendi: The former. The feedback loop on testing this is rather slow so if there's no show stoppers on the current implementation, it'd be nice to get it in so we can launch WETH support soon, and clean it all up at once.

frontend/src/pages/AccountSend.vue
apbendi: Good point! I like the default + backend override approach. Will create a new issue for tracking and continuing the discussion.

frontend/src/utils/utils.ts
mds1: Sounds good to me!

mds1 (approved):

mds1:
Good to merge once we create both of the above issues for  relayer update and rounding method update

frontend/src/utils/utils.ts
apbendi: Issue created [here](https://github.com/ScopeLift/umbra-protocol/issues/283)

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
apbendi: Issue created [here](https://github.com/ScopeLift/umbra-protocol/issues/283)

frontend/src/pages/AccountSend.vue
apbendi: https://github.com/ScopeLift/umbra-api/issues/38

PR #284, opened 2022-01-21
fix: properly handle different transaction types

mds1:


apbendi (approved):
Tested native asset + token sends with accounts that have Type 0 & Type 2 original sends. All worked without issue. Also tested normal sending. Everything working well.

Code looks good and tests pass. Thanks for getting this fix in 👍

Issue #262, opened 2021-12-17
fix: let users change gas limit when withdrawing native tokens


mds1:
We currently hardcode a 21000 gas limit, which means you can't withdraw to contracts that have a fallback function that does something, e.g. gnosis safes

thelostone-mc:
Would this be simply adding a new field on the withdraw section to allow users them to override 21000 gas limit ?

mds1:
First I'll explain how you would tackle this issue as described, then I'll describe what I think we should do instead:

------------------------

### Original issue intent 

Yep so in `Umbra.ts` (part of the umbra-js library) I don't think any changes will be needed — @thelostone-mc the part you linked to shows that if a gas limit override is specified it will be used. However, the frontend calls this withdraw method in `AccountReceiveTable.vue` [here](https://github.com/ScopeLift/umbra-protocol/blob/ae0c8901e2826bad502a20429b52c0feadf82608/frontend/src/components/AccountReceiveTable.vue#L503) and you can see no gas limit is passed in.

So to resolve this issue the way it's described:
1. In the image below, we need a gas limit field to the right to the gas price field
2. Update the line linked above to pass in the user specified gas limit

<img src="https://user-images.githubusercontent.com/17163988/149950619-27b03ea6-39c0-4baa-a3c4-1b48a0f6dbc6.png" width=400 />

------------------------

### New Idea

Instead of asking the user to specify a gas limit to support withdrawing to contracts with fallback functions, we should just estimate the gas limit required based on the address specified. 

This should be pretty simple to do with some changes to `AccountReceiveTable.vue`:

1. Once a `destinationAddress` is entered, call `provider.value.estimateGas({ to: destinationAddress, from: stealthAddress, gasPrice: 0, value })` which returns the gas limit needed
2. Default to using the returned gas limit instead of 21000, and use that to compute how much ETH to withdraw

@apbendi let me know what you think—this new approach should be better UX for users because it continues to hide the gas limit while providing the functionality we need

apbendi:
@mds1 it seems reasonable to me, my only question is around reliability. We've definitely seen plenty of cases of estimateGas being flaky over the years, and fallback addresses seem to be one of the things that cause trouble. That said I think it's worth implementing to see how it works. It's certainly better UX then making the user set it themselves.

mds1:
[...]

PR #285, opened 2022-01-28
fix: auto detect gaslimit on withdraw

thelostone-mc:
#### Description 

`withdrawal` function has been updated and when invoked
-  if `overrides.gasLimit` is passed ->  that value is used
-  else it will try to fetch the gasLimit via `estimateGas`
- and if any error occurs -> will default to `2100` 

#### Refs 

fixes : https://github.com/ScopeLift/umbra-protocol/issues/262

#### Testing 

- locally tested with debugger statements
- Additonally set the default gasLimit to 2500 and executed the withdraw -> and the gas limit was set to 21000 (by the estimateGas` function
https://polygonscan.com/tx/0x2c64a01853ad85e4e5df74d2a03bdbf8ffb269100c6cd5a05a8c46e873dac4db

### DIFF
--- a/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
+++ b/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
@@ -106,7 +106,7 @@
 import { computed, defineComponent, onMounted, PropType, ref } from '@vue/composition-api';
 import { UserAnnouncement } from '@umbra/umbra-js';
 import { FeeEstimate } from 'components/models';
-import { formatAddress } from 'src/utils/address';
+import { formatAddress, toAddress } from 'src/utils/address';
 import { BigNumber, formatUnits } from 'src/utils/ethers';
 import { getEtherscanUrl, getGasPrice, round, roundTokenAmount } from 'src/utils/utils';
 import useWalletStore from 'src/store/wallet';
@@ -207,9 +207,8 @@ export default defineComponent({
 
     onMounted(async () => {
       if (isNativeToken) {
-
         gasLimit.value = (await provider.value?.estimateGas({
-          to: props.destinationAddress,
+          to: toAddress(props.destinationAddress, provider.value),
           from: props.activeAnnouncement.receiver,
           value: amount,
           gasPrice: 0,
--- a/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
+++ b/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
@@ -176,6 +176,8 @@ export default defineComponent({
 
     // Get initial fee on component mount. If native token, calculate gas cost of a transfer, otherwise use the provided relayer fee
     const fee = ref<BigNumber | string>('0'); // default to a fee of zero
+    const gasLimit = ref<BigNumber>(BigNumber.from('21000')); // default to a 21k gaslimit
+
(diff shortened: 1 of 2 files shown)


umbra-js/src/classes/Umbra.ts
```diff
@@ -231,8 +231,23 @@ export class Umbra {
       // Withdraw ETH
       // Based on gas price, compute how much ETH to transfer to avoid dust
       const ethBalance = await this.provider.getBalance(stealthWallet.address); // stealthWallet.address is our stealthAddress
+
+      // Estimate gas limit if not provided
```
mds1, line 235: One thing we also want to add is a check, probably during setup of `AccountReceiveTableWithdrawConfirmation.vue`, that runs the same `estimateGas` call and updates the fee amount show to the user. Currently if I withdraw to e.g. [0x6851D6fDFAfD08c0295C392436245E5bc78B0185](https://rinkeby.etherscan.io/address/0x6851d6fdfafd08c0295c392436245e5bc78b0185#code), which is a random gnosis safe I found on rinkeby, then the send method here correctly uses the right gas limit, but the estimated fee shown to the user is incorrect.

My suggestion would be to keep the current logic in this file as a fallback, but our frontend should also call `estimateGas`, show the appropriate fee, and pass in the gas limit override to this `withdraw` method

mds1 (commented):
lgtm once conflict is resolved. I don't have permission to push to your branch (there should be a checkbox in the PR you can enable), but the conflict fix should be this:

```typescript
const gasPrice = network.value?.chainId === 1 ? await tryGetGasPrice() : await provider.value?.getGasPrice(); // use blocknative on mainnet, the node elsewhere
const ethFee = gasLimit.value.mul(gasPrice as BigNumber);
```

after conflict fix + @apbendi QA testing we can merge!

thelostone-mc:
@mds1 doneeee

apbendi (changes requested):
Hey @thelostone-mc, thanks for the contribution! The core fix seems to be working, however, when I test a withdrawal on Polygon using ENS, I get the following error when it attempts to estimate gas:

Screen Shot 2022-02-03 at 5 33 31 PM

This error does not occur with the currently deployed version.

mds1:
@apbendi When exactly are you seeing that? I tried to reproduce but using `msolomon.eth` as the recipient works for me on polygon locally—I was able to get to the MM confirmation then rejected the transaction. Maybe try re-installing packages as a sanity check? (though the version number there looks correct...)

apbendi:
@mds1 I did fresh install to be sure and I'm still seeing the issue. See below:

umbra-polygon-ens-fail

mds1:
Ah whoops I was looking at the send page, not receive page 🤦

Either way, just pushed a fix—let me know if that works!

apbendi (approved):
It works! 👍

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -207,9 +207,8 @@ export default defineComponent({
 
     onMounted(async () => {
       if (isNativeToken) {
-
         gasLimit.value = (await provider.value?.estimateGas({
-          to: props.destinationAddress,
+          to: toAddress(props.destinationAddress, provider.value),
```
thelostone-mc, line 211: Oh nice catch here! Derp can't believe i missed that
Sorry about that @apbendi

thelostone-mc:
Ah thanks for pushing that in @mds1  <3

PR #286, opened 2022-01-29
fix: fallback to node's gas price if API is down

mds1:
Current behavior: If https://api.txprice.com/ is down (like it is right now), mainnet ETH withdrawals are blocked because the call to their API fails and we do not handle that error

New behavior: Catch the error and fallback to using a gas price provided by the node. @apbendi feel free to rename `tryGetGasPrice` to `safeGetGasPrice` or any other method name you prefer

If https://api.txprice.com/ is no longer down at the time of testing, you may need to manually force `getGasPrice()` to throw to confirm the behavior works as expected


apbendi (approved):
Tested on Rinkeby by editing the chainId locally. Works properly and looks good. Thanks for the fast turn around.

PR #293, opened 2022-02-15
Arbitrum and Optimism support

mds1:
Marking this as ready for review, however the frontend has not yet been test much at all. That's next on my list, hoping to get to that tomorrow.

What's left before merging:
- [ ] Lots of frontend testing
- [x] Determine if we can fetch announcements from event logs on Arbitrum and Optimism. If not, disable using events as a fallback when the subgraph is down
    - [x] Optimism event logs do not work: `eth_getLogs and eth_newFilter are limited to a 10,000 blocks range`, so `fetchAllAnnouncementFromLogs` has been updated accordingly
    - [x] Arbitrum event logs do work
- [ ] Question for @apbendi and @davidlaprade: What changes, if any, do I need to make based on server-side changes around how we determine minimum send amounts?


tsconfig.settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+      "skipLibCheck": true,
```
mds1, line 3: @wildmolasses Curious if you have any better solutions here, since I know you're not a fan of `skipLibCheck`. The problem was that `@types/mocha` and `@types/jest` have conflicting definitions, and since we need both running `yarn build` in the `umbra-js` package failed with the errors described here: https://stackoverflow.com/questions/55680391/typescript-error-ts2403-subsequent-variable-declarations-must-have-the-same-typ

Removing one of the two packages doesn't seem like a great option since we use Jest in the frontend and Mocha in the contracts/umbra-js packages

wildmolasses: [`skipLibCheck`](https://www.testim.io/blog/typescript-skiplibcheck/) is a code smell. we want to know about declaration errors whether they're in library code or our code, and resolve them where possible.

furthermore, codebases that consume a package using `skipLibCheck` must also use `skipLibCheck`

If possible, let's avoid `skipLibCheck`; here, we do it by separating the mocha and jest types via a nohoist declaration so that each subpackage has the relevant declaration file.

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
   "plugins": ["prettier"],
```
apbendi, line 3: (Arbitrary comment location)

 First pass through testing on Optimism, Setup + Send worked flawlessly! 🎉 

First issue I hit was related to withdrawal, which failed with the following JSON-RPC error:

Screen Shot 2022-03-17 at 11 16 19 PM

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
...
   "rules": {
```
apbendi, line 4: Another issue on withdrawal on Optimism:

The fee estimate is shown in scientific notation (maybe that's ok?) but when I click the button to edit it, I get an error:

Screen Shot 2022-03-17 at 11 18 01 PM

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
...
     "code-complexity": ["error", 7],
```
apbendi, line 5: On Arbitrum, Setup worked w/o issue, but:

* The toast took 15+ seconds to change whereas the tx was shown as confirmed almost instantly on arbiscan
* After it was confirmed, in addition to seeing the toast saying it succeeded, I saw this error:

Screen Shot 2022-03-17 at 11 22 47 PM

The send did succeed, heres the tx: https://arbiscan.io//tx/0xbd34fa2b4fdf38cf8a27c677a7d11097b965c535ab03c0fbb63a974dcfaeee2e

apbendi: On the withdraw side, the fee estimation looks good:

Screen Shot 2022-03-17 at 11 27 27 PM

However clicking the button to edit it yields the precision related error:

Screen Shot 2022-03-17 at 11 25 46 PM

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
...
-    "compiler-version": ["error", "^0.6.12"],
+    "compiler-version": ["warn", "^0.6.12"],
     "const-name-snakecase": "off",
```
apbendi, line 7: Finally, attempting to perform a withdrawal on Arbitrum fails with the following JSON RPC error:

Screen Shot 2022-03-17 at 11 27 27 PM

apbendi (commented):
Just did a first QA pass through and left comments. Setup, Send, and fund scanning all seemed to work (with one minor sending issue on Arbitrum but it still worked). Withdrawal is not working on either network yet. Really solid first pass given you haven't tested yet!

Just out of curiosity, how did you choose the order in the network dropdown?

apbendi:
Oh, one additional comment: I think if we go-live with just ETH then it's ok to leave the hardcoded min send amounts, and we can tackle using the new backend fields in a separate PR.

contracts/.solhint.json
davidlaprade: Yeah, I was going to say, the rounding on that fee estimate did not look correct. Seems like this is a bug

mds1:
> Just out of curiosity, how did you choose the order in the network dropdown?

In the `supportedChains` array in `models.ts`, we them by Chain ID, and that just carries over to the network selector dropdown. Open to changing it to be alphabetical or otherwise if you prefer

> I think if we go-live with just ETH then it's ok to leave the hardcoded min send amounts, and we can tackle using the new backend fields in a separate PR.

Cool, so I'll keep the `getMinSendAmount` as-is for now then until I hear from you otherwise depending on what we choose to go live with

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
   "plugins": ["prettier"],
   "rules": {
     "code-complexity": ["error", 7],
-    "compiler-version": ["error", "^0.6.12"],
+    "compiler-version": ["warn", "^0.6.12"],
```
mds1, line 6: _(arbitrary comment location to start a thread)_

For Arbitrum and Optimism, let's make sure to test the advanced mode "send to standard pubkey" functionality using both EOAs that have sent transactions in the past and new EOAs

contracts/.solhint.json
mds1: From some quick google searches, it seems we'll have to write our own handling of very small numbers to avoid them being handled with scientific notation, which seems like a pain. For now I just pushed 8f2b392 which adds a failing test but perhaps we pull this into a separate issue?

mds1: Part of the issue here is that the 2.1e-8 fee that is being shown is just the L2 execution fee, but the L1 brings the fee total to roughly `0.00018`, which probably would not get shown as scientific notation. So we should update the confirm withdrawal modal to use the new `getEthSweepGasInfo` method from `Umbra.ts`, which returns a `txCost` value that should be shown

contracts/.solhint.json
mds1: I saw this the first time I tried, added a bunch of `console.log` statements for fancy debugging, then of course the second time it worked. I then changing my RPC to my own Alchemy one instead of the default https://mainnet.optimism.io, tried another send > withdraw, and that also worked. So I wonder if this was some temporary RPC flakiness? I sent 0.01 ETH which should be plenty of margin for Optimism gas prices. Can you try again and let me know what happens?

------------------------

Also, because we can't perfectly sweep ETH and leave a few cents of dust, after refreshing the page the table still thinks funds are not withdrawn because balance is non-zero. So we'll have to change how the "withdrawn" status is determined for Optimism/Arbitrum. Am I correct that token withdrawals will not leave dust, so this dust issue only applies to ETH withdrawals?

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
   "plugins": ["prettier"],
   "rules": {
     "code-complexity": ["error", 7],
-    "compiler-version": ["error", "^0.6.12"],
```
mds1, line 6: It seems the received funds are not sorted properly. You can see the most recently received funds are at the bottom, but they should be at the top

image

contracts/.solhint.json
mds1: That same `toPrecision() argument must be between 1 and 100` error shows up when clicking the edit button on Optimism also

contracts/.solhint.json
mds1: Re (1), I have now seen it again. I just pushed 70d8436 which I _think_ may fix it, but need to test a few more times. If that doesn't work, we probably just need to add some padding to the txCost estimate. Relevant Optimism docs: [1](https://community.optimism.io/docs/developers/build/transaction-fees/), [2](https://community.optimism.io/docs/developers/build/transaction-fees/#insufficient-funds), [3](https://community.optimism.io/docs/guides/wallet-dev/#sending-max-eth)

contracts/.solhint.json
mds1: > The toast took 15+ seconds to change whereas the tx was shown as confirmed almost instantly on arbiscan

This seems to happen sometimes on Optimism too. I'd guess the issue is either that the default provider config [we use](https://github.com/ScopeLift/umbra-protocol/blob/a71af1425c497c046a960ca0dea4e6fc31a51d7e/frontend/src/utils/alerts.ts#L71) to monitor tx status may not be polling frequently enough, or on L2s it waits for additional confirmations, but need to look more into it

contracts/.solhint.json
mds1: I think you forgot to show the JSON RPC error here

apbendi: Ahh, you're right. Sorry about that. Whatever the error was, it didn't occur on my latest test. I was able to do a withdrawal. However, it seems our "sweeping" algorithm needs some work. I had sent 0.01 ETH, and it [left behind](https://arbiscan.io/address/0x2973a807ea703c750fa38e9a43409fbe23ceeb8e) more than half in the stealth address.

contracts/.solhint.json
apbendi: Got it. Yeah, I'm not even sure we'd want to show a number like 2.1e-8 as `0.00000002.1`, it would end up looking pretty goofy anyway. But if the real fee is actually higher, we should show be showing that anyway, which sounds like it fixes the scientific notation problem.

contracts/.solhint.json
apbendi: Still seeing this error attempting a withdrawal on Optimism:

Screen Shot 2022-03-19 at 9 05 05 AM

mds1: This should now be resolved with 4788b35

contracts/.solhint.json
mds1: The fee should now be properly displayed as of 4788b35, but I still need to fix the edit button

contracts/.solhint.json
mds1: > On Arbitrum, Setup worked w/o issue, but:
> - After it was confirmed, in addition to seeing the toast saying it succeeded, I saw this error:

I was not able to reproduce this. @apbendi if you know how let me know. Worth noting that I have metamask configured to use Alchemy RPC's over default ones because it's less flaky and more reliable

mds1:
Ok everything above should be fixed with any exceptions listed in the below todos. All of these can optionally be separate issues/PRs if we prefer since they aren't critical to getting this merged/deployed

- [ ] Withdraw button should show loading spinner/disabled: There's a delay from async calls between clicking withdraw and seeing the modal, so we should disable the button in that time
- [ ] Fix table sorting: It seems sometimes the table is sorted in reverse
- [ ] RPC tx confirmation detection times seem variable: Sometimes detection of tx confirmation on optimism/arbitrum is  instant, other times it takes up to 15 seconds. There seems to be no correlation between the tx action and time, so I suspect this is RPC/sequencer related and likely not much we can do about it, but worth looking into a bit

contracts/.solhint.json
apbendi: Have not seen this on either network in my testing 👍

contracts/.solhint.json
apbendi: Looks good 👍

contracts/.solhint.json
apbendi: * toPrecision error is gone
* Not seeing the unsupported block number error anymore, at least in today's testing 🤷‍♂️ 
* Understood on the toast issue— we can tackle that separately I think

contracts/.solhint.json
apbendi: No RPC errors and Arbitrum withdrawals now leave very little dust 👍

contracts/.solhint.json
apbendi: So weird. I saw this a bit testing last week, but can't reproduce it today. My guess is it's a race-condition issue of some kind, and related to web workers. I think make an issue for this one and we can tackle it at some point later.

contracts/.solhint.json
apbendi: > functionality using both EOAs that have sent transactions in the past and new EOAs

When you say "new EOAs" here do you mean making sure that EOAs that have never sent a tx cannot receive an advanced mode tx? If so, I confirmed this is the case.

I also tested Advanced Mode on both networks with an account that sent its first tx on both networks today. Both worked as expected without issue!

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
...
+    "compiler-version": ["warn", "^0.6.12"],
     "const-name-snakecase": "off",
     "constructor-syntax": "error",
     "func-visibility": ["error", { "ignoreConstructors": true }],
```
apbendi, line 9: (arbitrary comment location)

The rounding doesn't properly show the toll in the total on the send page. Maybe if we use the same rounding function used on the withdrawal page for fees, we'll get what we need here?

Screen Shot 2022-03-25 at 4 26 40 PM

contracts/.solhint.json
```diff
@@ -3,11 +3,10 @@
   "plugins": ["prettier"],
   "rules": {
     "code-complexity": ["error", 7],
-    "compiler-version": ["error", "^0.6.12"],
+    "compiler-version": ["warn", "^0.6.12"],
     "const-name-snakecase": "off",
     "constructor-syntax": "error",
```
apbendi, line 8: (arbitrary location)

There seems to be an edge case when you have both a long network name (Arbibtrum One) and a long ENS name, the wrapping gets weird and thrown off for both of them. Not sure how hard of a fix this one is but if it's possible to fix easily we should— otherwise issue is fine.

Screen Shot 2022-03-25 at 4 51 28 PM

apbendi (commented):
Did a bunch of testing and it was quite rock solid. Most issues resolved, and things working fast and as expected. Left a couple small things I noticed, but otherwise it feels good.

As for the remaining issues you identified:

> Withdraw button should show loading spinner/disabled: There's a delay from async calls between clicking withdraw and seeing the modal, so we should disable the button in that time

Hmm yeah, this feels like it would be straightforward and if there was an issue that caused the response to be delayed, it would be nice to have it in place. If it's pretty quick maybe we bang it out. If it's more complicated than it seems then issue is OK.

> Fix table sorting: It seems sometimes the table is sorted in reverse

Weird, I've seen this before, but now can't reproduce. Do you have a consistently reproduce-able case? I'll do some more testing to try to get to the bottom of it. If not, an issue seems fine for now. Since it might be a race condition it could be tricky to figure out.

> RPC tx confirmation detection times seem variable: Sometimes detection of tx confirmation on optimism/arbitrum is instant, other times it takes up to 15 seconds. There seems to be no correlation between the tx action and time, so I suspect this is RPC/sequencer related and likely not much we can do about it, but worth looking into a bit

I think an issue is fine for this one

contracts/.solhint.json
mds1: > I also tested Advanced Mode on both networks with an account that sent its first tx on both networks today. Both worked as expected without issue!

Ok great! We should also test advanced mode with an EOA that has sent txs as long ago as possible, without recent ones, to test the opposite case (in case there's new/old tx types on either network)—and then this would cover the "EOA tests" my unclear comment referred to. This may not be too practical depending on our past usages

contracts/.solhint.json
mds1: Tracking in https://github.com/ScopeLift/umbra-protocol/issues/316

mds1:
> Withdraw button should show loading spinner/disabled: There's a delay from async calls between clicking withdraw and seeing the modal, so we should disable the button in that time

This was surprisingly tedious, because we (in my opinion, and this is my fault) rely too much on passing props between some tightly coupled components. Trying to do this with props probably would have been messy, so I created a new store called `statuses` which we can use over time to gradually migrate passing of state/statuses like this to a shared store. Implemented in 5df94b3

contracts/.solhint.json
mds1: Just pushed a81667f which makes the header columns floating width instead of fixed fractions of browser width. Seems to fix the issue for me

davidlaprade:
I'm going to push another improvement to rounding (hopefully the last we need).

Late yesterday I realized that the reason why rounding has been so challenging is because we've been leaving out an important piece of information: the subtotal. You really can't know the correct number of decimals and sig-figs to show unless you know the number of decimals and sig-figs for all operands in the equation.

Intuitively, when you're looking at basic arithmetic, you increase the precision of the resulting number to the precision of the most precise operand, and sometimes that is the larger number (in our case: the subtotal), e.g.

```
  12.001   (subtotal)
+  0.01    (fee)
  ======
  12.011   (total, gets its precision from the subtotal)
```

davidlaprade:
Okay, here's the new rounding in action:

better-frontend-native-token-rounding

contracts/.solhint.json
apbendi: hey @davidlaprade, rounding is looking great here, but now on the send page, if I type a decimal w/o the leading zero, I get an error. This seems to be a regression.

Screen Shot 2022-04-01 at 1 31 05 PM

Screen Shot 2022-04-01 at 1 31 25 PM

davidlaprade: Nice catch, thanks! Fixed with https://github.com/ScopeLift/umbra-protocol/pull/293/commits/227c5e8ba03799e35473b83379477404eab2d501

davidlaprade:
@mds1 

> Question for @apbendi and @davidlaprade: What changes, if any, do I need to make based on server-side changes around how we determine minimum send amounts?

My feeling is that you can just hardcode the minimums (0.01 ETH for both networks) and we can take care of this when we switch over to backend estimations in another PR

contracts/.solhint.json
apbendi: Working now, thanks @davidlaprade

contracts/.solhint.json
apbendi: I tested out what happens if you try to send to an account that has never sent a transaction, except for receiving funds over the bridge. (0x22222B7C879e6c09690d1E5Eb10437E68987882c)

On Optimism, it simply says this account has never done a txn, which is good.

On Arbitrum, you get this validation error:

Screen Shot 2022-04-01 at 5 04 16 PM

As long as we feel confident this will always be what happens for such addresses, I think it's fine. It's an edge case, and users are not able to send funds so there's no harm. What do you think @mds1?

mds1: Just pushed ad9f879, so pub key recovery will now work on Arbitrum. Details on the limitations and future improvements to remove those limitations is being tracked in https://github.com/ScopeLift/umbra-protocol/issues/320

mds1: Should be fixed with 1df8b62

PR #296, opened 2022-02-19
Render ENS Avatars

### LINKED ISSUE #248: app: use ENS avatars in place of jazzicons if the user has them set
Context/references:
- [Step-by-Step Guide to Setting an NFT as your ENS Profile Avatar](https://medium.com/the-ethereum-name-service/step-by-step-guide-to-setting-an-nft-as-your-ens-profile-avatar-3562d39567fc)
- [ENS as your web3 username and more](https://vimeo.com/showcase/8950429/video/637518644)
- https://github.com/ethers-io/ethers.js/issues/2185
- use ethers v5.5's [`provider.getAvatar()`](https://github.com/ethers-io/ethers.js/blob/d3079745c5c5c41e3ce26d7632b8a6657d91d0b5/packages/providers/src.ts/base-provider.ts#L1734-L1756)

davidlaprade:
Fixes #248 

When there is an [ENS avatar](https://app.ens.domains/name/nick.eth/details) associated with the wallet:

image

When there isn't an ENS avatar, we fall back to the jazzicon:

image

When we're on a chain that ENS isn't on, e.g. Polygon, we fallback to whatever avatar is on mainnet (if any):

image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,7 +23,7 @@
     "bnc-notify": "^1.5.1",
     "bnc-onboard": "^1.13.0",
     "core-js": "^3.6.5",
-    "ethers": "^5.5.3",
+    "ethers": "5.5",
     "quasar": "^1.0.0"
   },
   "devDependencies": {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,7 +23,7 @@
     "bnc-notify": "^1.5.1",
     "bnc-onboard": "^1.13.0",
     "core-js": "^3.6.5",
-    "ethers": "5.5",
+    "ethers": "^5.5.3",
     "quasar": "^1.0.0"
   },
   "devDependencies": {
--- a/frontend/src/layouts/AddressSettings.vue
+++ b/frontend/src/layouts/AddressSettings.vue
@@ -8,22 +8,30 @@
         <span v-if="userDisplayName">
           {{ userDisplayName }}
         </span>
-        <Jazzicon v-if="userAddress" :address="userAddress" class="q-ml-sm" />
+        <span v-if="userAddress">
+          <Avatar v-if="avatar" :avatar="avatar" class="q-ml-sm" />
+          <Jazzicon v-else :address="userAddress" class="q-ml-sm" />
+        </span>
       </div>
     </connect-wallet>
   </div>
 </template>
 
 <script lang="ts">
-import { defineComponent } from '@vue/composition-api';
+import { defineComponent, PropType } from '@vue/composition-api';
 import BaseTooltip from 'src/components/BaseTooltip.vue';
 import ConnectWallet from 'components/ConnectWallet.vue';
 import Jazzicon from 'src/components/Jazzicon.vue';
+import Avatar from 'src/components/Avatar.vue';
 
 export default defineComponent({
   name: 'AddressSettings',
-  components: { BaseTooltip, ConnectWallet, Jazzicon },
+  components: { BaseTooltip, ConnectWallet, Jazzicon, Avatar },
   props: {
+    avatar: {
+      type: (null as unknown) as PropType<string | null>,
+      required: true,
+    },
     userDisplayName: {
       type: String,
       required: true,
--- a/frontend/src/layouts/AddressSettings.vue
+++ b/frontend/src/layouts/AddressSettings.vue
@@ -8,10 +8,7 @@
         <span v-if="userDisplayName">
           {{ userDisplayName }}
         </span>
-        <span v-if="userAddress">
-          <Avatar v-if="avatar" :avatar="avatar" class="q-ml-sm" />
-          <Jazzicon v-else :address="userAddress" class="q-ml-sm" />
-        </span>
+        <Avatar v-if="userAddress" :address="userAddress" :avatar="avatar" class="q-ml-sm" />
       </div>
     </connect-wallet>
   </div>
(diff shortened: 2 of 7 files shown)


frontend/src/store/wallet.ts
```diff
@@ -243,6 +244,24 @@ export default function useWalletStore() {
...
+      if (typeof _userEns === 'string') { // ENS address must exist
+        const chainId = provider.value.network.chainId;
+        switch (chainId) {
+          case 1: // mainnet
+          case 137: // polygon mainnet
+            avatar.value = await MAINNET_PROVIDER.getAvatar(_userEns);
+            break;
+          case 4: // rinkeby
+            const currentNetworkAvatar = await provider.value.getAvatar(_userEns);
+            const mainnetAvatar= await MAINNET_PROVIDER.getAvatar(_userEns);
```
mds1, line 256: Running `yarn lint` in the frontend returns errors on these two lines:

```
/Users/mds/Documents/projects/umbra/umbra-protocol/frontend/src/store/wallet.ts
  255:13  error    Unexpected lexical declaration in case block  no-case-declarations
  256:13  error    Unexpected lexical declaration in case block  no-case-declarations
```

From https://eslint.org/docs/rules/no-case-declarations:

> This rule disallows lexical declarations (`let`, `const`, `function` and `class`) in `case`/`default` clauses. The reason is that the lexical declaration is visible in the entire switch block but it only gets initialized when it is assigned, which will only happen if the case where it is defined is reached.
>
> To ensure that the lexical declaration only applies to the current case clause wrap your clauses in blocks.

So, let's just wrap the case blocks in braces

mds1: Actually, a few lines up you'll see `lookupEnsName` where we pass a mainnet provider, i.e. we show the mainnet name even when connected on Rinkeby. Therefore we should probably be consistent and not special-case Rinkeby here. Which means we can probably get rid of the whole switch-case and just always use mainnet resolution. 

The current switch-case approach also means we'd need to add another step to adding support for additional networks

frontend/src/layouts/AddressSettings.vue
```diff
@@ -8,22 +8,30 @@
         <span v-if="userDisplayName">
           {{ userDisplayName }}
         </span>
-        <Jazzicon v-if="userAddress" :address="userAddress" class="q-ml-sm" />
+        <span v-if="userAddress">
+          <Avatar v-if="avatar" :avatar="avatar" class="q-ml-sm" />
+          <Jazzicon v-else :address="userAddress" class="q-ml-sm" />
```
mds1, line 13: What do you think about deleting the `Jazzicon.vue` component and having a single `Avatar.vue` component, to reduce the number of files? Arguably overkill right now, but anywhere we want to show the user's image we'd now need this logic, so we can simplify it with a single component that takes address or avatar and shows the right image accordingly

mds1 (commented):
Working great overall!

frontend/src/store/wallet.ts
davidlaprade: > Running yarn lint in the frontend returns errors

My bad. I checked for typescript errors but forgot we used eslint too

> a few lines up you'll see lookupEnsName where we pass a mainnet provider, i.e. we show the mainnet name even when connected on Rinkeby

Ahh, yeah, good point. I'll just get rid of the switch 👍

frontend/src/layouts/AddressSettings.vue
davidlaprade: I like that approach, good call 👍

frontend/package.json
```diff
@@ -23,7 +23,7 @@
     "bnc-notify": "^1.5.1",
     "bnc-onboard": "^1.13.0",
     "core-js": "^3.6.5",
-    "ethers": "^5.5.3",
+    "ethers": "5.5",
```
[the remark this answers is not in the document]
davidlaprade: I'll be honest, I'm not sure why this changed either. Initially I was not going to commit it, but then later I thought it was okay b/c we need 5.5.3 of the ethers _providers_ library, not the whole ethers library

apbendi (approved):
Looks and works great overall. Nice job!

I've approved this, but one nice-to-have, iff it isn't difficult:

Would it be possible to show the jazzicon while the avatar itself is loading? On first load of my new avatar, it took a long time for it to render in. Would be nice if Jazzicon was there, then switched to Avatar once it was available. If this would be difficult to implement, then don't sweat it, and we can rebase and merge!

davidlaprade:
Great idea!

avatar-async-loading

apbendi:
Awesome! Thanks @davidlaprade. This lazy loading works great. Thanks for making that change.

I'm now seeing one typescript build error when I run `yarn build` in preparation for deployment:

```
ERROR in umbra-protocol/frontend/src/components/Avatar.vue(33,7):
TS2322: Type 'string' is not assignable to type 'number'.

 App · ⚠️  [FAIL] Build failed with 1 error. Check log above.
```

mds1:
Kind of bikeshedding so maybe we move this to a separate issue, but initial reaction is that I'm not a fan of the jazzicon as a placeholder while avatar loads. I think the rationale behind this is that avatars might be slow to load depending on where they're hosted, such as on IPFS? So given that we know they have an avatar, why not just use a loading spinner or a placeholder image (such as [this](https://fontawesome.com/icons/user?s=solid) with a circular border) instead until the avatar loads? Some users might not know what jazzicons are, and the jumpy UX as it changes—without any explanation of why it changed—isn't ideal IMO. In other words, I think there should be a 1:1 mapping of profile picture to user address, whereas this is a 2:1 mapping

apbendi:
> So given that we know they have an avatar, why not just use a loading spinner or a placeholder image (such as [this](https://fontawesome.com/icons/user?s=solid) with a circular border) instead until the avatar loads?

I like the idea of the placeholder image. That said I think we're at risk of overthinking this detail 😅 . But I don't feel strongly about it— it just felt weird to have this empty space up there for 5-7 seconds while my pfp loaded in from IPFS the first time.

> Some users might not know what jazzicons are, and the jumpy UX as it changes—without any explanation of why it changed—isn't ideal IMO

Fair enough, but given most users will see jazzicons this would be a more generic problem if users were confused by them. Also, I'm guessing most users sophisticated enough to set an ENS PFP have probably seen a JazzIcon!

Like I said, don't feel strongly here. Should we move this to a different issue, update to use the placeholder, or leave as-is? Let me know your thoughts @mds1 @davidlaprade

mds1:
Since we agree it's bikeshedding, let's leave it for now and I'll create another issue to track + link here!

mds1 (approved):
Ok, tracking in https://github.com/ScopeLift/umbra-protocol/issues/299 and gave it a backlog label since it's not too important. I think this is good to merge now?

PR #303, opened 2022-03-02
Deploy Contracts to Optimism and Arbitrum

apbendi:
This PR:

* Updates the deployment script to:
    *  Deploy Umbra
    * Bump nonce to the number needed to keep Registry address consistent
    * Deploy the Stealth Key Registry
* Configures and documents deployments for Optimism and Arbitrum, and their respective test networks

All deployments are now completed and contracts are verified with matching address on the networks in question.


contracts/.env.example
```diff
@@ -1,4 +1,4 @@
 INFURA_ID=zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
 MNEMONIC=here is where your twelve words mnemonic should be put my friend
 DEPLOY_GSN=false
-ETHERSCAN_API_KEY="YOUR_API_KEY"
\ No newline at end of file
+ETHERSCAN_VALIDATION_API_KEY="YOUR_API_KEY"
```
mds1, line 4: just to be more precise on the use case for the API key
```suggestion
ETHERSCAN_VERIFICATION_API_KEY="YOUR_API_KEY"
```

contracts/deploy-history/arbitrum-2022-03-02T18:56:10.776Z.json
```diff
@@ -0,0 +1,8 @@
+{
+  "deployer": "0x2217432e5E0e2D77576461662d267bd293820811",
+  "contracts": {
+    "Umbra": "0x2C943761FEA42Eaca0AfE6064A6daB935fB5b18a",
```
mds1, line 4: Seems there's a few deploy files where the contracts were deployed to the wrong address, so we should probably delete those to avoid confusion in case any users poke around. Guessing you were doing a test deployment, but I don't think we need to care about the fact these exist and don't need to keep the files. IMO we should only keep the real deploy files that we care about

And actually, since every "real" file is the same, we can simplify this whole `deploy-history` folder into a single `deploys.json` file that maps chain ID to deploy addresses. But we don't need to do that in this PR

contracts/hardhat.config.ts
```diff
@@ -108,6 +112,46 @@ const config: HardhatUserConfig = {
       url: 'https://polygon-rpc.com/',
       gasPrice: 33000000000, // 33 gwei
     },
+    rinkarby: {
+      accounts: {
+        count: 10,
+        initialIndex: 0,
+        mnemonic,
+        path: "m/44'/60'/0'/0",
+      },
```
mds1, line 121: Might be nice to define the account config into a standalone const and reuse it for each network to make it more DRY, reduce copy/paste errors, and a bit more readable by shortening the file

contracts/scripts/deploy.js
```diff
@@ -1,24 +1,13 @@
-/**
- * @notice Umbra Deployment script
- * @dev At each contract deployment or transaction, we wait for the transaction to be mined before
- * continuing to the next step. This is done because it makes it simpler to continue from that spot
- * rather than restart the full deployment
- * @dev To initialize a deploy:
- *   `yarn deploy --network <network>`   (where network specifies a netwrk found in the hardhat.config.ts file)
- * If deploying to a local node (--network localhost), first start Hardhat using `yarn hardhat node`
- */
```
mds1, line 9: Why were these comments deleted? If they're outdated we could update them instead

contracts/scripts/deploy.js
```diff
@@ -44,89 +33,73 @@ const save = (value, field, subfield = undefined) => {
   fs.writeFileSync(fileName, JSON.stringify(parameters));
 };
 
+// Send 0.01 ETH to oneself and wait for it to be mined, bumping the wallet's nonce
+const doBumpNonceTx = async (wallet) => {
+  const tx = await wallet.sendTransaction({
+    to: wallet.address,
+    value: ethers.utils.parseEther('0.01'),
```
mds1, line 40: You could just send zero ETH, which I think is more common and clear. But doesn't make a difference

contracts/scripts/deploy.js
```diff
@@ -44,89 +33,73 @@ const save = (value, field, subfield = undefined) => {
...
 
     console.log('Deploying to: ', network);
     save(network, 'actions', 'DeployingContractsToNetwork');
 
-    const [adminWallet] = await ethers.getSigners();
-    save(adminWallet.address, 'admin');
+    const [deployerWallet] = await ethers.getSigners();
+    save(deployerWallet.address, 'deployer');
+
+    if (deployerWallet.address !== expectedDeployer) {
```
mds1, line 66: making sure to compare checksummed addresses since the address in the JSON file could be all lowercase
```suggestion
    if (deployerWallet.address !== ethers.utils.getAddress(expectedDeployer)) {
```

contracts/scripts/deploy.js
```diff
@@ -44,89 +33,73 @@ const save = (value, field, subfield = undefined) => {
...
+      throw new Error(`Unexpected deployer address. Found ${deployerWallet.address}, expected ${expectedDeployer}.`);
+    }
 
-    // deploy the Umbra contracts
-    const Umbra = await ethers.getContractFactory('Umbra', adminWallet);
+    let deployerNonce = await deployerWallet.getTransactionCount();
+
+    if (deployerNonce !== 0) {
+      throw new Error('Unexpected non-zero nonce before deploying Umbra');
+    }
```
mds1, line 74: Why does the nonce need to be non-zero? Seems we have logic to bump it if it's too low

contracts/scripts/deploy.js
```diff
@@ -44,89 +33,73 @@ const save = (value, field, subfield = undefined) => {
...
-        gasLimit: 1000000,
-      };
-      const receipt = await adminWallet.sendTransaction(tx);
-      await receipt.wait();
-      save(receipt.hash, 'actions', 'SendPaymasterFundsTx');
-      console.log(`Paymaster funding transaction successful with hash: ${receipt.hash}`);
+    // bump the nonce until the level needed for the registry
+    deployerNonce = await deployerWallet.getTransactionCount();
+
+    while (deployerNonce < expectedRegistryNonce) {
```
mds1, line 90: Might be good to add a check before this line:

```typescript
if (deployerNonce >= expectedRegistryNonce) {
    throw new Error(`Nonce too high for ${deployerWallet.address}: Nonce is ${deployerNonce}, need ${expectedRegistryNonce}`);
}
```

mds1 (approved):
LGTM, all comments are optional small things!

contracts/scripts/deploy.js
apbendi: I think I removed them in error while moving stuff around. Restored and updated in 9ab9ccf0bcb3a48e38ab6788d369c0df57e4f2a7

contracts/scripts/deploy.js
apbendi: Good catch. Done in 59a020255c75f7f0a10324474e7d550e5d30f942

contracts/scripts/deploy.js
apbendi: This check ensures the nonce *is* `0`. In order to get the same contract addresses across networks for our deployer address, we need:

* Umbra deployed with nonce 0
* The Stealth Key Registry deployed with nonce 5

This is the first half of that check, before deploying Umbra.

mds1: Oops misread, yep that makes sense

contracts/deploy-history/arbitrum-2022-03-02T18:56:10.776Z.json
apbendi: Given we timestamp the deployments and keep the `-latest` version up to date automatically, I don't see much harm in keeping around the records of the test deployments. They could even come in handy for testing or something in the future and this gives us an easy way to know where they are.

> And actually, since every "real" file is the same, we can simplify this whole deploy-history folder into a single deploys.json file that maps chain ID to deploy addresses. But we don't need to do that in this PR

Are you looking for something you can import into the frontend codebase or something along those lines? If so, I'm not opposed to creating it, but I think it could/should be separate from the deployment history. We could put a `json` file in the root of the contract dir, for example.

contracts/hardhat.config.ts
apbendi: Good idea. Tracking in https://github.com/ScopeLift/umbra-protocol/issues/305 since I don't have time to do it right now, and I want to unblock t his PR.

contracts/deploy-history/arbitrum-2022-03-02T18:56:10.776Z.json
mds1: >  I don't see much harm in keeping around the records of the test deployments. They could even come in handy for testing or something in the future and this gives us an easy way to know where they are.

`-latest` does not imply production deploy though, just that it was the most recent deploy, so it still just feels confusing to have useless non-production deploys in this folder — what's the canonical way to find the production deploy? If we keep this, it seems the answer would be the FAQ or umbra-js code, but it feels like that should be somewhere in the contracts folder

> Are you looking for something you can import into the frontend codebase or something along those lines? If so, I'm not opposed to creating it, but I think it could/should be separate from the deployment history. We could put a json file in the root of the contract dir, for example.

Nope, just looking to make things cleaner and easier to read. This folder looks extremely cluttered given that there's just a handful of total addresses (across all networks) that actually matter
image

mds1: Tracking in https://github.com/ScopeLift/umbra-protocol/issues/306 to unblock this PR

PR #304, opened 2022-03-07
Improve frontend token rounding 

### LINKED ISSUE #283: Implement a Saner Rounding Strategy and Apply It Across the App
Different currencies/amounts/tokens require different rounding strategies to provide useful numbers to humans. For example, stablecoin amounts should almost always be rounded to 2 decimals, while something like a fee of `0.000456748949` WETH needs to be rounded to something like `0.00046 WETH` for human readability.

A possible algorithm for this is something like:

```
if < 1, round to two sig figs, if > 1, round to two decimals
```

Or, roughly (in untested code):

```typescript
// Show two sig figs if number is <= 1, and show two decimals otherwise
export function formatAmount(amount: BigNumberish, decimals: number) {
  const formattedAmount = parseFloat(formatUnits(amount, decimals))
  // I think toLocaleString with maximumFractionDigits=2 should also work instead of toPrecision
  if (formattedAmount <= 1) return formattedAmount.toPrecision(2) 
  return formattedAmount.toLocaleString(undefined, {
      minimumFractionDigits: 0,
      maximumFractionDigits: 10,
    })
}
```

We should implement this algorithm properly and use it throughout the app, effectively replacing and unifying all current usage of the `round` method or `roundTokenAmount` method.

See [here](https://github.com/ScopeLift/umbra-protocol/pull/279#discussion_r789690434) and [here](https://github.com/ScopeLift/umbra-protocol/pull/279#discussion_r789691361) for further discussion.

davidlaprade:
Fixes #283

I believe this is basically a no-op from the user's perspective, but here's confirmation that the change works:

umbra-WETH-rounding

umbra-DAI-token-rounding

### DIFF
--- a/frontend/babel.config.js
+++ b/frontend/babel.config.js
@@ -1,4 +1,8 @@
 /* eslint-env node */
 module.exports = {
-  presets: ['@quasar/babel-preset-app'],
+  presets: [
+    '@babel/preset-env',
+    '@quasar/babel-preset-app',
+    '@babel/preset-typescript',
+  ],
 };
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -27,6 +27,7 @@
     "quasar": "^1.0.0"
   },
   "devDependencies": {
+    "@babel/preset-typescript": "^7.16.7",
     "@quasar/app": "^2.0.0",
     "@quasar/quasar-app-extension-dotenv": "^1.0.5",
     "@types/node": "^12.17.15",
@@ -35,6 +36,7 @@
     "babel-eslint": "^10.0.1",
     "eslint-loader": "^3.0.3",
     "eslint-plugin-vue": "^6.1.2",
+    "jest": "^27.5.1",
     "postinstall-postinstall": "^2.1.0",
     "worker-loader": "^3.0.8"
   },
@@ -51,5 +53,10 @@
   ],
   "volta": {
     "extends": "../package.json"
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "src/(.*)": "<rootDir>/src/$1"
+    }
   }
 }
--- a/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
+++ b/frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
@@ -108,7 +108,7 @@ import { UserAnnouncement } from '@umbra/umbra-js';
 import { FeeEstimate } from 'components/models';
 import { formatAddress, toAddress } from 'src/utils/address';
 import { BigNumber, formatUnits } from 'src/utils/ethers';
-import { getEtherscanUrl, getGasPrice, round, roundTokenAmount } from 'src/utils/utils';
+import { getEtherscanUrl, getGasPrice, round, humanizeTokenAmount } from 'src/utils/utils';
 import useWalletStore from 'src/store/wallet';
 
 export default defineComponent({
@@ -164,7 +164,7 @@ export default defineComponent({
     if (isNativeToken) {
       formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
     } else {
-      formattedAmount = roundTokenAmount(amount, props.activeFee.token);
+      formattedAmount = humanizeTokenAmount(amount, props.activeFee.token);
     }
     function isValidFeeAmount(val: string) {
       if (!val || !(Number(val) > 0)) return 'Please enter an amount';
@@ -233,7 +233,7 @@ export default defineComponent({
       if (isNativeToken) {
         round(formatUnits(fee.value, decimals), ethDisplayDecimals);
       } else {
-        return roundTokenAmount(props.activeFee.fee, props.activeFee.token);
+        return humanizeTokenAmount(props.activeFee.fee, props.activeFee.token);
       }
       return round(formatUnits(fee.value, decimals), numDecimals);
     }); // relayer fee, rounded
@@ -241,7 +241,7 @@ export default defineComponent({
       if (isNativeToken) {
         return round(formatUnits(amountReceived.value, decimals), numDecimals);
       } else {
-        return roundTokenAmount(amountReceived.value, props.activeFee.token);
+        return humanizeTokenAmount(amountReceived.value, props.activeFee.token);
       }
     }); // amount user will receive, rounded
     // prevent withdraw attempts if fee is larger than amount
--- a/frontend/src/utils/utils.ts
+++ b/frontend/src/utils/utils.ts
@@ -30,22 +30,6 @@ export const round = (value: number | string, decimals = 2) => {
 
 /**
  * @notice Rounds to appropriate human readable decimals for the token
- * @dev This is very fragile and should be replace with something robust, i.e. something that walks
- * value and rounds considering sig-figs or similar
- */
-export const roundTokenAmount = (amount: BigNumberish, token: { decimals: number; address: string }): string => {
-  const formatted = formatUnits(amount, token.decimals);
-  // WETH on Polygon
-  const isWeth = token.address === '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619';
-  if (isWeth) {
-    return round(formatted, 6);
-  } else {
-    return round(formatted, 2);
-  }
-};
-
-/**
- * @notice Rounds to appropriate human readable decimals for the token
  * @param amount
  * @param token
  * @returns string
--- a/frontend/src/utils/utils.ts
+++ b/frontend/src/utils/utils.ts
@@ -45,6 +45,24 @@ export const roundTokenAmount = (amount: BigNumberish, token: { decimals: number
 };
 
 /**
+ * @notice Rounds to appropriate human readable decimals for the token
+ * @param amount
+ * @param token
+ * @returns string
+ */
+export const humanizeTokenAmount = (
+  amount: BigNumberish,
+  token: { decimals: number; address: string },
+): string => {
+  const formattedAmount = parseFloat(formatUnits(amount, token.decimals));
+  if (formattedAmount <= 1) return formattedAmount.toPrecision(2);
+  return formattedAmount.toLocaleString(undefined, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 2,
+  })
+};
+
+/**
  * @notice GETs JSON from the provided `url`
  */
 export const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
--- a/frontend/test/utils.test.ts
+++ b/frontend/test/utils.test.ts
@@ -1,0 +1,34 @@
+import { humanizeTokenAmount } from '../src/utils/utils';
+import { BigNumber } from 'ethers';
+import { parseUnits } from "@ethersproject/units";
+
+describe('Utilities', () => {
+  describe('humanizeTokenAmount', () => {
+    const usdc = { decimals: 6, address: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48" }
+    const weth = { decimals: 18, address: "0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619" }
+    it("rounds USDC to 2 decimal places when > 1", () => {
+      const amount = parseUnits("42.11", usdc.decimals);
+      expect(humanizeTokenAmount(amount, usdc)).toEqual("42.11");
+    });
+    it("rounds USDC to 2 decimal places when < 1", () => {
+      const amount = parseUnits(".419", usdc.decimals);
+      expect(humanizeTokenAmount(amount, usdc)).toEqual("0.42");
+    });
+    it("does not round USDC when a whole number", () => {
+      const amount = parseUnits("42", usdc.decimals);
+      expect(humanizeTokenAmount(amount, usdc)).toEqual("42");
+    });
+    it("rounds WETH to 2 sig figs when < 1", () => {
+      const amount = parseUnits("0.007581732", weth.decimals);
+      expect(humanizeTokenAmount(amount, weth)).toEqual("0.0076");
+    });
+    it("rounds WETH to 2 decimal places when > 1", () => {
+      const amount = parseUnits("923.1928371", weth.decimals);
+      expect(humanizeTokenAmount(amount, weth)).toEqual("923.19");
+    });
+    it("does not round WETH when a whole number", () => {
+      const amount = parseUnits("53", weth.decimals);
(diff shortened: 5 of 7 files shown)


frontend/package.json
```diff
@@ -51,5 +53,10 @@
   ],
   "volta": {
     "extends": "../package.json"
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "src/(.*)": "<rootDir>/src/$1"
```
davidlaprade, line 59: Needed so that relative path imports in our application code work when run within jest tests

yarn.lock
```diff
@@ -16,11 +23,44 @@
...
   version "7.13.15"
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.13.15.tgz#7e8eea42d0b64fda2b375b22d06c605222e848f4"
   integrity sha512-ltnibHKR1VnrU4ymHyQ/CXtNXI6yZC0oJThyW78Hft8XndANwi+9H+UIklBDraIjFEJzw8wmcM427oDd9KS5wA==
 
+"@babel/compat-data@^7.16.4":
+  version "7.17.0"
+  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.17.0.tgz#86850b8597ea6962089770952075dcaabb8dba34"
+  integrity sha512-392byTlpGWXMv4FbyWw3sAZ/FrW/DrwqLGXpy0mbyNe9Taqv1mg9yON5/o0cnr8XYCkFTZbC1eV+c+LAROgrng==
+
+"@babel/core@^7.1.0", "@babel/core@^7.12.3", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
```
davidlaprade, line 43: Babel added a lot of dependencies :(

If there's a way to avoid adding it, while still having TS tests, I'm all ears

mds1: Sorry, why do we need Babel at all? The `umbra-js` package has TS tests but doesn't use babel

frontend/package.json
```diff
@@ -51,5 +53,10 @@
   ],
   "volta": {
     "extends": "../package.json"
+  },
+  "jest": {
```
mds1, line 57: Can we update the `test` script in this file so we can run tests with `yarn test`? Right now it's defined as `"test": "echo \"No test specified\" && exit 0",`. This way we can run all tests in the repo by running `yarn test` from the repo root

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -164,7 +164,7 @@ export default defineComponent({
     if (isNativeToken) {
       formattedAmount = round(formatUnits(amount, decimals), numDecimals); // amount being withdrawn, rounded
     } else {
-      formattedAmount = roundTokenAmount(amount, props.activeFee.token);
+      formattedAmount = humanizeTokenAmount(amount, props.activeFee.token);
     }
```
mds1, line 168: Why do we still need the `if` conditional for the native token everywhere, instead of simplifying it all to `formattedAmount = humanizeTokenAmount(amount, props.activeFee.token);`? There's a few parts that do 

```javascript
if (isNativeToken) {
  formattedAmount = round(formatUnits(...))
} else {
  formattedAmount = humanizeTokenAmount(...)
}
```

which is what I'm referring to

frontend/src/utils/utils.ts
```diff
@@ -30,18 +30,20 @@ export const round = (value: number | string, decimals = 2) => {
...
-  // WETH on Polygon
-  const isWeth = token.address === '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619';
-  if (isWeth) {
-    return round(formatted, 6);
-  } else {
-    return round(formatted, 2);
-  }
+export const humanizeTokenAmount = (
+  amount: BigNumberish,
+  token: { decimals: number; address: string },
```
mds1, line 39: Can we change the input here to type `TokenInfo` instead? This way it includes all properties of the token in case we ever want to have custom formatting overrides (e.g. override DAI formatting on all chains by checking the symbol instead of the address, since symbol would be constant by address varies)

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
davidlaprade: Great point, I think you're right that this can be simplified. Nice

yarn.lock
davidlaprade: > The umbra-js package has TS tests but doesn't use babel

Yeah, good point. Seems like umbra-js runs the tests through hardhat. I'm not sure how that works with TS.

When I initially tried to run the TS tests, they all failed with Syntax errors b/c none of the typescript of ES6 idioms were being interpretted. I eventually found [this](https://jestjs.io/docs/getting-started#using-typescript) in the Jest docs about using TS with Jest, and it said that you should install `@babel/preset-typescript`, so I just blindly did that. But looking past the first paragraph of the docs, I see that there's another option, `ts-jest`, which seems much more lightweight. I'll try that.

frontend/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "esModuleInterop": true,
     "outDir": "build",
     "resolveJsonModule": true,
-    "rootDir": "src",
```
davidlaprade, line 10: Details in the commit message for why these settings were changed: https://github.com/ScopeLift/umbra-protocol/pull/304/commits/6610b9a5465cd0729066e4b5b08ebf57f1d55deb

yarn.lock
mds1: Seems to work well! Any idea what this warning is? I don't see any async calls in the tests
image

frontend/babel.config.js
```diff
@@ -1,4 +1,6 @@
 /* eslint-env node */
 module.exports = {
-  presets: ['@quasar/babel-preset-app'],
-};
+  presets: [
+    '@quasar/babel-preset-app',
+  ],
+}
```
mds1, line 6: nit, but this is now just a formatting change so we should revert it

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -191,9 +186,9 @@ export default defineComponent({
     const customFeeInWei = computed(() => {
       return gasLimit.value.mul(customGasInWei.value);
     });
-    const formattedCustomFeeEth = computed(() =>
-      round(formatUnits(customFeeInWei.value, decimals), ethDisplayDecimals)
-    );
+    const formattedCustomFeeEth = computed(() => {
+      return humanizeTokenAmount(customFeeInWei.value, props.activeFee.token);
+    });
```
mds1, line 191: nit, but my personal preference (and therefore frontend convention in this repo) is to keep this on one line since it's <120 chars long (not sure why it wasn't like that before)

davidlaprade: Really it auto-breaks at 80 chars. Split screen vim has to wrap after 95 chars

120 character lines are inhumane. I will die on this hill.

mds1: use a wider monitor it's 2022! 😛

yarn.lock
davidlaprade: Chatted about this in standup 👍

mds1 (approved):

apbendi (commented):
Hey @davidlaprade, when I go to withdraw Matic, I'm seeing the fee rounded to 0 instead of showing appropriate decimals.

matic-rounding-0

In other places, things seem to be looking good.

apbendi:
@davidlaprade hmm I don't think that's quite right. On the prior screen, we show the fee is 0 MATIC, but that's referring to OUR relayer fee. In the modal, what's being displayed is the *gas fee* that will be charged when the transaction is submitted. The real number is *not* zero, as you can see when you click the button to customize it.

apbendi:
@davidlaprade here's what this looks like on the live site, for comparison:

Screen Shot 2022-03-10 at 3 05 56 PM

davidlaprade:
Okay, guys, thanks for bearing with me. I made a bunch of changes to hopefully make the code clearer and easier to reason about in the future. (Largely just fixing bad/ambiguous variable names that I picked when adding custom fee support.) Here's how things look like at this point:

umbra-rounding-fix

image

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -108,7 +108,12 @@ import { UserAnnouncement } from '@umbra/umbra-js';
 import { FeeEstimate } from 'components/models';
 import { formatAddress, toAddress } from 'src/utils/address';
 import { BigNumber, formatUnits } from 'src/utils/ethers';
-import { getEtherscanUrl, getGasPrice, round, roundTokenAmount } from 'src/utils/utils';
+import {
+  getEtherscanUrl,
+  getGasPrice,
+  humanizeTokenAmount,
+  roundReceivableAmountAfterFees,
+} from 'src/utils/utils';
```
mds1, line 116: If I run `yarn prettier` in the frontend folder, this gets collapsed to a single line (and there's a handful of other changes too, some in unrelated files). It should run in commit hooks, so odd that it didn't, but either let's remember to run prettier before merging this PR. (I'm intentionally not doing it now just to reduce diff noise from other, unrelated formatting changes)

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -182,18 +182,16 @@ export default defineComponent({
...
-      const customGasInGwei = formattedCustomFee.value ? formattedCustomFee.value : 0;
+    const customGasPriceInWei = computed(() => {
+      const customGasInGwei = formattedCustomTxCostGwei.value ? formattedCustomTxCostGwei.value : 0;
       return BigNumber.from(customGasInGwei).mul(10 ** 9);
     });
-    const customFeeInWei = computed(() => {
-      return gasLimit.value.mul(customGasInWei.value);
+    const customTxFeeInWei = computed(() => {
+      return gasLimit.value.mul(customGasPriceInWei.value);
     });
```
mds1, line 193: Resisting the urge to complain about this one liner taking up 3 lines 😝 (because it was already like that prior to this PR)

frontend/src/components/AccountReceiveTableWithdrawConfirmation.vue
```diff
@@ -228,34 +226,43 @@ export default defineComponent({
...
-    }); // relayer fee, rounded
+    const amountReceived = computed(() => amount.sub(useCustomFee.value ? customTxFeeInWei.value : fee.value)); // amount user will receive
+
+    // transaction fee, rounded
+    const formattedDefaultTxCost = computed(() => {
+      return humanizeTokenAmount(
+        (isNativeToken ? fee.value : props.activeFee.fee),
+        props.activeFee.token
+      )
+    });
```
mds1, line 237: Nit, but maybe a small refactor for readability?

```suggestion
    // transaction fee, rounded
    const txCost = isNativeToken ? fee.value : props.activeFee.fee;
    const formattedDefaultTxCost = computed(() => humanizeTokenAmount(txCost, props.activeFee.token));
```

frontend/src/utils/utils.ts
```diff
@@ -30,18 +30,43 @@ export const round = (value: number | string, decimals = 2) => {
 
 /**
  * @notice Rounds to appropriate human readable decimals for the token
- * @dev This is very fragile and should be replace with something robust, i.e. something that walks
- * value and rounds considering sig-figs or similar
+ * @param amount
+ * @param token
```
mds1, line 34: We should probably either add some words describing these inputs, or remove these two lines (I'm fine with either). As-is they just kinda take up space (since we'll never autogenerate docs for the frontend anyway, i.e. the NatSpec comments here are just for devs, not generated docs)

frontend/src/utils/utils.ts
```diff
@@ -30,18 +30,43 @@ export const round = (value: number | string, decimals = 2) => {
...
+    maximumFractionDigits: 2,
+  })
+};
+
+/**
+ * @notice Rounds the final amount that will be recieved during withdrawal.
+ * For display/UI purposes only -- it doesn't affect the actual amount
+ * withdrawn. Assumes that `amount` and `userFormattedFee` are both denominated in `token`.
+ * @param amount
+ * @param userFormattedFee
```
mds1, line 54: same as above

frontend/src/utils/utils.ts
```diff
@@ -30,18 +30,43 @@ export const round = (value: number | string, decimals = 2) => {
...
+ * @returns string
+ */
+export const roundReceivableAmountAfterFees = (
+  amount: BigNumberish,
+  userFormattedFee: string,
+  token: TokenInfo,
+): string => {
+  const humanizedFeeZeroTrimmed = userFormattedFee.replace(/0+$/, '');
```
mds1, line 62: What does this regex match/do? My check on https://regex101.com/ did not help. Wondering if `parseFloat` or `Number` would work as an alternative?

image

frontend/test/utils.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      const amount = parseUnits('0.0000', 'ether');
+      expect(humanizeTokenAmount(amount, eth)).toEqual('0.0');
+    });
+  });
+  describe('roundReceivableAmountAfterFees', () => {
+    it('should handle very high fees', () => {
+      //   50.0
+      // - 11.2
+      // ===========
+      //   38.8
```
mds1, line 76: Beautiful comments showing the frontend use case 👌

mds1 (commented):
Looking great!

frontend/src/utils/utils.ts
davidlaprade: It removes insignificant zeros from the user formatted fee. E.g. if the user formatted fee was `"0.01400"` it would trim this to `"0.014"`. `Number(userFormattedFee).toString()` would do the same thing -- you're right -- but has the downside of Javascript's Number overflow issues.

frontend/src/utils/utils.ts
davidlaprade: Did know this, thanks

frontend/src/utils/utils.ts
mds1: Mind just adding a comment explaining that? Otherwise sgtm

mds1 (approved):

apbendi (approved):
Working great! Nice work on this, @davidlaprade. Was trickier than it seemed at first, but the implementation we have now should make it easy to support any token we'd like.

Issue #298, opened 2022-02-28
Add batch send functionality


mds1:
Similar to zkSync checkout or https://disperse.app/, show users a table and allow them to enter multiple addresses, tokens, and amounts, then batch them all up into a single transaction

We can support this in the UI with a toggle for showing a mode similar to the one below. Then we'd write an `UmbraRouter` contract which has a method with a signature of `send(SendParams[] _params)` where the `SendParams` struct contains everything needed to call `Umbra.sendEth` and/or `Umbra.sendToken` in a loop. The flow would be:

1. UI shows user a table which they populate, similar to the one below
2. App generates the required random numbers, ephemeral pubkeys, ciphertext, etc. for each row in the table
3. User approves the `UmbraRouter` to send tokens as required
4. Use calls `UmbraRouter.send()` with all required calldata. This method then pulls the tokens from the user to the router, and in a loop calls the appropriate Umbra send method

Also need to decide on the best way for the router to handle approvals. Should there be an admin-only/public method to add new approvals? Or just do the approves as-required in the router? The former is cheaper for users, but ideally we don't need an admin and we'd need to ensure a public method is ok. The latter is more expensive for users, but results in a simpler contract with less attack surface

![image](https://user-images.githubusercontent.com/17163988/156021208-a2132244-018b-4e41-8eb7-80bc5386b87f.png)

garyghayrat:
Remaining tasks:
- [ ] Complete contract gas optimization
- [ ] Write contract deploy script
- [ ] Deploy contracts across relevant networks
- [ ] Create frontend interface for BatchSend

PR #311, opened 2022-03-29
Dapp batch send #298

garyghayrat:
@mds1


contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.7.6;
+
+contract UmbraRouter {
+  address internal constant umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
```
mds1, line 6: Recommend using an interface here for stronger type checking
```solidity
interface IUmbra {
  function sendEth(address payable receiver, uint256 tollCommitment, bytes32 pkx, bytes32 ciphertext) external payable;
  function sendToken(address receiver, address tokenAddr, uint256 amount, bytes32 pkx, bytes32 ciphertext) external payable
}

contract UmbraRouter {
  IUmbra internal constant umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
}
```

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
...
+   * @param _pkxes X-coordinate of the ephemeral public key used to encrypt the payload
+   * @param _ciphertexts Encrypted entropy (used to generated the stealth address) and payload extension
+   */
+  function batchSend(
+    address payable[] calldata _receivers,
+    address[] calldata _tokenAddrs,
+    uint256[] calldata _amounts,
+    uint256[] calldata _tollCommitments,
+    bytes32[] calldata _pkxes,
+    bytes32[] calldata _ciphertexts
```
mds1, line 24: Let's define a struct to clean things up and avoid having to verify that all array lengths are equal. We can have `SendEth` struct, a second `SendToken` struct, then a single `tollCommitment` argument (it will be the same for each ETH send, so no need to duplicate that). Then you can have the following function signatures:

```solidity
function batchSendEth(uint256 _tollCommitment, SendEth[] _params);
function batchSendTokens(SendToken[] _params);
function batchSend(uint256 _tollCommitment, SendEth[] _ethParams, SendToken[] _tokenParams);
```

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
...
+  ) external payable {
+    bytes memory payload;
+
+    for (uint256 i = 0; i < _receivers.length; i++) {
+      if (_tokenAddrs[i] == address(0x0)) {
+        payload = abi.encodeWithSignature(
+          "sendEth(address,uint256,bytes32,bytes32)",
```
mds1, line 31: Once we improve the interface, we can replace this with `abi.encodeWithSelector(IUmbra.sendEth.selector, ...)` which is safer than typing out the function signature

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
...
+    address[] calldata _tokenAddrs,
+    uint256[] calldata _amounts,
+    uint256[] calldata _tollCommitments,
+    bytes32[] calldata _pkxes,
+    bytes32[] calldata _ciphertexts
+  ) external payable {
+    bytes memory payload;
+
+    for (uint256 i = 0; i < _receivers.length; i++) {
+      if (_tokenAddrs[i] == address(0x0)) {
```
mds1, line 29: Common convention is to use `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` to represent ETH, which is what we do in Umbra. However with the new structs this check would be unnecessary

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
...
+        payload = abi.encodeWithSignature(
+          "sendEth(address,uint256,bytes32,bytes32)",
+          _receivers[i],
+          _tollCommitments[i],
+          _pkxes[i],
+          _ciphertexts[i]
+        );
+
+        (bool success, bytes memory data) = umbra.delegatecall(payload);
+        require(success && (data.length == 0 || abi.decode(data, (bool))), "TRANSFER_FAILED");
```
mds1, line 39: This should be a regular call using `umbra.sendEth(...)`, removing the need for the manual payload encoding. If we delegatecall, we're executing the Umbra logic in the context of this router. So storage writes and event emissions would come from the router—not from the Umbra contract—which is not what we want.

And make sure to either (1) execute approvals for supported tokens in the constructor, or (2) do it in-line here. 

Approach 1 is cheaper for users, but then we need to deploy new contracts each time a new token is supported, or add a method that lets anyone set allowance of a token to `type(uint256).max`. Need to think through if there are any security implications of the latter. I don't think so, because if there are then approach 2 will have those security concerns also

mds1 (commented):
This is the right general direction! Just left a few comments on how to clean things up, and @apbendi feel free to sanity check all this

contracts/contracts/UmbraRouter.sol
mds1: You can even just name all three methods `batchSend`, I don't have much preference on which we use

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.7.6;
```
apbendi, line 3: We should be able to configure hardhat to use multiple versions of the compiler, and bump this up to the latest version of Solidity, `^0.8.13`

garyghayrat:
@mds1 @apbendi Might worth another look now.

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,102 @@
...
+        abi.encodeWithSelector(
+          IUmbra.sendEth.selector,
+          _params[i].receiver,
+          _tollCommitment,
+          _params[i].pkx,
+          _params[i].ciphertext
+        );
+
+      (bool success, ) = umbra.call{value: _amount}(data);
+      require(success, "sendEth call failed");
```
mds1, line 61: Instead of doing this manually, we can make things cleaner by leveraging our interface. This is the reason for making the `umbra` variable of type `IUmbra` instead of `address`

```suggestion
      umbra.sendEth(_params[i].receiver, _tollCommitment, _params[i].pkx, _params[i].ciphertext);
```

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,102 @@
...
+          IUmbra.sendToken.selector,
+          _params[i].receiver,
+          _params[i].tokenAddr,
+          _amount,
+          _params[i].pkx,
+          _params[i].ciphertext
+        );
+
+      (bool success, ) = umbra.call(data);
+      require(success, "sendToken call failed");
```
mds1, line 90: Same comment about using `umbra.sendToken(...)` here instead of manually encoding and verifying the call

mds1 (commented):
Second look done!

contracts/contracts/UmbraRouter.sol
mds1: @apbendi Any preference? I'm guessing you'll prefer the current approach of separate names. If so just mark this resolved

contracts/contracts/UmbraRouter.sol
mds1: Probably can decide how we want to handle token approvals as part of a separate gas golfing PR, so marking this resolved

contracts/contracts/UmbraRouter.sol
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  function batchSendTokens(uint256 _tollCommitment, SendToken[] calldata _params) public payable {
+    for (uint256 i = 0; i < _params.length; i++) {
+      uint256 _amount = _params[i].amount;
+      address _tokenAddr = _params[i].tokenAddr;
+      IERC20 token = IERC20(address(_tokenAddr));
+
+      //User needs to approve router address as spender first
+      token.transferFrom(msg.sender, address(this), _amount);
```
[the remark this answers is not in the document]
garyghayrat: I made an attempt at this. The fn is called receiveTokens.

contracts/contracts/UmbraRouter.sol
garyghayrat: I thought you suggested using abi.encodeWithSelector. https://github.com/ScopeLift/umbra-protocol/pull/311#discussion_r837951449 What would be a situation where you might want to use selectors?

garyghayrat:
@mds1 Changing the filename messed up the diff, but I just made the calls use the interface; added receiveTokens in place of calling `token.transferFrom(msg.sender, address(this), _amount);` in a loop; added 3 global vars to be used in receiveTokens. Prob not the most efficient way but a start. Looking forward to another review while I start testing more.

contracts/contracts/UmbraRouter.sol
mds1: Yea sorry that was unclear—a few lines down when I saw what's going on I added a separate about using `umbra.sendETH` instead. In context of that specific line, `encodeWithSelector` was a better choice, but in context of the contract, `umbra.sendEth` is better.

Here's an example where a [timelock](https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/Timelock.sol#L95) contract uses `abi.encodePacked` because it doesn't know the sig/calldata until runtime. They could have also done `abi.encodeWithSignature`

contracts/contracts/UmbraRouter.sol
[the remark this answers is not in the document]
mds1: This works, but in some cases this will actually be more expensive than the previous approach! If I'm batch sending tokens with one send of DAI, one of USDC, and one of USDT, I still do three transfers, but now I have extra logic and storage writes. And storage refunds are capped at 20% so even though the cost of "[slot started zero, currently nonzero, now being reset to zero](https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a7-sstore)" is only 100 gas because of a 19,900 gas refund, you won't actually realize the full 59700 refund making this new approach a lot pricier.

That's why I think we should leave the original approach for now, and think harder about the options in a follow up PR 🙂

(As a side note, you don't need the nested mapping since the first arg is always `msg.sender` and gets cleared at the end of the transaction, so a standard mapping should work the same)

contracts/contracts/UmbraRouter.sol
mds1: (feel free to mark this one resolved whenever you see fit)

contracts/contracts/UmbraRouter.sol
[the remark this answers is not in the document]
mds1: In case you're interested, here are some of the approaches we'll consider:
- Use [memmove](https://github.com/brockelmore/memmove) for in memory arrays and mappings to avoid storage writes (not yet audited). With this approach, we'd basically do what you do now, but in memory, so it's cheaper than storage
- Require the array inputs to be sorted by address. Then we can more easily `require(params[i].tokenAddr >= params[i-1].tokenAddr)` on each loop iteration to simplify tracking the tokens and amounts
- Add a new struct array input called `TransferSummary[]` and have the user tell us what their token sends sum up to, and verify they told us the truth. This is how we did something similar in [dgrants](https://github.com/dcgtc/dgrants/blob/2732f3107fd497b8c7f6ba7a13fcfb9f9f0e89e9/contracts/contracts/GrantRoundManager.sol#L120-L124) but instead of simple token transfers we had varying amounts of tokens to swap

PR #313, opened 2022-03-30
Fix/private key access

mds1:
Changes:
1. When withdrawing, show a warning if you withdraw to the same stealth address
2. On the receive table:
    1. If advanced mode is off, no changes
    2. If advanced mode is on, funds that were previously withdrawn should show a cursor pointer over the green "Withdrawn", and when clicking it the expanded area should be shown with the only action being to show/hide the private key

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -472,7 +472,12 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
     if (!provider.value) throw new Error('Wallet not connected. Try refreshing the page and connect your wallet');
     if (!userAddress.value) throw new Error('Wallet not connected. Try refreshing the page and connect your wallet');
     activeAnnouncement.value = announcement;
-    const { safe, reasons } = await isAddressSafe(destinationAddress.value, userAddress.value, provider.value);
+    const { safe, reasons } = await isAddressSafe(
+      destinationAddress.value,
+      userAddress.value,
+      announcement.receiver,
+      provider.value
+    );
 
     if (safe) {
       showConfirmationModal.value = true;
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -119,7 +119,15 @@
               </q-card-section>
               <q-separator />
               <q-card-actions class="row justify-center items-center">
-                <div v-if="props.row.isWithdrawn" class="text-positive">
+                <div
+                  v-if="props.row.isWithdrawn"
+                  class="text-positive"
+                  :class="{ 'cursor-pointer': advancedMode }"
+                  @click="
+                    hidePrivateKey();
+                    if (advancedMode) expanded = expanded[0] === props.key ? [] : [props.key];
+                  "
+                >
                   Withdrawn<q-icon name="fas fa-check" class="q-ml-sm" />
                 </div>
                 <base-button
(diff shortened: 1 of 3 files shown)


frontend/src/components/WithdrawForm.vue
```diff
@@ -1,55 +1,57 @@
...
+        will be lost.
+        <router-link class="hyperlink" to="/faq#receiving-funds" target="_blank"> Learn more </router-link>.
+      </div>
     </div>
 
     <!-- Advanced feature: show private key -->
     <div v-if="advancedMode">
-      <div @click="emit('togglePrivateKey')" class="text-caption hyperlink q-mt-lg">
+      <div @click="emit('togglePrivateKey')" class="text-caption hyperlink">
         {{ spendingPrivateKey ? 'Hide' : 'Show' }} withdrawal private key
```
apbendi, line 55: Wondering if we should change the copy here to "Show/Hide stealth address private key" instead of "withdrawal"

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -119,7 +119,15 @@
...
+                <div
+                  v-if="props.row.isWithdrawn"
+                  class="text-positive"
+                  :class="{ 'cursor-pointer': advancedMode }"
+                  @click="
+                    hidePrivateKey();
+                    if (advancedMode) expanded = expanded[0] === props.key ? [] : [props.key];
+                  "
+                >
                   Withdrawn<q-icon name="fas fa-check" class="q-ml-sm" />
```
apbendi, line 131: Should we show some kind of caret or other indicator that this can be expanded?

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -119,7 +119,15 @@
               </q-card-section>
```
apbendi, line 119: *(arbitrary comment location)*

Looks fine, and all functionality seems to work OK. One issue I did encounter. If you:

1. Send base asset
2. Withdraw to the stealth address (ignoring the warning)

Then upon refresh, that transaction is still shown as not withdrawn. Attempting withdrawal at that point through the interface results in a JSON RPC error.

This does not happen with token withdrawals to the stealth address, which properly show as withdrawn.

Ideally, we would show the txn as withdrawn even if the user withdrew to the same stealth address.

apbendi (commented):
Looking good! A few small things and one issue.

frontend/src/components/WithdrawForm.vue
mds1: Yea I like that phrasing, updated in 5b0a150

frontend/src/components/AccountReceiveTable.vue
apbendi: > The JSON RPC error is odd though, what is the specific error? Wondering if we should also pull that into a separate issue since it's not relevant for this hotfix and likely a rare case

Screen Shot 2022-03-30 at 11 52 21 AM

This makes sense, since the stealth address already paid some gas for the "withdrawal" tx to itself, the calculation no longer works correctly.

> For ETH/native tokens we show that it's withdrawn if stealth address balance is less than the amount sent.

This seems not to be the case, or to have a bug of some kind. If it were working, the above address would have shown as withdrawn, because it had paid for gas and thus had a slightly smaller balance. Also, in one case, I actually imported the private key and withdrew ~all of the MATIC (metamask left some dust) but that case *also* shows as not-withdrawn in the interface.

apbendi: Got it, what's the current algo— checks for 0 balance?

apbendi (approved):

PR #319, opened 2022-04-02
Consume minSend amounts on frontend

davidlaprade:
Consume the minSend amounts that are sent by the backend and use them for frontend validation. Values from the backend are rounded so as to not be weirdly specific for the user.

frontend-min-send-amounts


frontend/src/pages/AccountSend.vue
```diff
@@ -292,21 +293,60 @@ function useSendForm() {
...
+      case 137: // Polygon
+        if (tokenAddress === '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619') {
+          return 0.001; // weth token minimum
+        } else {
+          return 3; // stablecoin token minimum
+        }
+      default:
+        // Mainnet, Rinkeby, and other networks have higher ETH and stablecoin minimums
+        // due to higher fees
+        return 100
```
davidlaprade, line 318: This number should probably be revised. The live values are at least 5x as high. In the event that the backend went down and we didn't have live minSends beind sent, we'd probably still want to enforce the high estimates

frontend/test/utils.test.ts
```diff
@@ -144,4 +148,22 @@ describe('Utilities', () => {
...
+  describe('humanizeMinSendAmount', () => {
+    const tests = [
+     {input: 123, output: 120},
+     {input: 1235, output: 1200},
+     {input: 1295, output: 1300},
+     {input: 432498, output: 430000},
+     {input: 4.8956, output: 5},
+     {input: 4.1956, output: 4},
+     {input: 0.1956, output: 0.2},
+     {input: 0.0194, output: 0.019},
```
davidlaprade, line 160: I'm curious if anyone disagrees with these expectations RE how we round minSend amounts for users

frontend/src/components/models.ts
```diff
@@ -97,7 +97,15 @@ export interface CnsQueryResponse {
 
 // Relayer types
 export type ApiError = { error: string };
-export type TokenListResponse = TokenList | ApiError;
+export interface TokenInfoWithMinSendAmount extends TokenInfo {
```
mds1, line 100: Is there a chance we ever add more data to these types? If so, maybe we rename to `TokenInfoExtended` and `TokenListExtended` or something similar

frontend/src/components/models.ts
```diff
@@ -97,7 +97,15 @@ export interface CnsQueryResponse {
...
+  minSendAmount: string;
+}
+// Omit the TokenList.tokens type so we can override it with our own.
+export interface TokenListWithMinSendAmount extends Omit<TokenList, 'tokens'> {
+  nativeTokenMinSendAmount: string;
```
mds1, line 105: Should the backend just return the native token info and min send amount just like any other token, instead of special casing it? Right now in `wallet.ts` we manually extend the array of tokens to add the native token, but that doesn't seem necessary/ideal

frontend/src/pages/AccountSend.vue
```diff
@@ -292,21 +293,60 @@ function useSendForm() {
...
 
   const isNativeToken = (address: string) => getAddress(address) === NATIVE_TOKEN.value.address;
-  const getMinSendAmount = (tokenAddress: string) => {
+
+  const getNativeTokenMinSendAmountFallback = (chainId: number): number => {
+    switch (chainId) {
+      case 137: return 1.0; // Polygon
+      default: return 0.01; // everything else
+    }
+  }
```
mds1, line 302: I think we should get rid of this method and just key off the native token address (`0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`) in `getTokenMinSendAmountFallback`. As you can probably tell, I'm not a fan of special-casing the native token if we don't have to since it just makes things clunkier and harder to read/maintain

frontend/src/pages/AccountSend.vue
mds1: I'll defer to you and @apbendi on good fallback values here

frontend/src/pages/AccountSend.vue
```diff
@@ -292,21 +293,60 @@ function useSendForm() {
...
+        }
+      default:
+        // Mainnet, Rinkeby, and other networks have higher ETH and stablecoin minimums
+        // due to higher fees
+        return 100
+    }
+  }
+
+  const getMinSendAmount = (tokenAddress: string): number => {
     const chainId = BigNumber.from(currentChain.value?.chainId).toNumber();
```
mds1, line 323: We use `chainId` enough that we should probably just add a `chainId` export from `wallet.ts` and start using that. We can do it as a computed property from `network.value.chainId`. I know this line was unmodified, so we can pull this out to a separate issue if you want

frontend/src/pages/AccountSend.vue
```diff
@@ -292,21 +293,60 @@ function useSendForm() {
...
+          return 3; // stablecoin token minimum
+        }
+      default:
+        // Mainnet, Rinkeby, and other networks have higher ETH and stablecoin minimums
+        // due to higher fees
+        return 100
+    }
+  }
+
+  const getMinSendAmount = (tokenAddress: string): number => {
```
mds1, line 322: Since the relayer now includes a min send amount on the returned tokens array, this method overly complicated. It seems a cleaner approach would be:
- Update `getMinSendAmount` to take the new `TokenInfoExtended` (or whatever the name is) type as input
- Refactor `wallet.ts` so the `tokens` computed property we export is of that type, and includes min send amounts
- If the relayer is down, that `tokens` export should only have ETH with the default value we define in the frontend. Otherwise, the API is working so we have all tokens and their default min send amounts, therefore we don't need to define token overrides anywhere on the frontend aside from native token overrides in `wallet.ts` (the assumption here is that we'll never get back an array of tokens without min send amounts from the API—if we currently can, then we should move the defining of token minimums to the backend so that doesn't happen, and just duplicate the native token min amount on the frontend)
- Update this method to just parse the `tokens` array for the minimum send amount and humanize it

frontend/src/utils/relayer.ts
```diff
@@ -2,20 +2,25 @@
...
   ITXStatusResponse,
   Provider,
-  TokenInfo,
+  TokenInfoWithMinSendAmount,
   RelayResponse,
   TokenListResponse,
   WithdrawalInputs,
 } from 'components/models';
 
 export class ITXRelayer {
```
mds1, line 17: Since we also use OZ now, this is no longer ITX specific so we should probably rename this class this to `UmbraRelayer` or or similar

frontend/src/utils/relayer.ts
```diff
@@ -2,20 +2,25 @@
...
-  constructor(readonly baseUrl: string, readonly tokens: TokenInfo[], readonly chainId: number) {}
+  constructor(
+    readonly baseUrl: string,
+    readonly tokens: TokenInfoWithMinSendAmount[],
+    readonly chainId: number,
+    readonly nativeTokenMinSendAmount: BigNumber | undefined,
```
mds1, line 22: Same comment about not needing to special-case the native token and just include it in the tokens array

mds1 (commented):
Also just want to note that I'd prefer if we can merge #293 before this one, just in case there's merge conflicts or anything since I think #293 is more "sensitive"

frontend/src/components/models.ts
davidlaprade: Good call, I agree

frontend/src/store/wallet.ts
```diff
@@ -78,8 +87,8 @@ export default function useWalletStore() {
               await configureProvider();
             }
           },
-          network: async (chainId) => {
-            if (network.value?.chainId && network.value.chainId !== chainId) {
+          network: async (chainIdNum) => {
+            if (chainId.value && chainId.value !== chainIdNum) {
```
mds1, line 91: `newChainId` seems like a more descriptive name

frontend/src/store/wallet.ts
```diff
@@ -362,22 +369,41 @@ export default function useWalletStore() {
...
+    if (!network.value || !chainId.value) return true; // assume valid if we have no network information
+    return supportedChainIds.includes(chainId.value);
   });
 
+  const getNativeTokenMinSend = (chainId: number): string => {
+    switch (chainId) {
+      case 137:
+        return parseUnits('0.15', 'ether').toString(); // Polygon
+      default:
+        return parseUnits('0.02', 'ether').toString(); // everything else
```
mds1, line 385: How do these compare to the prior defaults / how did we come up with these?

frontend/src/store/wallet.ts
```diff
@@ -362,22 +369,41 @@ export default function useWalletStore() {
...
+    };
   });
 
-  const tokens = computed((): TokenInfo[] => {
-    // Add ETH as a supported token
+  const tokens = computed((): TokenInfoExtended[] => {
     const supportedTokens = relayer.value?.tokens || [];
+    const isNativeTokenInRelayerTokenList = supportedTokens
+      .map((token) => token.address)
+      .includes(NATIVE_TOKEN.value.address);
```
mds1, line 403: Can we make sure the native token is always first in this list? Since I believe this order is the order used in the token dropdown. I think native token first, sort alphabetically by symbol for the rest would be great.

frontend/src/utils/umbra_api.ts
```diff
@@ -25,11 +30,19 @@ export class ITXRelayer {
...
     const data = (await response.json()) as TokenListResponse;
-    if ('error' in data) console.warn(`Could not fetch tokens from relayer: ${data.error}`);
+    let nativeMinSend: string | undefined;
+    let tokens: TokenInfoExtended[];
+    if ('error' in data) {
+      tokens = [];
+      console.warn(`Could not fetch tokens from backend: ${data.error}`);
+    } else {
+      tokens = data.tokens;
+      nativeMinSend = data.nativeTokenMinSendAmount;
```
mds1, line 40: same question about special cased native token min send amount

frontend/src/components/models.ts
davidlaprade: We have to because that's what the backend is currently returning. Until the backend is updated (which I plan to do later) this is what we need to use

frontend/src/store/wallet.ts
davidlaprade: These are the current values returned by the backend, calculated using the formula:

Gas Price x Estimated Gas Limit x Scale Factor

I'll add a note about it

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -554,14 +554,14 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
         };
 
         if (chainId === 137) {
-          // No ITX support on this network, so this is a regular transaction hash
+          // No relayer support on this network, so this is a regular transaction hash
```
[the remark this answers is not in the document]
davidlaprade: I'm not honestly sure

frontend/src/pages/AccountSend.vue
```diff
@@ -304,21 +319,16 @@ function useSendForm() {
...
-        return 3; // stablecoin token minimum
-      }
-    }
-    // Mainnet, Rinkeby, and other networks have higher ETH and stablecoin minimums due to higher fees
-    if (isNativeToken(tokenAddress)) return 0.01;
-    else return 100; // stablecoin token minimum
+
+  const getMinSendAmount = (tokenAddress: string): number => {
+    const tokenInfo = tokenList.value.filter((token) => token.address === tokenAddress)[0];
+    if (!tokenInfo) return 0;
```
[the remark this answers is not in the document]
davidlaprade: I agree. I had initially taken that approach, but then that led to noisy errors popping up when switching chains (I believe) because there was a race condition between the provider returning chain data and the backend returning token data. That might not be a problem anymore though. Let me double check...

frontend/src/pages/AccountSend.vue
```diff
@@ -248,12 +256,19 @@ function useSendForm() {
         await humanAmountInputRef.validate();
       }
 
+      const currentTokens = (tokenListValue as TokenInfoExtended[]).map((tok) => tok.address);
+      const prevTokens = (prevTokenListValue as TokenInfoExtended[]).map((tok) => tok.address);
+      if (currentTokens.toString() != prevTokens.toString()) {
+        token.value = tokenList.value[0];
+        humanAmount.value = undefined;
+      }
```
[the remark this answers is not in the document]
davidlaprade: I had needed this to avoid some noisy errors, but when I just removed it now I wasn't able to reproduce them. Not sure what happened exactly. I'll leave it out

frontend/src/components/models.ts
mds1: Are we tracking this in an issue somewhere, either here or in the backend?

frontend/src/components/AccountReceiveTable.vue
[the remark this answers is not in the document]
mds1: @apbendi Can you share the mapping between network => relayer/tx identifier used? Once we have that we should pull this out into a separate issue so we don't lose track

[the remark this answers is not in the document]
apbendi: With ITX (on Mainnet/Rinkeby), their API returns a custom hash and you have to subsequently query their endpoint to figure out the list of actual network transaction hashes. That is what we pass forward from Umbra API.

With Defender, we get back the hash of the first transaction submitted, and that is what we pass forward from Umbra API. Defender [also](https://docs.openzeppelin.com/defender/relay#transaction-ids) gives you a separate, custom Identifier which can use to query their API, to find out if that first transaction has been replaced or cancelled. We currently don't expose this in any response from Umbra API.

In practice, since we're using this on Polygon and using `fast`, this virtually never happens, and so the hash we get back ends up being the hash of the transaction that is ultimately mined. This should also be the case on Optimism and Arbitrum.

If we ever started using Defender on mainnet, we might have to revisit this assumptino.

frontend/src/pages/AccountSend.vue
apbendi: Sorry this diff is outdated now and I'm not clear what was being asked. Is my input still needed on something here?

mds1: Umm I'm not certain either haha, deferring to @davidlaprade 😅

frontend/src/components/models.ts
apbendi: Just created [this issue](https://github.com/ScopeLift/umbra-api/issues/51) on the API side to track it

davidlaprade: Looks great, thanks @apbendi

frontend/src/pages/AccountSend.vue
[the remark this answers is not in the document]
davidlaprade: It's no longer a problem. Removed 👍

frontend/src/components/AccountReceiveTable.vue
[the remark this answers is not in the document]
mds1: Created https://github.com/ScopeLift/umbra-protocol/issues/329 to track this and make sure the logic in this code block is correct

frontend/test/utils.test.ts
mds1: This seems fine to me! @apbendi any thoughts?

frontend/src/utils/relayer.ts
```diff
@@ -25,11 +30,23 @@ export class ITXRelayer {
...
     const data = (await response.json()) as TokenListResponse;
-    if ('error' in data) console.warn(`Could not fetch tokens from relayer: ${data.error}`);
+    let nativeMinSend;
+    if ('error' in data) {
+      console.warn(`Could not fetch tokens from relayer: ${data.error}`);
+    } else {
+      nativeMinSend = parseUnits(data.nativeTokenMinSendAmount, 'wei');
+    }
+
+    // TODO pre-parse token minSendAmounts?
```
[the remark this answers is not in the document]
mds1: @davidlaprade looks like this was removed, any objections to resolving this comment?

frontend/test/utils.test.ts
mds1: Actually right now on Optimism it says the minimum send is `0.0097 OETH`, but that should probably be rounded to `0.01 OETH`?

mds1: Though maybe not, If it was 0.0051 OETH we wouldn't want to round that up. In general, I guess it just feels like the variable min-send UX is kinda confusing. Gas prices might be at a low when I send funds, then they're basically stuck. It seems better to have a consistent min send based on e.g. the 90th percentile gas price, so the floor is constant and 90% of the time we know users will be able to withdraw. And then we just periodically update that.

But, this can be discussed in a separate issue

mds1: Ahh got it, sounds good. Yes I'll resolve now!

apbendi (approved):
Working well. Thanks @davidlaprade 👍

frontend/src/pages/AccountSend.vue
davidlaprade: I revised the numbers so they were close to recent minSend values from the backend. No further action needed. Thanks, guys!

mds1 (approved):

PR #326, opened 2022-04-07
added UmbraBatchSend contract and test

garyghayrat:
PR https://github.com/ScopeLift/umbra-protocol/pull/334 has the same commits as this one plus some polishing commits. Please just merge that one if you think [#344](https://github.com/ScopeLift/umbra-protocol/pull/334) PR's additional commits is ready to be merged as well.


contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
+# include .env file and export its env vars
+# (-include to ignore error if it does not exist)
+-include .env
+
+all: clean remove install update solc build 
+
+# Install proper solc version.
+solc:; nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-static-versions.solc_0_8_10
```
mds1, line 8: (sorry, should have told you there's a lot we don't need in the template I linked you to)

This is something you only need for dapptools, which uses nix, so this can be deleted

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
...
+test   :; forge test -f ${ETH_RPC_URL} --fork-block-number 10465092
+
+# Remove modules
+remove :; rm -rf .gitmodules && rm -rf .git/modules/* && rm -rf lib && touch .gitmodules && git add . && git commit -m "modules"
+
+# Install the Modules
+install :; 
+	forge install dapphub/ds-test 
+	forge install brockelmore/forge-std
+	forge install OpenZeppelin/openzeppelin-contracts
```
mds1: Personally not a fan of this approach because this will update all submodules to the latest `main`, which we don't want. I'd delete `remove` and just have `install` run `forge install`

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+
+# Update Dependencies
+update:; forge update
+
+# Builds
+build  :; forge clean && forge build --optimize --optimize-runs 1000000
```
mds1, line 30: Let's rely on the config file for build settings
```suggestion
# Builds
build  :; forge build
```

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+setup-yarn:
+	yarn 
+
+local-node: setup-yarn 
+	yarn hardhat node 
+
+deploy:
+	forge create UmbraBatchSend --private-key ${PRIVATE_KEY} # --rpc-url
```
mds1, line 39: we can get rid of all of these since we're not using them, and haven't decided how to deploy yet

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
+# include .env file and export its env vars
+# (-include to ignore error if it does not exist)
+-include .env
+
+all: clean remove install update solc build
```
mds1, line 5: After all the other changes, this can become simpler too:
```suggestion
all: clean install test
```

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
+# include .env file and export its env vars
+# (-include to ignore error if it does not exist)
+-include .env
+
+all: clean remove install update solc build 
+
+# Install proper solc version.
+solc:; nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-static-versions.solc_0_8_10
+
+# Clean the repo
```
mds1, line 10: Lots of comments that don't add much, IMO we should delete them so the makefile is a lot leaner, such as this: https://github.com/mds1/drai/blob/main/Makefile

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,87 @@
...
+    bytes32 ciphertext
+  ) external payable;
+}
+
+contract UmbraBatchSend {
+  IUmbra internal constant umbra = IUmbra(0xFb2dc580Eed955B528407b4d36FfaFe3da685401);
+
+  struct SendEth {
+    address payable receiver;
+    uint256 amount;
```
mds1, line 29: One thing we should document is that this is the total amount, i.e. amount being sent + toll. 

We should also use `forge-std` to overwrite the toll, and make sure all tests pass with both a zero toll and nonzero toll

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
...
+test   :; forge test -f ${ETH_RPC_URL} --fork-block-number 10465092
+
+# Remove modules
+remove :; rm -rf ../.gitmodules && rm -rf ../.git/modules/* && rm -rf lib"
+
+# Install the Modules
+install :; 
+	forge install dapphub/ds-test 
+	forge install brockelmore/forge-std
+	forge install OpenZeppelin/openzeppelin-contracts
```
mds1, line 23: These well delete all deps and update them to the latest main, which we don't want. We should delete `remove` and have `install` just run `forge install`

As a side note, did you install OZ by specifying a tag?

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import "ds-test/test.sol";
+import "forge-std/Vm.sol";
+import "forge-std/stdlib.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../UmbraBatchSend.sol";
+
+contract UmbraBatchSendTest is DSTest, stdCheats {
```
mds1, line 10: Let's create a `DSTestPlus.sol is DSTest, stdCheats` contract, then have all our test files inherit from that. It can live in `src/test/utils`. This way we don't need to duplicate forge-std/Vm initialization in future test contracts

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+    using stdStorage for StdStorage;
+    StdStorage stdstore;
+
+    UmbraBatchSend public router = new UmbraBatchSend();
+
+    address umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+    address Dai = 0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735;
```
mds1, line 17: nit
```suggestion
    address dai = 0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735;
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        vm.label(address(this), "TestContract");
+
+        // address umbraClone = deployCode(umbraArtifact);
+        // emit log_named_address("umbraClone address is ", umbraClone);
+
+         stdstore
+            .target(Dai)
+            .sig(IERC20(token).balanceOf.selector)
+            .with_key(address(this))
+            .checked_write(1000*1e18);
```
mds1, line 49: you can use the `tip` helper method here instead of all this

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+    address Dai = 0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735;
+
+    address alice = address(0x202204);
+    address bob = address(0x202205);
+
+    Vm vm = Vm(HEVM_ADDRESS);
+    bytes payload;
+
+    uint public tollCommitment;
+    bytes32 test = "test";
```
mds1, line 26: Let's use more meaningful variable names here for clarity. I wasn't sure what this was for until I scrolled further down. Perhaps:
```suggestion
    bytes32 pkx = "pkx";
    bytes32 ciphertext = "ciphertext";
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+    address umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+    address Dai = 0xc7AD46e0b8a400Bb3C915120d284AafbA8fc4735;
+
+    address alice = address(0x202204);
+    address bob = address(0x202205);
+
+    Vm vm = Vm(HEVM_ADDRESS);
+    bytes payload;
+
+    uint public tollCommitment;
```
mds1, line 25: These are never used, neither are a few other vars in this section

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+    UmbraBatchSend.SendToken newSendToken;
+    UmbraBatchSend.SendToken[] sendToken;
+
+    IERC20 token = IERC20(address(Dai));
+
+    function setUp() public {
+
+        vm.label(alice, "Alice");
+        vm.label(bob, "Bob");
+        vm.label(address(this), "TestContract");
```
mds1, line 40: just using the contract name here
```suggestion
        vm.label(address(this), "UmbraBatchSendTest");
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+
+        // address umbraClone = deployCode(umbraArtifact);
+        // emit log_named_address("umbraClone address is ", umbraClone);
```
mds1, line 43: delete?

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    function testBatchSendEth() public {
+
+        uint alicePrevBal = alice.balance;
+        uint bobPrevBal = bob.balance;
+
+        uint amount = 1 ether;
+        uint amount2 = 2 ether;
+        uint256 toll = 0;
```
mds1, line 59: nit, but let's be consistent and use `uint256` everywhere, since that's what the main Umbra contract does

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        uint amount = 1 ether;
+        uint amount2 = 2 ether;
+        uint256 toll = 0;        
+
+        sendEth.push(UmbraBatchSend.SendEth(payable(alice), amount, test, test));
+        sendEth.push(UmbraBatchSend.SendEth(payable(bob), amount2, test, test));
+        router.batchSendEth{value: amount + amount2}(toll, sendEth);
+
+        assertEq(alice.balance, alicePrevBal + amount);
+        assertEq(bob.balance, bobPrevBal + amount2);
```
mds1, line 66: Checking balances works, but to be more rigorous with assertions we can use `expectCall` and/or `expectEmit`

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    }
+
+    function testFuzz_BatchSendEth(uint8 amount, uint8 amount2, bytes32 testBytes) public {
```
mds1, line 70: Why such small amounts? These are unrealistically small 😛

Also we don't need the concrete tests if we also have the fuzz tests, doesn't seem to add much value

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+
+        uint alicePrevBal = alice.balance;
+        uint bobPrevBal = bob.balance;
+        vm.assume(amount > 0 && amount < type(uint8).max/2);
+        vm.assume(amount2 > 0 && amount2 < type(uint8).max/2);
```
mds1, line 75: What's this `<` check for?

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        
+
+        uint256 toll = 0;
```
mds1, line 78: We should read the toll from the contract, since we're forking

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        router.batchSendEth{value: totalAmount}(toll, sendEth);
+
+        assertEq(alice.balance, alicePrevBal + amount);
+        assertEq(bob.balance, bobPrevBal + amount2);
+
+    }
+
+    function testBatchSendTokens() public {
+
+        assertTrue(token.balanceOf(address(this)) > 0, "caller's dai balance is zero");
```
mds1, line 94: The test will fail anyway if this isn't true right, so I don't think we need this

contracts-periphery/Makefile
```diff
@@ -0,0 +1,39 @@
...
+all: clean remove install update solc build 
+
+# Install proper solc version.
+solc:; nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-static-versions.solc_0_8_10
+
+# Clean the repo
+clean  :; forge clean
+
+# Test your contract on a forked network
+test   :; forge test -f ${ETH_RPC_URL} --fork-block-number 10465092
```
mds1, line 14: We should add a similar snapshot command, then add an `UmbraBatchSendGas` contract which does all required setup in `setUp` and has one-line tests that only send various combinations of tokens and ETH. This way we can use it as a benchmark for optimizations in a future PR

```
snapshot   :; forge snapshot -f ${ETH_RPC_URL} --fork-block-number 10465092
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        token.approve(address(router), 1500);
+
+        router.batchSendTokens{value: toll}(toll, sendToken);
+        assertEq(token.balanceOf(umbra), umbraPrevBal + 1500);
+
+    }
+
+    function testFuzz_BatchSendTokens(uint8 amount, uint8 amount2) public {
+
+        assertTrue(token.balanceOf(address(this)) > 0, "caller's dai balance is zero");
```
mds1, line 110: similar comment, this line isn't necessary

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+
+        uint256 toll = 0;
+        uint umbraPrevBal = token.balanceOf(umbra);
+
+        sendToken.push(UmbraBatchSend.SendToken(alice, Dai, 1000, test, test));
+        sendToken.push(UmbraBatchSend.SendToken(bob, Dai, 500, test, test));
+        token.approve(address(router), 1500);
+
+        router.batchSendTokens{value: toll}(toll, sendToken);
+        assertEq(token.balanceOf(umbra), umbraPrevBal + 1500);
```
mds1, line 104: similar comment around stricter assertions

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    function testBatchSend() public {
+
+        uint alicePrevBal = alice.balance;
+        uint bobPrevBal = bob.balance;
+
+        uint amount = 1 ether;
+        uint amount2 = 2 ether;
+        uint total = amount + amount2;
+        uint256 toll = 0;
```
mds1, line 135: the tests have a lot of repetitive variable declarations that can be pulled into the `setUp` method to deduplicate

mds1 (commented):
Sorry, lots of the comments are because I sent you a makefile template that was overkill for our use case here!

contracts-periphery/src/UmbraBatchSend.sol
mds1: A good pattern for this is define all the tests in an `abstract` contract, then have two contracts that inherit it: one that sets a zero toll in `setUp`, the other that sets a nonzero toll in `setUp`. Both contracts will then run all the tests that the abstract contract has

contracts-periphery/src/test/UmbraBatchSend.t.sol
mds1: Make sure to move the `Vm vm = Vm(HEVM_ADDRESS);` definition out of `UmbraBatchSendTest` and into `DSTestPlus` 🙂

contracts-periphery/src/test/UmbraBatchSend.t.sol
mds1: On a related note, let's change the these contracts to use 2 spaces instead of 4, again for consistency with `Umbra.sol`

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,93 @@
...
+      valAccumulator += _amount;
+
+      umbra.sendEth{value: _amount}(_params[i].receiver, _tollCommitment, _params[i].pkx, _params[i].ciphertext);
+    }
+
+    if(msg.value != valAccumulator) revert ValueMismatch();
+    emit Log(msg.sender, msg.value, "called batchSendEth");
+  }
+
+  function batchSendTokens(uint256 _tollCommitment, SendToken[] calldata _params) public payable {
```
mds1, line 60: This method needs a check that `_tollCommitment * _params.length == msg.value` to prevent excess ETH from getting stuck

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,93 @@
...
+    emit Log(msg.sender, msg.value, "called batchSendTokens");
+  }
+
+  function batchSend(
+    uint256 _tollCommitment,
+    SendEth[] calldata _ethParams,
+    SendToken[] calldata _tokenParams
+  ) external payable {
+    batchSendEth(_tollCommitment, _ethParams);
+    batchSendTokens(_tollCommitment, _tokenParams);
```
mds1, line 90: Once you add the above `_tollCommitment * _params.length == msg.value` check, this method probably will not work due to conflicting `msg.value checks. Instead we'll need to refactor into the following interface:
- `_batchSendEth` is an internal method with no safety checks and executes all logic
- `batchSendETH` is an external method that executes safety checks the calls `_batchSendEth` 
- same pattern as above for `batchSendTokens`
- `batchSend` is an `external` method with its own safety checks, which then calls the two internal methods

This may result in us having to loop through arrays twice—once for safety checks, then a second time for execution—but I think that's ok for now and we can optimize in a future PR

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,93 @@
...
+    for (uint256 i = 0; i < _params.length; i++) {
+      uint256 _amount = _params[i].amount;
+      address _tokenAddr = _params[i].tokenAddr;
+      IERC20 token = IERC20(address(_tokenAddr));
+
+      token.transferFrom(msg.sender, address(this), _amount);
+
+      if (token.allowance(address(this), address(umbra)) == 0) {
+        token.approve(address(umbra), type(uint256).max);
+      }
```
mds1, line 71: We should use OZ's `safeTransferFrom` and `safeApprove` here

contracts-periphery/src/test/BatchSendwToll.t.sol
```diff
@@ -0,0 +1,41 @@
...
+
+contract BatchSendwTollTest is UmbraBatchSendTest {
+
+    address mockOwner;
+
+    function setUp() public override {
+
+        vm.label(alice, "Alice");
+        vm.label(bob, "Bob");
+        vm.label(address(this), "UmbraBatchSendTest");
```
mds1, line 14: Did you find that the test contract was not labeled automatically in traces? It should be, and if so we can remove this

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,121 @@
...
+
+    IERC20 token = IERC20(address(dai));
+    UmbraBatchSend.SendEth[] sendEth;
+    UmbraBatchSend.SendToken[] sendToken;
+
+    function setUp() virtual public {
+
+        vm.label(alice, "Alice");
+        vm.label(bob, "Bob");
+        vm.label(address(this), "UmbraBatchSendTest");
```
mds1, line 43: same comment, pretty sure this label is not needed

contracts-periphery/src/test/BatchSendwToll.t.sol
```diff
@@ -0,0 +1,41 @@
...
+        // mockOwner = UmbraToll(umbra).owner();
+        // emit log_named_address("Umbra owner addr is", mockOwner);
+        // vm.deal(mockOwner, type(uint256).max);
+        // vm.startPrank(mockOwner);
+        // UmbraToll(umbra).setToll(10);
+        // vm.stopPrank();
+        toll = 0;
+
+        tip(dai, address(this), type(uint256).max);
+        vm.deal(address(this), type(uint256).max);
```
mds1, line 27: If we're inheriting from the other test contract, we can call `super.setUp()` instead of duplicating the logic.

Also, I'd recommend setting balances significantly smaller than max uint256, like `100 ether` is probably plenty. Max uint is not necessary/realistic, and likely to result in overflow if you ever try transferring funds to that address

contracts-periphery/src/UmbraBatchSend.sol
mds1: Looks like this was added but no test was added for it. In general there should be at least 1 test for each condition checked in a function that could cause a revert (currently I don't see any `expectRevert` tests)

contracts-periphery/src/UmbraBatchSend.sol
mds1: Looks like this was done, also still needs some tests on the safety checks. Also looks like `batchSendEth` and `batchSendToken` are `public`, but they can now be `external`

contracts-periphery/.env.exmaple
```diff
@@ -0,0 +1,2 @@
+ETH_RPC_URL=
```
apbendi, line 1: We could use the addition of README to this subdirectory containing:

- [ ] A quick description of what the subdir is for
- [ ] Instructions to getting setup, including copying the .env.example and what the vars are supposed to be/used for

apbendi: Also is this supposed to be a Rinkeby RPC or mainnet RPC? Or either?

contracts-periphery/.env.exmaple
```diff
@@ -0,0 +1,2 @@
...
+PRIVATE_KEY=
```
apbendi, line 2: I don't see this used anywhere, what's it for?

contracts-periphery/Makefile
```diff
@@ -0,0 +1,14 @@
...
+-include .env
+forkBlock = 10465092
+
+all:; clean install test
+# Clean the repo
+clean  :; forge clean
+# Install the Modules
+install :; forge install
+# Test your contract on a forked network
+test   :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock}
```
apbendi, line 12: Running `make test` I currently get:

```
The application panicked (crashed).
Message:  couldn't deploy:
   0: deployment failed: Revert

Location:
   /Users/runner/work/foundry/foundry/evm/src/executor/mod.rs:448

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Location: /Users/runner/work/foundry/foundry/forge/src/runner.rs:226

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
make: *** [test] Abort trap: 6
```

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+    address receiver,
+    address tokenAddr,
+    uint256 amount,
+    bytes32 pkx,
+    bytes32 ciphertext
+  ) external payable;
+}
+
+contract UmbraBatchSend {
+  IUmbra internal constant umbra = IUmbra(0xFb2dc580Eed955B528407b4d36FfaFe3da685401);
```
apbendi, line 25: Even though we're aiming for the same address on all networks, I think I'd still lean toward an `immutable` that is set in the constructor for more flexibility in future unexpected contexts.

contracts-periphery/src/UmbraBatchSend.sol
apbendi: > One thing we should document is that this is the total amount, i.e. amount being sent + toll.

As I read the code below, this is *not* the total amount. Am I wrong?

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+  event Log(address indexed caller, uint256 indexed value, string message);
+
+  function batchSendEth(uint256 _tollCommitment, SendEth[] calldata _params) external payable {
+    uint256 valAccumulator;
+
+    for (uint256 i = 0; i < _params.length; i++) {
+      //amount to be sent per receiver
+      uint256 _amount = _params[i].amount;
+      valAccumulator += _amount;
+      valAccumulator += _tollCommitment;
```
apbendi, line 52: We can reduce this to one line `valAccumulator = valAccumulator + _amoung + _tollCmmitment`. Might also save gas depending on how smart the optimizer is.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+    uint256 amount;
+    bytes32 pkx;
+    bytes32 ciphertext;
+  }
+  
+  error ValueMismatch();
+  event Log(address indexed caller, uint256 indexed value, string message);
+
+  function batchSendEth(uint256 _tollCommitment, SendEth[] calldata _params) external payable {
+    uint256 valAccumulator;
```
apbendi, line 46: nit: would prefer a more descriptive name like `totalValueSending` or something like this

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    for (uint256 i = 0; i < _params.length; i++) {
+      //amount to be sent per receiver
+      uint256 _amount = _params[i].amount;
```
apbendi, line 50: Why the temporary variable here?

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+  struct SendToken {
+    address receiver;
+    address tokenAddr;
+    uint256 amount;
+    bytes32 pkx;
+    bytes32 ciphertext;
+  }
+  
+  error ValueMismatch();
+  event Log(address indexed caller, uint256 indexed value, string message);
```
apbendi, line 43: A few things:

* Would like to think of a better name for this event, maybe `BatchedSend`
* Nit: `caller` -> `sender` is more descriptive
* I think we need to include the token address in this event
* Nit: `value` -> `amount` is more generic
* What's the string `message` about?

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    if(msg.value != valAccumulator + _tollCommitment * _tokenParams.length) revert ValueMismatch();
+
+    _batchSendEth(_tollCommitment, _ethParams);
+    _batchSendTokens(_tollCommitment, _tokenParams);
+    emit Log(msg.sender, msg.value, "called batchSend");
+  }
+
+  function _batchSendEth(uint256 _tollCommitment, SendEth[] calldata _params) internal {
```
apbendi, line 85: Nit: suggest renaming `_tollCommitment` to `_totalTolls` or something similar to distinguish from a single toll.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+    uint256 valAccumulator;
+
+    for (uint256 i = 0; i < _ethParams.length; i++) {
+      //amount to be sent per receiver
+      uint256 _amount = _ethParams[i].amount;
+      valAccumulator += _amount;
+      valAccumulator += _tollCommitment;
+    }
+
+    if(msg.value != valAccumulator + _tollCommitment * _tokenParams.length) revert ValueMismatch();
```
apbendi, line 78: Style nit: put parens around the math on the RHS, or even better pull it out into a well named temp variable.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  function batchSend(
+    uint256 _tollCommitment,
+    SendEth[] calldata _ethParams,
+    SendToken[] calldata _tokenParams
+  ) external payable {
+    uint256 valAccumulator;
+
+    for (uint256 i = 0; i < _ethParams.length; i++) {
```
apbendi, line 71: Logic here is repeated above. Might be worth considering extracting into an internal pure method that takes the params and tollCommitment and returns total value.

apbendi (changes requested):
Awesome first pass at this @garyghayrat! Looking forward to seeing this come together.

contracts-periphery/Makefile
garyghayrat: Not sure why this is, it's working for me. Have you tried `foundryup`?

contracts-periphery/src/UmbraBatchSend.sol
mds1: Right, sorry! Either way should make sure it's clear that total amount paid by user is `sum(_params.amounts) + tollCommitment * _params.length` (and similar docs for each method)

contracts-periphery/src/test/UmbraBatchSend.t.sol
mds1: From a quick check, I think only `DSTestPlus` is still using 4 spaces and needs an update

contracts-periphery/.env.exmaple
mds1: It's supposed to be rinkeby since the token address hardcoded in the tests in rinkeby DAI, but agreed docs would be great.

But now that I think of it, we should just write a mock ERC-20 token that way tests will work regardless of which network you fork from. Even more generally, we can just use `deployCode` to deploy Umbra locally and avoid the need to fork at all

contracts-periphery/.env.exmaple
mds1: I think it's remnants from the Makefile we used as a template, which had a private key input for deployment. If this isn't used anywhere right now I agree we should remove it

contracts-periphery/Makefile
mds1: Foundry currently has poor support for showing useful error messages when your test contract deployments fail in the constructor. It panics instead of showing a trace, and that's what's happening here.

The most likely reason is you're forking from the wrong network. Even though there's no explicit constructor, the test contract calls `token.balanceOf(umbra)` in the top storage portion, and that's executed during construction. Since you are on the wrong network, that token will not exist and that call will revert, causing deployment to fail. Two things we should do:
1. Document what to fork against, OR remove the need for forking, as we are discussing in a comment above
2. The test contract should not make any calls that set values in a constructor: variables should all be set in the `setUp` function to prevent opaque errors like this

contracts-periphery/src/UmbraBatchSend.sol
mds1: What about `valueSentAccumulator`? I do like `accumulator` as it's a common descriptor for variables like this

contracts-periphery/src/UmbraBatchSend.sol
mds1: Yea this isn't necessary since we only access the amount once

contracts-periphery/src/UmbraBatchSend.sol
mds1: I was on the fence at whether or not we should even have a log here at all, but I think it will be useful for UI/UX helpers. I think the use case is to find batch sends from a user. Given that, this doesn't need many details except for the sender:
- Have a single event: `BatchSendExecuted(address indexed caller)`
- Then we search for all those events by the user
- That gives a tx hash/block, then we query the node for `Announcement` events from the same sender in that block

Having this be more complex means you need to emit it each iteration of the loop, which seems like a waste of gas since the `Announcement` contains all the same info. It does reduce an RPC call on the frontend, but that's not worth the gas IMO, especially cause we can write a subgraph to fetch batch-sends and just make one request

contracts-periphery/.env.exmaple
garyghayrat: I liked the no-fork-needed idea, but was having trouble deploying the main Umbra contract. testDeployMainContract() returns deployed address is 0x000...00. https://gist.github.com/garyghayrat/8a2bc2e7d03d3b9493a0b3a520c62ca1.js

contracts-periphery/src/UmbraBatchSend.sol
mds1: This is the value of the toll you are committing to for each send, not the total toll cost you pay, so I think that rename would be confusing

contracts-periphery/src/UmbraBatchSend.sol
apbendi: >  I think the use case is to find batch sends from a user

Well, more specifically, we show the from address on the Receive page. Are you accounting for how we'll do that? Since the msg.sender from the `Umbra` contract's perspective will be the batch send contract, we'll need a way to find it each time.

Given that, I'm actually wondering if we should be emitting all the params of the Umbra announcement each time? Then on transaction scanning we could:

* Fetch all normal `Umbra` Announcement events
* Throw out the ones where sender is the Batch Send contract
* Fetch all Batch Send Announcements
* Combine the remaining Umbra Announcements w/ the Batch Send Announcements

If we don't do it this way, we'll need to do something like what you described above each time to reconcile the sender. It could get messy. Agree the gas overhead is not ideal though.

> That gives a tx hash/block, then we query the node for Announcement events from the same sender in that block

But you could have more than one per block (multiple txs) or even more than one per txn (if a contract batches our Batch Send method 😅 ), right?

contracts-periphery/src/UmbraBatchSend.sol
apbendi: You're right, I misread this.

contracts-periphery/Makefile
apbendi: > The most likely reason is you're forking from the wrong network. 

My RPC url is for Rinkeby so I don't think that's it

contracts-periphery/src/UmbraBatchSend.sol
mds1: > Well, more specifically, we show the from address on the Receive page. Are you accounting for how we'll do that? Since the msg.sender from the `Umbra` contract's perspective will be the batch send contract, we'll need a way to find it each time.

`Annoucement` events do not emit a sender parameter, so right now we use the transaction's `from` address as the sender (i.e. `tx.origin`). So that approach wouldn't change here. (Obviously this doesn't work if sends are from a multsig/contract wallet, but that's a separate issue)

So I think my proposal is the simplest/cheapest and accomplishes what we need?

contracts-periphery/Makefile
mds1: RIP my whole theory haha. We're removing the need to fork anyway so probably not worth debugging

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  function batchSend(
+    uint256 _tollCommitment,
+    SendEth[] calldata _ethParams,
+    SendToken[] calldata _tokenParams
+  ) external payable {
+    uint256 valueSentAccumulator = _valueSentAccumulator(_tollCommitment, _ethParams);
+    //ethParams sum(amount + toll) + tokenParams' total toll
+    uint totalAmount = valueSentAccumulator + (_tollCommitment * _tokenParams.length);
```
mds1, line 70: Nit on being consistent with `uint256` instead of `uint`. Also this comment is a bit unclear to me—is it just saying the same thing as L70?

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+}
+
+contract BatchSendWithTollTest is UmbraBatchSendTest {
+  using stdStorage for StdStorage;
+
+  function setUp() public override {
+    super.setUp();
+    stdstore.target(address(umbra)).sig("toll()").checked_write(0.1 ether);
```
mds1, line 145: For readability, I'd add a `setToll(uint256)` method in DSTestPlus that does this. Will likely be useful for future contracts too

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+pragma solidity ^0.8.13;
+
+import "./utils/DSTestPlus.sol";
+import "../UmbraBatchSend.sol";
+
+interface UmbraToll {
+  function toll() external returns(uint256);
+}
+
+contract UmbraBatchSendTest is DSTestPlus {
```
mds1, line 11: Let's declare this as abstract. Currently tests run for this contract, but we don't want that—we only want the child contracts' tests to run
```suggestion
abstract contract UmbraBatchSendTest is DSTestPlus {
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+  bytes32 ciphertext = "ciphertext";   
+
+  UmbraBatchSend.SendEth[] sendEth;
+  UmbraBatchSend.SendToken[] sendToken;
+
+  error ValueMismatch();
+
+  function setUp() virtual public {
+    umbra = deployCode("src/test/utils/Umbra.json", bytes(abi.encode(0, address(this), address(this))));
+    vm.etch(0xFb2dc580Eed955B528407b4d36FfaFe3da685401, umbra.code);
```
mds1, line 38: I like what you're doing here! Just would suggest adding this comment to clarify the intent (assuming I understood correctly)
```suggestion
    // Deploy Umbra at an arbitrary address, then place the resulting bytecode at the same address as the production deploys.
    umbra = deployCode("src/test/utils/Umbra.json", bytes(abi.encode(0, address(this), address(this))));
    vm.etch(0xFb2dc580Eed955B528407b4d36FfaFe3da685401, umbra.code);
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+    umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
```
mds1, line 39: This can be a constant that we declare on L15, then we can use that value in the `etch` call on L38

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+    router = new UmbraBatchSend(address(umbra));
+    token = new MockERC20("Test","TT", 18);
+    tokenAddr = address(token);
+    toll = UmbraToll(umbra).toll();
```
mds1, line 43: Let's delete this line and only explicitly set `toll` in the child contracts so the intent is more clear

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+
+  function testFuzz_BatchSendEth(uint64 amount, uint64 amount2) public {
+    vm.assume(amount > 0);
+    vm.assume(amount2 > 0);
+    vm.assume(amount > toll && amount2 > toll);
+
+    sendEth.push(UmbraBatchSend.SendEth(payable(alice), amount, pkx, ciphertext));
+    sendEth.push(UmbraBatchSend.SendEth(payable(bob), amount2, pkx, ciphertext));
+
+    uint256 totalAmount = uint256(amount) + uint256(amount2) + (toll * sendEth.length);
```
mds1, line 56: Nit, but I think you only need to upcast one of these, e.g. 
```suggestion
    uint256 totalAmount = uint256(amount) + amount2 + (toll * sendEth.length);
```

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+}
+
+contract BatchSendWithTollTest is UmbraBatchSendTest {
+  using stdStorage for StdStorage;
+
+  function setUp() public override {
+    super.setUp();
+    stdstore.target(address(umbra)).sig("toll()").checked_write(0.1 ether);
+    toll = UmbraToll(umbra).toll();
+    assertEq(toll, 0.1 ether);
```
mds1, line 147: Let's avoid assertions here and either add a `testPostSetupState()` test to the abstract contract, or use `require` here—recall assertions in `setUp` may give some unexpected behavior: https://github.com/foundry-rs/foundry/issues/1291

mds1, line 147: Let's avoid assertions here and either add a `testPostSetupState()` test to the abstract contract, or use `require` here—recall assertions in `setUp` may give some unexpected behavior: https://github.com/foundry-rs/foundry/issues/1291

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,160 @@
...
+}
+
+contract BatchSendWithoutTollTest is UmbraBatchSendTest {
+  using stdStorage for StdStorage;
+
+  function setUp() public override {
+    super.setUp();
+    stdstore.target(address(umbra)).sig("toll()").checked_write(uint256(0));
+    toll = UmbraToll(umbra).toll();
+    assertEq(toll, 0);
```
mds1, line 158: Same comment about assertions

mds1:
Really looking good now, almost finished! 🙌

contracts-periphery/.env.exmaple
apbendi: hey @garyghayrat I think we could improve our description a bit, specifically something like

```markdown
# Umbra Periphery Contracts

Smart contracts that integrate with [Umbra](../README.md), but are not part of the [core protocol contracts](../contracts-core/).
```

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,112 @@
...
+    address tokenAddr;
+    uint256 amount;
+    bytes32 pkx;
+    bytes32 ciphertext;
+  }
+  
+  error ValueMismatch();
+  event BatchSendExecuted(address indexed sender);
+
+  constructor(address umbraAddr) {
```
apbendi, line 45: nit: Can declare the param as `IUmbra` instead of casting below.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,112 @@
...
+    umbra = IUmbra(umbraAddr);
+  }
+
+  function batchSendEth(uint256 _tollCommitment, SendEth[] calldata _params) external payable {
+    uint256 valueSentAccumulator = _valueSentAccumulator(_tollCommitment, _params);
```
apbendi, line 50: The temporary variable here is not adding any value. It doesn't make the code clearer or terse. Let's just use the call to the method directly in the `if` condition below.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,112 @@
...
+    _batchSendTokens(_tollCommitment, _params);
+    emit BatchSendExecuted(msg.sender);
+  }
+
+  function batchSend(
+    uint256 _tollCommitment,
+    SendEth[] calldata _ethParams,
+    SendToken[] calldata _tokenParams
+  ) external payable {
+    uint256 valueSentAccumulator = _valueSentAccumulator(_tollCommitment, _ethParams);
```
apbendi, line 68: Same comment as above— the temporary variable adds no value. Let's move the function call directly onto line 70 where `totalAmount` is calculated.

contracts-periphery/src/UmbraBatchSend.sol
apbendi, line 70: `totalAmount` should be have `_` prefix

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,112 @@
...
+
+  function _batchSendEth(uint256 _tollCommitment, SendEth[] calldata _params) internal {
+    for (uint256 i = 0; i < _params.length; i++) {
+      umbra.sendEth{value: _params[i].amount + _tollCommitment}(_params[i].receiver, _tollCommitment, _params[i].pkx, _params[i].ciphertext);
+    }
+  }
+
+  function _batchSendTokens(uint256 _tollCommitment, SendToken[] calldata _params) internal {
+    for (uint256 i = 0; i < _params.length; i++) {
+      address _tokenAddr = _params[i].tokenAddr;
```
apbendi, line 86: Temporary variable here doesn't add much value, just use `_param[i].tokenAddr` in the cast.

contracts-periphery/src/UmbraBatchSend.sol
apbendi: Ahh ok, you're right! I think the only edge case to consider then is a batch of batch sends. Any integration logic assuming one `BatchSendExecuted` per txn could break on that, right? But yeah shouldn't be a big issue I guess.

mds1: That would be a weird edge case, i.e. why would someone do that, but with our logic of "tx.origin = sender" I think everything should still work out fine in that situation?

contracts-periphery/src/UmbraBatchSend.sol
mds1: Still two `uint`s in `UmbraBatchSend.t.sol` 🙂

contracts-periphery/src/test/UmbraBatchSend.t.sol
mds1: Looks like now we declare `address umbra` on L13 and `address constant umbraAddr` on L20, and use both in the tests. We only need one of these—I'd stick with with `address constant umbra = 0x123` and delete `umbraAddr`

contracts-periphery/src/test/UmbraBatchSend.t.sol
mds1: Looks good, but we can move the `testPostSetupState` into the abstract contract to avoid having to copy it twice in the two child contracts

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: I'm so sorry😭... you shall never see `uint`s again

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,157 @@
...
+  function toll() external returns(uint256);
+}
+
+abstract contract UmbraBatchSendTest is DSTestPlus {
+
+  UmbraBatchSend router;
+  MockERC20 token;
+
+  address constant umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+  address tokenAddr;
```
mds1, line 17: nit, but let's move `tokenAddr` right below `token`, since they are the same value but different types. Alternatively you can delete one and just cast to the other type as needed

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,157 @@
...
+  address alice = address(0x202204);
+  address bob = address(0x202205);
+
+  uint256 alicePrevBal;
+  uint256 bobPrevBal;
+  uint256 umbraPrevBal;
```
mds1, line 23: These are never initialized and therefore are always zero—do we need these variables or should they be removed?

contracts-periphery/src/test/UmbraBatchSend.t.sol
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  function testPostSetupState() public {
+    uint256 currentToll = UmbraToll(umbra).toll();
+    assertEq(toll, currentToll);
+  }
+
+  function testFuzz_BatchSendEth(uint64 amount, uint64 amount2) public {
+    vm.assume(amount > 0);
+    vm.assume(amount2 > 0);
```
mds1, line 51: I don't think we need the `> 0` assumptions, do we?
```suggestion
```

apbendi (approved):
Awesome work @garyghayrat! Thanks for bearing with us on the review process. Excited to have this!

contracts-periphery/src/UmbraBatchSend.sol
mds1: @garyghayrat just bumping this docs request

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -0,0 +1,108 @@
...
+    uint256 amount;
+    bytes32 pkx;
+    bytes32 ciphertext;
+  }
+  
+  error ValueMismatch();
+  event BatchSendExecuted(address indexed sender);
+
+  constructor(IUmbra umbraInterface) {
+    umbra = umbraInterface;
```
mds1, line 46: Here we're passing in the address of the Umbra contract, so convention is that the var name is the contract name. 
```suggestion
  constructor(IUmbra _umbra) {
    umbra = _umbra;
```

mds1 (approved):

PR #334, opened 2022-04-28
Contracts batch send polish Closes #333

### LINKED ISSUE #333: chore: umbra-periphery polish
- Bump to latest forge-std version. This will require small refactor in `DSTestPlus`:
  - Now we `import "forge-std/Test.sol";` and `contract DSTestPlus is Test`.
  - Libraries are not inherited so still need `using stdStorage for StdStorage`. 
- Most of the current README is how to use a makefile. Let's trim that down, and add a section that describes all main contracts in this folder (right now just the batch send contract, and more will be added in the future).
- Setup CI. Suggest config file to use is below, but not sure how this is affected by the project being a subdirectory instead of at the repo root. I think you might just need a `cd` step before running the forge commands.
- Move `test/` to the root, instead of having it nested within `src/`.
- Remove the unnecessary items in the current foundry config.
- Add a CI profile that has 10k fuzz runs.
- Nit: Some spacing cleanup.
    - Functions and contracts should start on the line below their declaration. We do this in most places, but some have a blank line in between.
    - A few lines have trailing spaces we can delete.
    - Makefile spacing of commands is inconsistent. I'd suggest aligning them vertically at the `:;` or having just one space on either side of each `:;`.

## CI Setup

Suggested contents of `.github/workflows/ci.yaml` are below:

```yaml
name: CI
on:
  push:
    branches:
      - main
  pull_request:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: recursive

      - name: Install Foundry
        uses: onbjerg/foundry-toolchain@v1
        with:
          version: nightly

      - name: Build contracts
        run: |
          forge --version
          forge build --sizes
  tests:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: recursive

      - name: Install Foundry
        uses: onbjerg/foundry-toolchain@v1
        with:
          version: nightly

      - name: Run tests
        run: FOUNDRY_PROFILE=ci forge test
```

garyghayrat:
- Updated to the latest forge-std version
- Updated README added a section that describes the contract in this folder
- Setup CI. Check `Actions` tab in this branch
- Moved `test/` to the root
- Added a foundry profile that does 10k fuzz runs
- Cleaned up makefile


mds1:
@garyghayrat should this be taken out of draft?

garyghayrat:
> @garyghayrat should this be taken out of draft?

Yes!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
+name: CI
```
mds1, line 1: _(arbitrary comment location to start a thread)_

In DSTestPlus, we should be able to trim it down to this (fewer imports, one less contract we inherit from)

```solidity
import "forge-std/Test.sol";
import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
import "solmate/test/utils/mocks/MockERC20.sol";

contract DSTestPlus is Test {
  using stdStorage for StdStorage;

  function setToll(address umbraAddr, uint256 newToll) public {
    stdstore.target(address(umbraAddr)).sig("toll()").checked_write(newToll);
  }
}
```

contracts-periphery/foundry.toml
```diff
@@ -4,4 +4,7 @@ out = 'out'
 libs = ['lib']
```
mds1, line 4: Pretty sure we don't need the `src`, `out`, or `libs` entries here

contracts-periphery/foundry.toml
```diff
@@ -4,4 +4,7 @@ out = 'out'
...
 remappings = ['ds-test/=lib/ds-test/src/', '@openzeppelin/=lib/openzeppelin-contracts']
```
mds1, line 5: Do we need this ds-test mapping?

Also thoughts on removing the openzeppelin one too and just using the actual path instead of the `@` remapping? I don't think the custom `@` syntax buys us anything, and is kind of just an artifact of npm/hardhat resolution

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,44 @@
...
+on:
```
mds1, line 2: _(arbitrary comment location to start a thread)_

Should we delete the Makefile? Originally it was helpful because we tested against a forked network. Now it does nothing except alias the forge commands so I'm thinking we should remove it

contracts-periphery/foundry.toml
```diff
@@ -4,4 +4,7 @@ out = 'out'
...
 
+[ci]
+fuzz_runs = 10000
+fuzz_max_global_rejects=150000
```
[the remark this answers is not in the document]
garyghayrat: I was getting too many global rejects error when testing under the 10k run profile. I guess I could use smaller uints instead?

contracts-periphery/foundry.toml
garyghayrat: Looks clean without them. I think it's there just so if you want to deploy using hardhat, you don't have to change the imports.

garyghayrat: Might have to keep the ds-test one tho because dependencies like forge-std imports uses it even though the build didn't fail..

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
 
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "../lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
```
apbendi, line 4: Why did we change the way we're doing the imports here? Just curious.

apbendi (approved):
One small question from me, but otherwise LGTM!

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: That's because @ syntax was more for working with hardhat easier. And Matt suggested we don't really need it. https://github.com/ScopeLift/umbra-protocol/pull/334#discussion_r869808108

contracts-periphery/foundry.toml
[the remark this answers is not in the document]
mds1: Yea, usually if you hit this and feel like you need to increase the reject limit, it's better to find an alternative like smaller uint types or `bound`. Too many rejections from `assume` can slow down tests a lot

contracts-periphery/foundry.toml
mds1: We don't need to keep the ds-test one either. I tried deleting it and everything still worked.

However, I personally like to add 
```toml
[default]
remappings = ["test/=test/"] # Allows imports using `test/` to refer to root test directory.
```

this way we can import all paths with absolute paths, e.g. `import src/File.sol` and `import test/File.t.sol` instead of using relative paths

garyghayrat: Sounds good, fixed it here https://github.com/ScopeLift/umbra-protocol/pull/334/commits/191847973022f7850f017907571562d023f4918b

mds1: I also just pushed 2f5c656 to update our dep versions and 3b85fb1 to leverage your commit to use absolute paths instead of relative ones

contracts-periphery/src/UmbraBatchSend.sol
mds1: Also relevant: https://github.com/ScopeLift/umbra-protocol/pull/334#discussion_r876403812

PR #336, opened 2022-05-10
Contracts BatchSend gas improvements

garyghayrat:
Trying out different approaches listed here https://github.com/ScopeLift/umbra-protocol/pull/311#discussion_r840115363

The goal is to replace the batchSendTokens function with something more gas efficient.
I attempted:
1. `batchSendTokensOption1` function with `receiveTokens` as the internal function.
2. `batchSendTokensOption2` function with `_sumBatchSendTokens`, `_batchTransferTokens`, and `_batchSendTokensToDestination` as the internal functions.
Both of which are still costing more gas than the original `batchSendTokens` function.


contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -39,16 +52,120 @@ contract UmbraBatchSend {
...
+      totalPerToken[tokenAddr] += _params[i].amount;
+    }
+
+    for (uint256 i = 0; i < tokenAddrs.length; i++) {
+      if (tokenAddrs[i] != address(0)) {
+        IERC20 token = IERC20(tokenAddrs[i]);
+        //User needs to approve router address as spender first
+        token.transferFrom(msg.sender, address(this), totalPerToken[tokenAddrs[i]]);
+
+        delete tokenInArray[tokenAddrs[i]];
```
garyghayrat, line 85: Not deleting these key values actually cause more gas?

PR #341, opened 2022-05-25
WithdrawHook contract

garyghayrat:
To test:
- Add `ETH_RPC_URL` in `.env`
- Run `make test-hook`  or `make test-hook-50` to test with 50 different recipient and `feeReceiver` addresses.


.github/workflows/ci.yaml
```diff
@@ -41,4 +41,4 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: FOUNDRY_PROFILE=ci forge test
\ No newline at end of file
+        run: FOUNDRY_PROFILE=ci forge test --match-path test/UmbraBatchSend.t.sol
```
mds1, line 44: Why do we limit the tests being run now?

contracts-periphery/.env.example
```diff
@@ -0,0 +1 @@
+ETH_RPC_URL=
```
mds1, line 1: Guessing we're forking mainnet, so let's make this self-documenting
```suggestion
ETH_RPC_URL=yourMainnetRpcUrlHere
```

contracts-periphery/Makefile
```diff
@@ -0,0 +1,8 @@
+-include .env
+
+forkBlock = 14800081
+
+build :; FOUNDRY_PROFILE=hook forge build
+install :; forge install
+Test :; FOUNDRY_PROFILE=hook forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-contract WithdrawData -vv
+Testv :; Test :; FOUNDRY_PROFILE=hook forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-contract WithdrawData -vvvv
```
mds1, line 8: I don't think this is the makefile setup we want. With this I'd run `make Test` to run hook tests and presumably `forge test` to run all tests. But since these tests require forking, `forge test` will fail. We should add `test-{description}` lines for each type we want

.gitmodules
```diff
@@ -10,3 +10,12 @@
 [submodule "contracts-periphery/lib/solmate"]
 	path = contracts-periphery/lib/solmate
 	url = https://github.com/Rari-capital/solmate
+[submodule "contracts-periphery/lib/v3-periphery"]
```
mds1, line 13: Why do we have these uniswap dependencies / remappings? Those contracts are written in 0.7.6 so we can't use the contracts directly, and it seems we only use a single function's interface.

My suggestion would be remove these dependencies/remappings, and just create `src/interface/IUniswapV3SwapCallback.sol` and define the one interface we need there

contracts-periphery/foundry.toml
```diff
@@ -1,5 +1,8 @@
 [default]
-remappings = ['test/=test/']
+remappings = ['test/=test/', '@uniswap/=lib/']
 [ci]
 fuzz_runs = 10000
+remappings = ['test/=test/', '@uniswap/=lib/']
+[hook]
+remappings = ['test/=test/', '@uniswap/=lib/']
```
mds1, line 7: This profile seems like it's identical to the default one? Either way, as mentioned above I think this profile + the new remappings can be removed

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+pragma abicoder v2;
+
+import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
+import "src/utils/ISwapRouter.sol";
+
+contract WithdrawHook {
```
mds1, line 9: This file and contract should be named something more specific, like `WithdrawHookUniswap`. Otherwise we'll have to rename this once we add a second withdrawal hook contract for another protocol integration

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
```
mds1, line 1: Is there an issue somewhere describing the spec/scope of this PR? If so we should link that in the PR description. I think I get the gist of it, but am unable to verify that this PR actually implements what we want.

Example questions:
- The Uniswap router lets you swap with V2 and V3, but this hook only supports Uniswap V3.
- I believe the Uniswap router lets the integrator charge their own fee (and if it doesn't, that's something we could add ourselves). Having the option to charge our own fee seems useful.
- This only supports one swap, so a user can't receive DAI then swap some to ETH, some to USDC, and send both to the same wallet.

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
...
+    uint256 senderBalance = _amount;
+    if(_sponsorFee > 0) {
+      SafeERC20.safeTransferFrom(IERC20(_tokenAddr), address(this), _sponsor, _sponsorFee);
+      senderBalance -= _sponsorFee;
+    }
+
+    (address _dest, uint256 _swapAmount, uint256 _amountOutMinimum, uint24 _poolFee) = abi.decode(_data, (address, uint256, uint256, uint24));
+    senderBalance -= _swapAmount;
+
+    SafeERC20.safeApprove(IERC20(_tokenAddr), address(swapRouter), _swapAmount);
```
mds1, line 48: Instead of approving the router with every swap, we should consider a method like this that anyone can call to initialize this contract for that token, to save gas for users. We can add an onlyOwner modifier if we're worried about the risk of arbitrary token approvals (though this contract should never hold funds so it should be safe, but need to think on it more)

```solidity
function approveToken(IERC20 _token) external {
  _token.safeApprove(address(swapRouter), type(uint256).max)
}
```

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
...
+pragma abicoder v2;
+
+import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
+import "src/utils/ISwapRouter.sol";
+
+contract WithdrawHook {
+
+  ISwapRouter internal immutable swapRouter;
+  address internal constant WETH9 = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```
mds1, line 12: This needs to be a constructor parameter because it's address is different on every chain

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    (address _dest, uint256 _swapAmount, uint256 _amountOutMinimum, uint24 _poolFee) = abi.decode(_data, (address, uint256, uint256, uint24));
+    senderBalance -= _swapAmount;
+
+    SafeERC20.safeApprove(IERC20(_tokenAddr), address(swapRouter), _swapAmount);
+
+    ISwapRouter.ExactInputSingleParams memory params =
+      ISwapRouter.ExactInputSingleParams({
+        tokenIn: _tokenAddr,
+        tokenOut: WETH9,
```
mds1, line 53: Why do we hardcode ETH as the output here? As mentioned in the spec comment, I think both of these should be input parameters. This approach prevents someone from e.g. swapping ETH to UDSC and sending it directly to an exchange address.

mds1 (commented):
As mentioned in one of the comments, my main issue is that I'm not sure what the spec for this PR is so I'm not sure if this actually does what we want.

contracts-periphery/Makefile
apbendi: Agree! Also not sure I understand the capital T pattern here?

contracts-periphery/Makefile
```diff
@@ -0,0 +1,8 @@
+-include .env
+
+forkBlock = 14800081
+
+build :; FOUNDRY_PROFILE=hook forge build
+install :; forge install
+Test :; FOUNDRY_PROFILE=hook forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-contract WithdrawData -vv
```
apbendi, line 7: When I attempt to run this I get a weird message and no tests run 🧐 

```
$ make Test
make: `Test' is up to date.
```

contracts-periphery/src/WithdrawHook.sol
apbendi: I would probably swap the order, and name it something like `UniswapWithdrawHook`

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
...
+  function _swapForEth(
+    uint256 _amount,
+    address _stealthAddr,
+    address _acceptor,
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data) internal {
+    uint256 senderBalance = _amount;
+    if(_sponsorFee > 0) {
```
apbendi, line 40: Not sure what this section of the code is supposed to be doing. The withdrawer [will have paid](https://github.com/ScopeLift/umbra-protocol/blob/master/contracts-core/contracts/Umbra.sol#L271-L273) the sponsor fee already as part of the withdrawal. These params are simply passed forward to the hook in case the hook needed them for any reason. In this case, they can probably be ignored. Unless I'm missing something with the intent here?

contracts-periphery/src/WithdrawHook.sol
apbendi: The withdraw hook would never be used by a user withdrawing ETH, since that ETH is already in an EOA, hooks don't apply.

The use case imagined here is specifically being able to swap some of your tokens into ETH and send it all to a freshly generated address, thus avoiding the need to fund that address before using it. You could argue we could make it more flexible by allowing a swap to any token, but I don't really see a use case for that, and it does add a small amount of gas and complexity.

I've documented the imagined usecase in an issue here: https://github.com/ScopeLift/umbra-protocol/issues/345

contracts-periphery/src/WithdrawHook.sol
```diff
@@ -0,0 +1,70 @@
...
+    bytes[] memory data = new bytes[](2);
+    data[0] = abi.encodeWithSelector(swapRouter.exactInputSingle.selector, params);
+    data[1] = abi.encodeWithSelector(swapRouter.unwrapWETH9.selector, _amountOutMinimum, _dest);
+
+    swapRouter.multicall(data);
+    SafeERC20.safeTransferFrom(IERC20(_tokenAddr), address(this), _dest, senderBalance);
+  }
+
+  receive() external payable {}
+  fallback() external payable {}
```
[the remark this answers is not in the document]
apbendi: i don't see a need for these and I think they can be removed.

apbendi (approved):
Great start @garyghayrat! I don't think it will be too hard to get this to a production ready point.

contracts-periphery/src/WithdrawHook.sol
mds1: That's true, if a user received ETH they can just use the regular SwapRouter. But hardcoding ETH as an output here still prevents you from e.g. receiving DAI and swapping half for ETH and half for USDT, and sending ETH and USDT to the new address. That scenario is not mentioned in #345

contracts-periphery/Makefile
mds1: This is because there's a folder called `test`, this came up and was fixed here: https://github.com/ScopeLift/umbra-protocol/pull/338#discussion_r882234110 (though seems this PR was not built off that PR, so the change didn't carry over)

contracts-periphery/src/interfaces/ISwapRouter.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
```
mds1, line 1: Nit, but I think we should call the folder `interface`, not `interfaces`—the singular version is what we/foundry use everywhere: `test` not `tests`, `script` not `scripts`, and `src` not `srcs`

contracts-periphery/src/WithdrawHook.sol
mds1: Looks like this change was implemented without an `onlyOwner` modifier. I'm pretty sure this should be safe (still haven't spent the time thinking about it more though 😅), but just curious how you decided on this decision?

contracts-periphery/src/WithdrawHook.sol
mds1: Looking again, it seems the contract doesn't use the `WETH9` variable at all?

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "contracts-periphery/lib/solmate"]
 	path = contracts-periphery/lib/solmate
 	url = https://github.com/Rari-capital/solmate
+[submodule "contracts-periphery/lib/swap-router-contracts"]
```
mds1, line 13: I don't see this imported anywhere, is this used?

.github/workflows/ci.yaml
```diff
@@ -41,4 +41,4 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: FOUNDRY_PROFILE=ci forge test
\ No newline at end of file
+        run: make test
```
mds1, line 44: CI needs to be configured with an ETH_RPC_URL env var, see [here](https://github.com/foundry-rs/foundry/blob/911a76e46753fd266cec182d17debb5429d88c9f/.github/workflows/test.yml#L33-L34) for an example of how to configure this

contracts-periphery/foundry.toml
```diff
@@ -2,4 +2,5 @@
 remappings = ['test/=test/']
 [ci]
 fuzz_runs = 10000
+remappings = ['test/=test/']
```
mds1, line 5: As of a week or so ago foundry now adds this remapping by default, so we can remove these two remappings line. I'd also suggest adding `verbosity = 3` to the default profile. This will show traces for failing tests

contracts-periphery/src/interfaces/ISwapRouter.sol
```diff
@@ -0,0 +1,24 @@
...
+pragma solidity ^0.8.13;
+pragma abicoder v2;
```
mds1, line 3: ABI Encoder V2 is the default as of 0.8.0 so this line can be removed

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+pragma abicoder v2;
```
mds1, line 3: As mentioned above, this is not needed

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+  uint256 public feeBips;
+  address public feeCollector;
+  address payable public feeReceiver;
+
+  constructor(ISwapRouter _swapRouter, address _WETH9, uint256 _feeBips, address payable _feeReceiver) {
+  swapRouter = _swapRouter;
+  WETH9 = _WETH9;
+  feeBips = _feeBips;
+  feeReceiver = _feeReceiver;
+  }
```
mds1, line 24: Indentation is off

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+
+  ISwapRouter internal immutable swapRouter;
+  address internal WETH9;
+
+  uint256 public feeBips;
+  address public feeCollector;
+  address payable public feeReceiver;
+
+  constructor(ISwapRouter _swapRouter, address _WETH9, uint256 _feeBips, address payable _feeReceiver) {
+  swapRouter = _swapRouter;
```
mds1, line 20: Indentation in the constructor is off

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+  WETH9 = _WETH9;
+  feeBips = _feeBips;
```
mds1, line 22: Why not make the fee an input to each call, like Uniswap's router does, and just hardcode a max fee? That way we can more easily do things like:
- Change the fee without deploying a new contract
- Change fee by use case/token/for some time like a promotion or something

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+
+  function tokensWithdrawn(
+    uint256 _amount,
+    address _stealthAddr,
+    address _acceptor,
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data
+  ) external{
```
mds1, line 34: I'd recommend doing a formatting check to make things easier to read:
- Missing space before this bracket
- Close parentheses followed by modifier and open parentheses should be on their own line, e.g. L52 below should be changed to

```solidity
  bytes memory _data
) internal {
```

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  function tokensWithdrawn(
+    uint256 _amount,
+    address _stealthAddr,
+    address _acceptor,
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data
```
mds1, line 33: Can you add some comments to document what the expected format of `data` is`?

contracts-periphery/src/WithdrawHook.sol
apbendi: I would argue we really shouldn't be worrying about that scenario, because as long as the user can swap some of the token for ETH, they can then use the address afterwards to do whatever they want (including swapping DAI to USDT). I don't think we should take on complexity in our contract just to spare the user a transaction in a hypothetical use case.

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+    (bool _singleTrade, , ) = abi.decode(_data, (bool, address, ISwapRouter.ExactInputParams[]));
+    if(_singleTrade) {
+      _swapPartForEth(_amount, _stealthAddr, _acceptor, _tokenAddr, _sponsor, _sponsorFee, _data);
+    } else {
+      _swapPartForTokenPartForEth(_amount, _stealthAddr, _acceptor, _tokenAddr, _sponsor, _sponsorFee, _data);
+    }
+
+  }
+
+  function _swapPartForEth(
```
mds1, line 45: IIUC, this method swaps some of the tokens for ETH, then sends the rest of the tokens to the same recipient? If so sounds good and makes sense

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+    // Need to rethink usage of params[0].amountOutMinimum below
+    data[1] = abi.encodeWithSelector(swapRouter.unwrapWETH9WithFee.selector, params[0].amountOutMinimum, _recipient, feeBips, feeReceiver);
+
+    swapRouter.multicall(data);
+    if(senderBalance > 0) {
+      IERC20(_tokenAddr).safeTransferFrom(address(this), _recipient, senderBalance);
+    }
+  }
+
+  function _swapPartForTokenPartForEth(
```
mds1, line 68: I think a good way to implement this is similar to the above comment, where the `data` includes:
- One encoded multicall to execute the first swap to ETH and send it to the recipient
- Some struct or signal that says "swap the full balance of token X, expect a minimum output of Y, and swap into token Z and send it to the recipient". This is required because the caller won't know the remaining token X balance prior to swapping

contracts-periphery/src/WithdrawHook.sol
garyghayrat: 1. To support V2, would it be better to pass in a user param that specifies they're using V2 and add additional functions that do the swaps using V2's `_swap(address[] memory path, address _to)` OR should the only function just call `swapRouter.call(data)` data being the one from Auto Router API.
2. Implemented `unwrapWETH9WithFee`
3. One could argue that the user can just swap some DAI to ETH, then once withdrawn to a new address, they can swap ETH to USDC, which would keep the contract much simpler. 

But if we were to accommodate that scenario, I can think of an implementation that only accommodates that particular use case of swapping some for ETH and some for another token. If we want the contract to be more flexible and maybe support multiple swaps, i.e. some for eth, some for different token1 and token2, etc., I could try to implement that using multiple loops, please let me know if there's a simpler and more gas efficient way to do things that I might be missing.

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Yes

contracts-periphery/src/WithdrawHook.sol
garyghayrat: I see that user passing in the multicall data might be the easiest way as you mentioned in https://github.com/ScopeLift/umbra-protocol/pull/341#discussion_r890298937 . I was just struggling to balance how much of the logic should be done within the contract vs. by the front end that's calling it

mds1: Ok you and @apbendi agree on not supporting two token swaps and that we should only support a single swap to ETH. So my suggestion is that we just have the swap method take in the encoded multicall data that gets called on the `SwapRouter02` as mentioned in https://github.com/ScopeLift/umbra-protocol/pull/341#discussion_r890298937. This seems to solve all our problems:
- Supports Uni V2, V3, or a combination
- Lets use us their fee methods instead of re-implementing our own
- Our contract becomes very simple and just needs a single method, doesn't even need to decode the data, and just pass it on to the uniswap contract. Maybe we decode to verify the output token or something, but doesn't really seem necessary since the contract is basically just a pass-through to the router which handles the checks. 

Is there a reason not to handle it this way?

contracts-periphery/src/WithdrawHook.sol
mds1: My proposal for how to implement this is here: https://github.com/ScopeLift/umbra-protocol/pull/341#discussion_r890318294

IMO that approach is better than hardcoding ETH because it's more flexible, still limits you to a single swap, but also supports withdrawing arbitrary tokens to an exchange such as "swap DAI for USDC and send it to an exchange"

contracts-periphery/src/WithdrawHook.sol
mds1: > I was just struggling to balance how much of the logic should be done within the contract vs. by the front end that's calling it

The answer definitely varies depending on what you're doing specifically, but I don't see any benefit to doing this on-chain for our use case. It will just be more expensive, less flexible, and Uniswap's router has the required checks we need anyway

contracts-periphery/src/WithdrawHook.sol
garyghayrat: Ended up adding onlyOwner after since the frontend won't include random tokens anyway

contracts-periphery/src/WithdrawHook.sol
mds1: Bumping this

contracts-periphery/src/UniswapWithdrawHook.sol
mds1: `forge fmt` eventually, but until then I just format manually. A regex search for ` \n` is a good way to find trailing spaces

contracts-periphery/src/WithdrawHook.sol
garyghayrat: We're now taking the simpler approach in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/WithdrawHook.sol
garyghayrat: Removed in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/WithdrawHook.sol
garyghayrat: Token outputs, swap paths, are now passed in as multicall data https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/interfaces/ISwapRouter.sol
garyghayrat: Renamed as suggested in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/75faac7fcc1bbfed25b90ee04633145ddcd72693

.gitmodules
garyghayrat: Indeed it wasn't, since we are calling the contract address and using our own interface. Removed in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/fc7758fd87a1ada3e4e3d85826f2914a317f7454

.github/workflows/ci.yaml
garyghayrat: Thanks for the example! added here: https://github.com/ScopeLift/umbra-protocol/pull/341/commits/c1ad588dbc2dd150818f714ca371f0d0ea0f8578

contracts-periphery/Makefile
```diff
@@ -0,0 +1,10 @@
+-include .env
+.PHONY: test
+
+forkBlock = 14800081
+
+build :; FOUNDRY_PROFILE=hook forge build
+install :; forge install
+snapshot :; forge snapshot --match-path test/Gas.sol
```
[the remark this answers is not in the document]
garyghayrat: Yup, this branch is now rebased with master

contracts-periphery/foundry.toml
garyghayrat: Removed unnecessary mappings and added verbosity here https://github.com/ScopeLift/umbra-protocol/commit/fc7758fd87a1ada3e4e3d85826f2914a317f7454#diff-1fab2f918de2e6b852e2d49b471dc962fe328740f1a48d85b4380befe1bac691=

contracts-periphery/src/interfaces/ISwapRouter.sol
garyghayrat: Thanks! Removed them in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/75faac7fcc1bbfed25b90ee04633145ddcd72693

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Now removed https://github.com/ScopeLift/umbra-protocol/pull/341/commits/75faac7fcc1bbfed25b90ee04633145ddcd72693

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Fixed in the same commit as above https://github.com/ScopeLift/umbra-protocol/pull/341/commits/75faac7fcc1bbfed25b90ee04633145ddcd72693

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Contract should now be much different https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Thank you!

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Are these ok https://github.com/ScopeLift/umbra-protocol/pull/341/commits/02954d25e18753401f4f15642e1604aac005daa8

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+    address _acceptor,
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data
+  ) external{
+
+    (bool _singleTrade, , ) = abi.decode(_data, (bool, address, ISwapRouter.ExactInputParams[]));
+    if(_singleTrade) {
+      _swapPartForEth(_amount, _stealthAddr, _acceptor, _tokenAddr, _sponsor, _sponsorFee, _data);
```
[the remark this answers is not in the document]
garyghayrat: We're only decoding once now to get the final destination address and the multicall data https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Currently we use the multicall data that's passed in and send the leftover balance of tokens to the final recipient. https://github.com/ScopeLift/umbra-protocol/pull/341/commits/b40d4a8da174dc150d24e62c76a20189c2d8cfc2

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,117 @@
...
+    uint256 _sponsorFee,
+    bytes memory _data) internal {
+    uint256 senderBalance = _amount;
+    ( , address _recipient, ISwapRouter.ExactInputParams[] memory params) = abi.decode(_data, (bool, address, ISwapRouter.ExactInputParams[]));
+    senderBalance -= params[0].amountIn;
+
+    bytes[] memory data = new bytes[](2);
+    data[0] = abi.encodeWithSelector(swapRouter.exactInput.selector, params[0]);
+    // Need to rethink usage of params[0].amountOutMinimum below
+    data[1] = abi.encodeWithSelector(swapRouter.unwrapWETH9WithFee.selector, params[0].amountOutMinimum, _recipient, feeBips, feeReceiver);
```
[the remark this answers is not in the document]
garyghayrat: We're no longer encoding on the main contract, but I changed selectors in the test contract to `encodeCall` in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/d80e0be300b2c39b64b7bb9e55ad7e24aea73ba0

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,41 @@
...
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data
+  ) external{
+    (address _recipient, bytes[] memory _multicallData) = abi.decode(_data, (address, bytes[]));
+    swapRouter.multicall(_multicallData);
+
+    if((IERC20(_tokenAddr).balanceOf(address(this)) > 0)) {
+      IERC20(_tokenAddr).safeTransferFrom(address(this), _recipient, IERC20(_tokenAddr).balanceOf(address(this)));
```
mds1, line 34:
```suggestion
      IERC20(_tokenAddr).safeTransfer(_recipient, IERC20(_tokenAddr).balanceOf(address(this)));
```

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+  address feeReceiver = address(0x202206);
+  address constant umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+
+  address public constant DAI = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+  address public constant WETH9 = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+  address public constant Router = 0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45;
+  uint24 poolFee = 3000;
+
+  function setUp() public {
+    vm.etch(umbra, (deployCode("test/utils/Umbra.json", bytes(abi.encode(0, address(this), address(this))))).code);
```
mds1, line 69: We have this and the `pkx` definitions etc. in `DeployUmbraTest` so let's reuse those in this test

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+  bytes32 ciphertext = "ciphertext";
+
+  address alice = address(0x202204);
+  address bob = address(0x202205);
+  address feeReceiver = address(0x202206);
+  address constant umbra = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+
+  address public constant DAI = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+  address public constant WETH9 = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+  address public constant Router = 0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45;
```
mds1, line 65: A comment adding that these are mainnet addresses and these are fork tests would be helpful. Related naming convention idea: name this `UniswapWithdrawHook.fork-1.t.sol` to indicate it's a fork test against mainnet (chain ID 1)

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    dai.approve(address(umbraContract), amount);
+
+
+    umbraContract.sendToken{value: toll}(
+    address(alice),
+    address(DAI),
+    amount,
+    pkx,
+    ciphertext
+    );
```
mds1, line 91: Nit on indentation of the params here, probably can ust put them all on one line. Similarly, let's put spaces around the `=` a few lines above

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+  address public constant Router = 0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45;
+  uint24 poolFee = 3000;
+
+  function setUp() public {
+    vm.etch(umbra, (deployCode("test/utils/Umbra.json", bytes(abi.encode(0, address(this), address(this))))).code);
+    umbraContract = IUmbra(address(umbra));
+    swapRouter = ISwapRouter(Router);
+    withdrawHook = new UniswapWithdrawHook(ISwapRouter(swapRouter));
+    dai = IERC20(DAI);
+    //Owner approves tokens
```
mds1, line 74:
```suggestion
    // Owner approves tokens
```

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    address(DAI),
+    amount,
+    pkx,
+    ciphertext
+    );
+
+    vm.startPrank(alice); // Withdraw as Alice
+    address destinationAddr = bob;
+    uint256 minOut;
+    poolFee = 3000;
```
mds1, line 96: Is this supposed to be varied / swept over? We declare it as 3000 above so if not, no need to redeclare here

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    params =
+      ISwapRouter.ExactInputParams({
+        path: _path,
+        recipient : address(swapRouter),
+        amountIn : swapAmount,
+        amountOutMinimum: minOut
+      });
+
+    bytes[] memory multicallData = new bytes[](2);
+    // multicallData[0] = abi.encodeWithSelector(swapRouter.exactInput.selector, params);
```
mds1, line 111: This can be deleted

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    multicallData[0] = abi.encodeCall(swapRouter.exactInput, params);
+
+    // params.amountOutMinimum might need to be a different value
```
mds1, line 114: What does this comment mean?

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    umbraContract.withdrawTokenAndCall(address(withdrawHook), address(DAI), receiver, data);
+
+    assertEq(IERC20(DAI).balanceOf(address(bob)), amount-swapAmount);
+    assertTrue(address(destinationAddr).balance > minOut);
+  }
+
+  function testExpectRevert_HookTest() public {
+
+  }
```
mds1, line 130: Delete unused test, or do we plan to add this?

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    swapRouter = ISwapRouter(Router);
+    withdrawHook = new UniswapWithdrawHook(ISwapRouter(swapRouter));
+    dai = IERC20(DAI);
+    //Owner approves tokens
+    withdrawHook.approveToken(dai);
+    deal(address(DAI), address(this), 1e7 ether);
+  }
+
+  function testFuzz_HookTest(uint256 amount, uint256 swapAmount) public {
+    amount=bound(amount, 10 ether, 10e20);
```
mds1, line 80: 10 ETH to 100 ETH is a pretty narrow bound. You could probably do 0.1 ETH to 1000 ETH or something broader

contracts-periphery/test/UniswapWithdrawHook.t.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    bytes memory data = abi.encode(destinationAddr, multicallData);
+
+    vm.expectCall(umbra, abi.encodeWithSelector(umbraContract.withdrawTokenAndCall.selector));
+    vm.expectCall(address(withdrawHook), abi.encodeWithSelector(withdrawHook.tokensWithdrawn.selector));
+
+    umbraContract.withdrawTokenAndCall(address(withdrawHook), address(DAI), receiver, data);
+
+    assertEq(IERC20(DAI).balanceOf(address(bob)), amount-swapAmount);
+    assertTrue(address(destinationAddr).balance > minOut);
```
mds1, line 125: Can we add tighter assertions around the input/output amounts and ensure fee is properly sent to the fee recipient? Since we fork some a specific block, the amounts will be constant, meaning you can run the test once to see the results, sanity check them, and if they make sense hardcode them. We could fuzz over the recipient, fee amount, and fee receiver too

garyghayrat: Added this for now, but will try to add more fuzz https://github.com/ScopeLift/umbra-protocol/pull/341/commits/1410f85e728f1e0051733f921574124a022e85c5

contracts-periphery/test/UniswapWithdrawHook.t.sol
garyghayrat: Decided to delete at least for now because `UniswapWithdrawHook.sol` don't have any checks, not sure if we need any.

contracts-periphery/test/UniswapWithdrawHook.t.sol
garyghayrat: I just what that value should be? Maybe the Auto Router has the variable/quote that we can use?

contracts-periphery/test/UniswapWithdrawHook.t.sol
garyghayrat: Added stricter assertions in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/da3ef52c8c954203fe1cb68c85b3207c5436ac90

garyghayrat:
What different scenarios should we account for when building out the gas framework? Currently thinking `SwapHalfForEth` and `SwapAllForEth`.

contracts-periphery/test/UniswapWithdrawHook.t.sol
mds1: This is a nit, but what do you think about `UniswapWithdrawHook.fork.1.t.sol` instead? First time trying this convention of fork chain ID in the file name, and aesthetically don't love the look of the hypen. Open to other ideas too!

contracts-periphery/test/UniswapWithdrawHook.t.sol
mds1: Ah got it. Yea so in real usage the frontend will need to get the min amount out and make sure to include that. For the actual fork tests we can just pass in zero and update this comment to note that we don't care about `minOut` for the tests.

contracts-periphery/test/UniswapWithdrawHook.t.sol
mds1: Two few comments here:
1. Why do we have two separate tests that generate random addresses in a loop (one only runs for one iteration), as opposed to just using `testFuzz_HookTest` directly which will fuzz over the fee parameters?
2. Instead of `assertTrue(x > y)` it's more clear IMO to use `assertGe` / `assertGt`

mds1:
> What different scenarios should we account for when building out the gas framework? Currently thinking `SwapHalfForEth` and `SwapAllForEth`.

Yep I think that works! Reminder to make sure they're concrete tests and not fuzz tests. Are you planning on adding those in this PR too? (If so that's fine)

contracts-periphery/Makefile
```diff
@@ -6,8 +7,12 @@ else
...
+
 build :; forge build
 clean :; forge clean
 install :; $(INSTALL_CMD)
 snapshot :; forge snapshot --match-path test/UmbraBatchSend.gas.t.sol
-test :; forge test
\ No newline at end of file
+test :; forge test --no-match-path test/UniswapWithdrawHook.fork-1.t.sol && make test-hook
+test-hook :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path test/UniswapWithdrawHook.fork-1.t.sol --match-test "testFuzz_HookTest_10Addrs"
+test-hook-50 :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path test/UniswapWithdrawHook.fork-1.t.sol --match-test "testFuzz_HookTest_50Addrs"
```
mds1, line 18: I think we should clean up the make file commands a bit here:
- There's no command to just run the `UniswapWithdrawHook` tests
- Having `test-hook` and `test-hook-50` is a bit clunky—the better way to do this is with standard fuzz tests, and change the number of fuzz runs by choosing a different profile or by using the `FOUNDRY_FUZZ_RUNS` env var (more info on both in the foundry book)

contracts-periphery/foundry.toml
```diff
@@ -1,5 +1,5 @@
 [default]
-remappings = ['test/=test/']
+verbosity = 3
 [ci]
```
mds1, line 3: Nit, let's get a blank line between profiles

contracts-periphery/test/UniswapWithdrawHook.t.sol
garyghayrat: Agree that it looks better without the hyphen. Should be changed with the latest commit.

contracts-periphery/foundry.toml
garyghayrat: Should be fixed now

contracts-periphery/Makefile
garyghayrat: Implemented env variable in 70e55ca3ade277cc79101d7dc61ddabb7b022e23

garyghayrat: So CI is failing because we have to set a github env var, but what do you think about just using `setEnv` in `Setup` instead since it's not really a secret we need to hide.

garyghayrat: But then might as well just create a `uint256` variable inside the function or `setUp` that specifies how many addrs needs to be fuzzed over... so it seems like there are many options, which would you recommend?

contracts-periphery/test/UniswapWithdrawHook.t.sol
mds1: oooo nice I like that!

contracts-periphery/Makefile
mds1: In this particular case, I don't think we need to stress test the number of addresses we cover in the fuzzer since that's a straightforward thing to test, so I don't want to overcomplicate it. Your idea of a `uint256` state variable in the test contract is simple and works. We can just hardcode something small like 10 which should be sufficient.

And just make sure to add some comments explaining why we're using a for loop with manually generated addresses instead of a fuzz param

garyghayrat: Sounds good, addressed it here https://github.com/ScopeLift/umbra-protocol/pull/341/commits/c0e8d99f787cb8502889889b8fb7dfcfdd216833

contracts-periphery/Makefile
```diff
@@ -6,8 +7,11 @@ else
...
+forkBlock = 14800081
+
 build :; forge build
 clean :; forge clean
 install :; $(INSTALL_CMD)
-snapshot :; forge snapshot --match-path test/UmbraBatchSend.gas.t.sol
-test :; forge test
\ No newline at end of file
+snapshot :; forge snapshot -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path \*.gas.t.sol
+test :; forge test --match-path test/UmbraBatchSend.t.sol && make test-hook
```
garyghayrat, line 16: Note change: `make test` no longer tests `UmbraBatchSend.gas.t.sol`, should I add `&& forge test --match-path test/gas/UmbraBatchSend.gas.t.sol` to the end of this line?

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  function testFuzz_HookTest(
+    uint256 amount,
+    uint256 swapAmount,
+    uint256 feeBips
+  ) public {
+    address feeRecipient;
+    address recipient;
+    uint256 NumOfAddrs = 10;
```
mds1, line 45: nit on naming convention, same on L48 where we use this var
```suggestion
    uint256 numOfAddrs = 10;
```

contracts-periphery/Makefile
mds1: > Also bumping a comment/gas framework here in case you missed it https://github.com/ScopeLift/umbra-protocol/pull/341#issuecomment-1155652628

Ah yes I saw. Since it's not a thread there's nothing to resolve, but just gave it a thumbs up to make acknowledgement explicit—sorry!

mds1:
This is looking great @garyghayrat! 🚀

I'm going to do a final full pass later today, @apbendi let us know if you plan to do another review before merging

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: Thanks for the reviews. Fixed it here https://github.com/ScopeLift/umbra-protocol/pull/341/commits/753879207c7ce0d74824acc41ed421338a73eefc

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+      swapRouter.unwrapWETH9WithFee,
+      (params.amountOutMinimum, destinationAddr, feeBips, feeReceiver)
+    );
+
+    bytes memory data = abi.encode(destinationAddr, multicallData);
+
+    vm.expectCall(umbra, abi.encodeWithSelector(umbraContract.withdrawTokenAndCall.selector));
+    vm.expectCall(address(withdrawHook), abi.encodeWithSelector(withdrawHook.tokensWithdrawn.selector));
+
+    umbraContract.withdrawTokenAndCall(address(withdrawHook), address(DAI), receiver, data);
```
mds1, line 99: Once deployed, we'll actually end up using `withdrawTokenOnBehalf` because the stealth address will have no ETH, so we should have a test that calls that too

mds1 (commented):
Nice job @garyghayrat! Most of these final comments are nits around style and readability 👌

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+    // Owner approves tokens
+    withdrawHook.approveToken(dai);
+    deal(address(DAI), address(this), 10e23);
+  }
+
+  function testFuzz_HookTest(
+    uint256 amount,
+    uint256 swapAmount,
+    uint256 feeBips
+  ) public {
```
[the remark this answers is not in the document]
garyghayrat: My prettier is not liking it..

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+  function _testFuzz_HookTest(
+    address destinationAddr,
+    uint256 amount,
+    uint256 swapAmount,
+    uint256 feeBips,
+    address feeReceiver
+  ) public {
+    amount = bound(amount, 0.01 ether, 10e21);
+    swapAmount = bound(swapAmount, 0.01 ether, amount);
+    feeBips = bound(feeBips, 1, 100);
```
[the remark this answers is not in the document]
garyghayrat: It actually can't https://docs.uniswap.org/protocol/reference/periphery/interfaces/IPeripheryPaymentsWithFee , need to incorporate this https://docs.uniswap.org/protocol/reference/periphery/interfaces/IPeripheryPayments if we don't want to charge fees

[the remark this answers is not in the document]
garyghayrat: setting it to 1 charges 0.01% so I think it should be fine. Also if we're going to get the multicallData from the API, I believe we can just call `unwrapWETH9` instead when we don't want to charge fees.

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: You prob mean `withdrawTokenAndCallOnBehalf`, right?

contracts-periphery/Makefile
```diff
@@ -6,8 +7,11 @@ else
...
+
 build :; forge build
 clean :; forge clean
 install :; $(INSTALL_CMD)
-snapshot :; forge snapshot --match-path test/UmbraBatchSend.gas.t.sol
-test :; forge test
\ No newline at end of file
+snapshot :; forge snapshot -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path \*.gas.t.sol
+test :; forge test --match-path test/UmbraBatchSend.t.sol && make test-hook && forge test --match-path test/gas/UmbraBatchSend.gas.t.sol
+test-hook :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path test/UniswapWithdrawHook.fork.1.t.sol
```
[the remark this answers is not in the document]
garyghayrat: I might have made it worse? But check it out https://github.com/ScopeLift/umbra-protocol/pull/341/commits/63bf010e98444a791515d4318d060e4ad53dcf7b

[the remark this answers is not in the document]
garyghayrat: Latest update: https://github.com/ScopeLift/umbra-protocol/pull/341/commits/1f2e316d61a78f415f81c382983e1de64a334795

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+
+  function _testFuzz_HookTest(
+    address destinationAddr,
+    uint256 amount,
+    uint256 swapAmount,
+    uint256 feeBips,
+    address feeReceiver
+  ) public {
+    amount = bound(amount, 0.01 ether, 10e21);
+    swapAmount = bound(swapAmount, 0.01 ether, amount);
```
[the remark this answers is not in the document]
garyghayrat: Since those amounts are actually tokens it should be fine, but lowered them anyway.

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: Went down a bit of a rabbit hole playing around with signatures and cheatcodes, lmk if what I did is actually good at all https://github.com/ScopeLift/umbra-protocol/pull/341/commits/8a37728eec70f21bb10c25d8d8f581b422446c51

contracts-periphery/src/UniswapWithdrawHook.sol
garyghayrat: Should be good now, forgot to save it again after setting up prettier.

contracts-periphery/Makefile
```diff
@@ -6,8 +7,11 @@ else
...
 
+forkBlock = 14800081
+
 build :; forge build
 clean :; forge clean
 install :; $(INSTALL_CMD)
-snapshot :; forge snapshot --match-path test/UmbraBatchSend.gas.t.sol
-test :; forge test
\ No newline at end of file
+snapshot :; forge snapshot -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path \*.gas.t.sol
```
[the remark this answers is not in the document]
garyghayrat: Removed it!

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,41 @@
...
+    address _acceptor,
+    address _tokenAddr,
+    address _sponsor,
+    uint256 _sponsorFee,
+    bytes memory _data
+  ) external{
+    (address _recipient, bytes[] memory _multicallData) = abi.decode(_data, (address, bytes[]));
+    swapRouter.multicall(_multicallData);
+
+    if((IERC20(_tokenAddr).balanceOf(address(this)) > 0)) {
```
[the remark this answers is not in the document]
garyghayrat: Should be good now

contracts-periphery/src/UniswapWithdrawHook.sol
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor(ISwapRouter _swapRouter) {
+    swapRouter = _swapRouter;
+  }
+
+  /**
+   * @param _acceptor Address of this contract where withdrawn funds are sent
+   * @param _data The encoded bytes of recipient address and calldata for Uniswap SwapRouter02's multicall
+   */
+  function tokensWithdrawn(
```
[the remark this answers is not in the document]
garyghayrat: This is good to know, thanks!

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+  IQuoter quoter;
+  IERC20 dai;
+
+  uint256 toll;
+
+  // Mainnet Addresses
+  address public constant DAI = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+  address public constant WETH9 = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+  address public constant Router = 0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45;
+  address public constant Quoter = 0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6;
```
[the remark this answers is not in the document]
garyghayrat: I keep forgetting my bad

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+    );
+
+    bytes memory data = abi.encode(destinationAddr, multicallData);
+
+    vm.expectCall(umbra, abi.encodeWithSelector(umbraContract.withdrawTokenAndCall.selector));
+    vm.expectCall(address(withdrawHook), abi.encodeWithSelector(withdrawHook.tokensWithdrawn.selector));
+
+    umbraContract.withdrawTokenAndCall(address(withdrawHook), address(DAI), receiver, data);
+
+    vm.stopPrank();
```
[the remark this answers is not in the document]
garyghayrat: Used vm.prank(addr) and brought it closer!

contracts-periphery/Makefile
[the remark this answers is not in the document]
mds1: This looks great, thanks! My only nit would be to use e.g. `test-gas-batch-send`instead of `test-batch-send-gas` as the command name, that way when sorted alphabetically all the gas commands are next to each other. Not a big deal though, feel free to update if you want + mark this one resolved

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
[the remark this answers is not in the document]
mds1: No worries at all! Looks great now 👌

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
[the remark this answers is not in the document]
mds1: Heh I haven't used solidity prettier in a while but yea it can be a bit inflexible when you disagree with it's line breaks IIRC, so all good here

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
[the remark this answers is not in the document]
mds1: Ahh ok, that makes sense!

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    vm.startPrank(alice); // Withdraw as Alice
+    IUmbraHookReceiver receiver = IUmbraHookReceiver(address(withdrawHook));
+
+    bytes memory _path = abi.encodePacked(address(DAI), poolFee, WETH9);
+
+    uint256 feeReceiverPrevBalance = feeReceiver.balance;
+    uint256 minOut = quoter.quoteExactInput(_path, swapAmount);
+    uint256 feeAmount = (minOut * feeBips) / 10_000;
+    minOut -= feeAmount;
```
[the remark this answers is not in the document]
garyghayrat: Yes, it should actually be the min amount the receiver is expected receive. So I've adopted the `recipientAmountReceived` naming for this particular variable. And used `quote` instead of `minOut` for the `Quoter` estimate.

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,173 @@
...
+
+    IUmbraHookReceiver receiver = IUmbraHookReceiver(address(withdrawHook));
+
+    vm.prank(alice); // Withdraw as Alice
+    vm.expectCall(address(withdrawHook), abi.encodeWithSelector(withdrawHook.tokensWithdrawn.selector));
+
+    if (fn == UmbraFns.withdrawTokenAndCall) {
+      vm.expectCall(umbra, abi.encodeWithSelector(umbraContract.withdrawTokenAndCall.selector));
+      umbraContract.withdrawTokenAndCall(address(withdrawHook), dai, receiver, data);
+      assertEq(daiToken.balanceOf(destinationAddr), amount - swapAmount);
```
garyghayrat, line 142: Random test comment

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,173 @@
...
+        swapAmount,
+        feeBips,
+        feeRecipient,
+        sponsorFee,
+        UmbraFns.withdrawTokenAndCall
+      );
+    }
+  }
+
+  function testFuzz_HookTest_withdrawTokenAndCallOnBehalf(
```
mds1, line 72: This and `testFuzz_HookTest_withdrawTokenAndCall` both have the same `for` loop logic then call `_testFuzz_HookTest`. Maybe we dedupe that loop/logic by moving it into `_testFuzz_HookTest` instead?

contracts-periphery/test/utils/DeployUmbraTest.sol
```diff
@@ -3,28 +3,64 @@ pragma solidity ^0.8.13;
...
 
   function createMockERC20AndMint(address addr, uint256 amount) public {
-    token = new MockERC20("Test","TT", 18);
+    token = new MockERC20("Test", "TT", 18);
     token.mint(addr, amount);
   }
-}
\ No newline at end of file
+
+  function _createDigestAndSign(
```
mds1, line 34: This can be internal since it doesn't need to be called by any other contract. I also personally prefer not having the leading underscore for methods in test contracts, since for test contracts only `test*` and `setUp` need to be public and everything else should be internal anyway

```suggestion
  function createDigestAndSign(
```

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -0,0 +1,173 @@
...
+    uint256 swapAmount,
+    uint256 feeBips,
+    address feeReceiver,
+    uint256 sponsorFee,
+    UmbraFns fn
+  ) public {
+    amount = bound(amount, 10e10, 100000 ether);
+    swapAmount = bound(swapAmount, 10e8, amount);
+    feeBips = bound(feeBips, 1, 100);
+    sponsorFee = bound(sponsorFee, 1, 1000);
```
garyghayrat, line 107: Taking a fresh look at this, I can't quite recreate the issue I was having. Thinking of changing the bound for `sponsorFee` to 
```suggestion
    sponsorFee = bound(sponsorFee, 1, amount - swapAmount);
```
Which seem to pass the test with 10k runs.

garyghayrat:
> Do you have steps to reproduce or an exact error message / stack trace? Offhand I'm not exactly sure



> Do you have steps to reproduce or an exact error message / stack trace? Offhand I'm not exactly sure

Moved the `sponsorFee` relevant conversation to PR wide comment here: https://github.com/ScopeLift/umbra-protocol/pull/341#discussion_r901881355

contracts-periphery/Makefile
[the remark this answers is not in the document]
garyghayrat: I like the suggestion and changed it accordingly!

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: To do that won't we need to either not fuzz over vars like `amount`, `swapAmount`, and `feeBips` OR test both `testFuzz_HookTest_withdrawTokenAndCallOnBehalf` and `testFuzz_HookTest_withdrawTokenAndCall` within the same `_testFuzz_HookTest`?

contracts-periphery/test/utils/DeployUmbraTest.sol
```diff
@@ -3,28 +3,64 @@ pragma solidity ^0.8.13;
...
+    returns (
+      uint8,
+      bytes32,
+      bytes32
+    )
+  {
+    uint256 _chainId;
+    assembly {
+      _chainId := chainid()
+    }
```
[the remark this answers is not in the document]
garyghayrat: Nice, good to know. Changed it in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/9c8b8c07c174e352d4fe328137f80b8f4b8a52ec

contracts-periphery/test/utils/DeployUmbraTest.sol
```diff
@@ -3,28 +3,64 @@ pragma solidity ^0.8.13;
...
+    uint256 _sponsorFee,
+    IUmbraHookReceiver _hook,
+    bytes memory _data,
+    uint256 _privateKey
+  )
+    public
+    returns (
+      uint8,
+      bytes32,
+      bytes32
```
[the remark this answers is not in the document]
garyghayrat: Addressed in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/9c8b8c07c174e352d4fe328137f80b8f4b8a52ec

contracts-periphery/test/utils/DeployUmbraTest.sol
garyghayrat: Addressed in https://github.com/ScopeLift/umbra-protocol/pull/341/commits/9c8b8c07c174e352d4fe328137f80b8f4b8a52ec

.github/workflows/ci.yaml
```diff
@@ -41,4 +43,4 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: FOUNDRY_PROFILE=ci forge test
\ No newline at end of file
+        run: make test-ci
```
garyghayrat, line 46: Added new make test-ci commands.

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
mds1: I think this current approach is fine, and you don't necessarily need to dedupe because it is a small piece of code / it's a test. But I'm not sure I understand the question, because all of those are already inputs to the `hookTest` method?

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
mds1: Basically we're saying is:
- Send `amount` with Umbra
- Of that, the recipient swaps `swapAmount` of it to ETH
- This leaves a max fee for the sponsor of `amount - swapAmount`

Makes sense to me!

.github/workflows/ci.yaml
mds1: I don't think we need a special command for this. Can we define the profile as an env var as shown [here](https://github.com/foundry-rs/forge-template/blob/d18f232927dba100b614230a88af59c67f62df0e/.github/workflows/ci.yml#L8-L9) and just have this run `make test`?

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: I think the values passed in would end up being the same for every address we test, instead of being different like it is now.

.github/workflows/ci.yaml
garyghayrat: Nice, updated it here https://github.com/ScopeLift/umbra-protocol/pull/341/commits/cf4f4c0e15648b55b029ee5d98e249781b3ee22d

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
mds1: Hm why is that? Right now:
- Fuzzer generates values for amount, swapAmount, feeBips, and sponsorFee
- In a loop:
    - We generate values for the addresses
    - Pass the fuzzer data plus addresses to `hookTest` which executes the test in that loop

Possible but not required change:
- Fuzzer generates values for amount, swapAmount, feeBips, and sponsorFee
- That is passed to `hookTest`
- In a loop:
    - We generate values for the addresses
    - Then execute `hookTest` in that loop

apbendi (approved):
Looks great! Nice work on this :)

PR #343, opened 2022-05-27
Properly handle response from OZ relayer for all networks

apbendi:
Removes special case handling of relayer response and assumes response is always a regular transaction hash.

**Must be merged and deployed at the same time as** https://github.com/ScopeLift/umbra-api/pull/54

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -336,7 +336,7 @@ import AccountReceiveTableWarning from 'components/AccountReceiveTableWarning.vu
 import AccountReceiveTableWithdrawConfirmation from 'components/AccountReceiveTableWithdrawConfirmation.vue';
 import BaseTooltip from 'src/components/BaseTooltip.vue';
 import WithdrawForm from 'components/WithdrawForm.vue';
-import { ConfirmedRelayerStatusResponse, FeeEstimateResponse } from 'components/models';
+import { FeeEstimateResponse } from 'components/models';
 import { formatAddress, lookupOrFormatAddresses, toAddress, isAddressSafe } from 'src/utils/address';
 import { MAINNET_PROVIDER } from 'src/utils/constants';
 import { getEtherscanUrl } from 'src/utils/utils';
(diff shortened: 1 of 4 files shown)


frontend/src/components/AccountReceiveTable.vue
```diff
@@ -594,19 +594,12 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
...
-          );
-          const { receipt } = (await relayer.value?.waitForId(relayTransactionHash)) as ConfirmedRelayerStatusResponse;
-          console.log(vm.$i18n.t('AccountReceiveTable.withdraw-successful-receipt'), receipt);
-        }
+        // This is a regular transaction hash, though it's possible OZ Defender will replace it to ensure it gets
+        // included quickly, in which case the frontend would not automatically reflect when the relay was successful.
+        // Because we relay with the "fast" setting, this is unlikely to be the case.
+        console.log(`${vm.$i18n.t('AccountReceiveTable.relayed-with-tx-hash').toString()} ${relayTransactionHash}`);
+        const receipt = await provider.value.waitForTransaction(relayTransactionHash);
+        console.log(vm.$i18n.t('AccountReceiveTable.withdraw-successful-receipt'), receipt);
```
mds1, line 602: I know you didn't change this, but we should gradually migrate from `console.log()` to our logger class. Any objections on replacing the console.logs with `window.logger.info()`? (debug and warn are other possible options, but info seems right here)

frontend/src/components/models.ts
```diff
@@ -151,7 +151,6 @@ export type RelayResponse = { relayTransactionHash: string } | ApiError;
 export type RelayerStatusResponse =
   | { receivedTime: string; broadcasts?: any[]; receipt?: TransactionReceipt }
   | ApiError;
-export type ConfirmedRelayerStatusResponse = { receivedTime: string; broadcasts: any[]; receipt: TransactionReceipt };
```
mds1, line 154: _(arbitrary comment location to start a thread)_

> Must be merged and deployed at the same time as https://github.com/ScopeLift/umbra-api/pull/54

Given this, how do I test the new behavior? I don't think I've ever tested tested out the app using a localhost backend, so instructions for testing would be helpful!

apbendi: Good question. I've added [a commit](https://github.com/ScopeLift/umbra-api/pull/54/commits/d22c942ed373e8fa7dbcf326e5bfeaaaffc185ed) to the API PR that explains how to do this.

frontend/src/components/AccountReceiveTable.vue
apbendi: Sure! Done and amended into the commit.

mds1: Appreciate it!

frontend/src/i18n/locales/en-us.json
```diff
@@ -227,7 +227,7 @@
     "received": "Received",
     "withdrawn": "Withdrawn",
     "hide": "Hide",
-    "withdraw": "withdraw",
+    "withdraw": "Withdraw",
```
mds1, line 230: @apbendi I pushed this change which ensures the yellow "Withdraw" button in this table starts with a capital W

image

PR #349, opened 2022-06-07
update: unstoppable domains resolution library

0xfuje:
## Why?
To support new unstoppable domain names and domains minted on polygon
## Modified files:
- update @unstoppabledomains/resolution to v7.1.4 (in umbra-js/package.json)
- modify getResolutionInstance to fit new typings (in umbra-js/src/utils/utils.ts)

### DIFF
--- a/umbra-js/package.json
+++ b/umbra-js/package.json
@@ -23,7 +23,7 @@
   "author": "Matt Solomon <matt@mattsolomon.dev>, Ben DiFrancesco <ben@scopelift.co>",
   "license": "ISC",
   "dependencies": {
-    "@unstoppabledomains/resolution": "3.0.0",
+    "@unstoppabledomains/resolution": "^7.1.4",
     "ethers": "^5.5.3",
     "noble-secp256k1": "^1.2.5"
   },
--- a/umbra-js/src/utils/utils.ts
+++ b/umbra-js/src/utils/utils.ts
@@ -20,7 +20,7 @@ import {
 } from '../ethers';
 import { Point, Signature, recoverPublicKey } from 'noble-secp256k1';
 import { ens, cns } from '..';
-import { default as Resolution, Eip1993Factories } from '@unstoppabledomains/resolution';
+import { default as Resolution } from '@unstoppabledomains/resolution';
 import { StealthKeyRegistry } from '../classes/StealthKeyRegistry';
 import { TxHistoryProvider } from '../classes/TxHistoryProvider';
 import { EthersProvider, TransactionResponseExtended } from '../types';
@@ -283,9 +283,9 @@ function getResolutionInstance(provider: EthersProvider) {
   const networkName = provider.network.name === 'homestead' ? 'mainnet' : provider.network.name;
   return new Resolution({
     sourceConfig: {
-      cns: {
-        provider: Eip1993Factories.fromEthersProvider(provider),
-        network: networkName as 'mainnet' | 'rinkeby',
+      uns: {
+        api: true,
+        network: networkName === 'mainnet' ? 1 : 4,
       },
     },
   });
@@ -414,7 +414,7 @@ async function getTransactionByHash(txHash: string, provider: EthersProvider): P
   ]);
   const tx = <TransactionResponseExtended>{ ...partialTx };
   const existingFields = new Set(Object.keys(tx));
-  Object.keys(fullTx).forEach((key) => {
+  Object.keys(fullTx).forEach(key => {
     // Do nothing if this field already exists (i.e. it was formatted by the ethers formatter).
     if (existingFields.has(key)) return;
     // Otherwise, add the field and format it
--- a/umbra-js/src/utils/utils.ts
+++ b/umbra-js/src/utils/utils.ts
@@ -20,7 +20,7 @@ import {
 } from '../ethers';
 import { Point, Signature, recoverPublicKey } from 'noble-secp256k1';
 import { ens, cns } from '..';
-import { default as Resolution } from '@unstoppabledomains/resolution';
+import { default as Resolution, Eip1993Factories } from '@unstoppabledomains/resolution';
 import { StealthKeyRegistry } from '../classes/StealthKeyRegistry';
 import { TxHistoryProvider } from '../classes/TxHistoryProvider';
 import { EthersProvider, TransactionResponseExtended } from '../types';
@@ -279,13 +279,22 @@ export function isDomain(name: string) {
  * @param provider ethers provider instance
  */
 function getResolutionInstance(provider: EthersProvider) {
-  // If network name is homestead, use 'mainnet' as the network name
-  const networkName = provider.network.name === 'homestead' ? 'mainnet' : provider.network.name;
+  const eip1993Provider = Eip1993Factories.fromEthersProvider(provider);
   return new Resolution({
     sourceConfig: {
       uns: {
-        api: true,
-        network: networkName === 'mainnet' ? 1 : 4,
+        locations: {
+          Layer1: {
+            network: 'mainnet',
+            provider: eip1993Provider,
+            proxyReaderAddress: '0x58034A288D2E56B661c9056A0C27273E5460B63c',
+          },
+          Layer2: {
+            network: 'polygon',
+            provider: eip1993Provider,
+            proxyReaderAddress: '0x423F2531bd5d3C3D4EF7C318c2D1d9BEDE67c680',
+          },
+        },
       },
     },
   });
--- a/umbra-js/test/cns.test.ts
+++ b/umbra-js/test/cns.test.ts
@@ -6,11 +6,22 @@ import * as cns from '../src/utils/cns';
 import { expectRejection } from './utils';
 
 const ethersProvider = ethers.provider;
+const eip1993Provider = Eip1993Factories.fromEthersProvider(ethersProvider);
 const resolution = new Resolution({
   sourceConfig: {
-    cns: {
-      provider: Eip1993Factories.fromEthersProvider(ethersProvider),
-      network: 'rinkeby',
+    uns: {
+      locations: {
+        Layer1: {
+          network: 'mainnet',
+          provider: eip1993Provider,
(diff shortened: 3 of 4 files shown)


apbendi:
Thanks @0xfuje, just kicked off a preview build so we can do some testing! Will also take a look at the code.

umbra-js/src/utils/utils.ts
```diff
@@ -283,9 +283,9 @@ function getResolutionInstance(provider: EthersProvider) {
   const networkName = provider.network.name === 'homestead' ? 'mainnet' : provider.network.name;
   return new Resolution({
     sourceConfig: {
-      cns: {
-        provider: Eip1993Factories.fromEthersProvider(provider),
-        network: networkName as 'mainnet' | 'rinkeby',
+      uns: {
```
mds1, line 286: Looks like the tests don't run since this config change was not updated in the tests. So  I made this same config change, but the unstoppable tests in `cns.test.ts` now fail:

- `gets the public keys associated with a CNS address` fails with `ResolutionError: Domain udtestdev-msolomon.crypto is not configured`
- `throws when the user has not set their stealth keys` fails with `Domain udtestdev--c38898.crypto is not configured` (which is not the expected error message).

Tests in `utils.test.ts` fail similarly. Those were the original domains I registered for testing, but it seems they are no longer configured? I don't quite understand what changed—was everything migrated to a new system and this should still work, or was the old system discarded and/or my names expired?

umbra-js/src/utils/utils.ts
```diff
@@ -283,9 +283,9 @@ function getResolutionInstance(provider: EthersProvider) {
...
+        api: true,
```
mds1, line 287: Does this mean we rely on a UD-hosted API instead of reading directly from the chain? What are the other options/tradeoffs here?

umbra-js/src/utils/utils.ts
```diff
@@ -283,9 +283,9 @@ function getResolutionInstance(provider: EthersProvider) {
...
+        network: networkName === 'mainnet' ? 1 : 4,
```
mds1, line 288: I thought that domains are now minted on polygon, so why are the only two networks used here mainnet and rinkeby?

umbra-js/package.json
```diff
@@ -23,7 +23,7 @@
   "author": "Matt Solomon <matt@mattsolomon.dev>, Ben DiFrancesco <ben@scopelift.co>",
   "license": "ISC",
   "dependencies": {
-    "@unstoppabledomains/resolution": "3.0.0",
+    "@unstoppabledomains/resolution": "^7.1.4",
```
mds1, line 26: _(arbitrary comment location to start a thread)_

Is this below portion of the code still correct? https://github.com/ScopeLift/umbra-protocol/blob/5f5cb883d2d4d6b1cf0caed994d51c2477a04ba5/frontend/src/utils/address.ts#L40-L64

More generally, we are deployed on mainnet, polygon, arbitrum, and optimism—how should name resolution work on each network, given that IIUC domains are minted on polygon? We need to be able to resolve names in both directions: from an address to name (when a user connects their wallet, so it shows in the top right instead of an address) and from a name to an address.

Related to that, can a name resolve to a contract, and if so how should we handle that? For example, right now all ENS name resolution is done with a mainnet provider. If someone tries withdrawing funds to `vitalik.eth`, and that resolves to a contract, we warn the user because that contract may not exist on the current network, and even if it does, it might not have the same state/owners (e.g. if it's a multisig). Relevant code here:

https://github.com/ScopeLift/umbra-protocol/blob/5f5cb883d2d4d6b1cf0caed994d51c2477a04ba5/frontend/src/utils/address.ts#L132-L141

yarn.lock
```diff
@@ -5052,18 +5052,17 @@
   resolved "https://registry.yarnpkg.com/@uniswap/token-lists/-/token-lists-1.0.0-beta.21.tgz#5a4844236e328d3b0e87161c19edaace669e3e6e"
   integrity sha512-IQ1YpWKgXIsfYqym9Nl/uaB1t5Wk70pyqJ2W4FmvQmt3FIQz+pHlzlEl91aYloPrlcZaFluE0Jq7gkfYpIjsBw==
 
-"@unstoppabledomains/resolution@3.0.0":
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/@unstoppabledomains/resolution/-/resolution-3.0.0.tgz#148080757f3ab2e94993228a5934ea60f1e0807b"
-  integrity sha512-iAvVkDWsSdi9WOiCp8t6GwWQZYIQjca+JAzfnNB7LvRDim3jrgcLlbeSYu+Mwzn2JKUGf/Ygw7YtnmjihnT5fw==
+"@unstoppabledomains/resolution@^7.1.4":
```
mds1, line 5055: If I run `yarn` from the repo root, I get a diff in the lockfile—could you commit the updated version?

mds1 (commented):
Thanks for the PR @0xfuje! In general, I'm not sure what the expected behavior is supposed to be for the updated UD system, so if you could provide links to the docs and/or update the PR with these details that would be greatly appreciated. I've laid out some specific questions in the below comments.

Additionally, it seems the tests fail, so we'll need to get those working again too before merging 🙂

umbra-js/src/utils/utils.ts
0xfuje: I didn't run the tests at first, but i tested it by sending and receiving a stealth payment and it worked. I modified the config but tests still fail (and testing it by sending a payment still works), it's possible that the old system was discarded or the names expired.. i don't know honestly. Maybe a test with a newly registered ud domain be worthwhile (with it's spending and viewing keys).

umbra-js/src/utils/utils.ts
0xfuje: You don't have to rely on UD-hosted APIs, i modified the resolution to read from on-chain with Eip1993Factories provider with a new commit.

umbra-js/src/utils/utils.ts
0xfuje: I'm sorry that network line of code is not necessary and i don't think it changes anything about the functionality so I removed it.

umbra-js/package.json
0xfuje: The domains are minted on polygon, but the library resolves to a mainnet address. The library first looks for a domain on Polygon mainnet, and if it's not found there than it will check on Ethereum mainnet.

I'm afraid I don't know if the code in frontend/src/utils/address.ts is still correct and I'm couldn't find in the docs if a name can resolve to a contract. I will ask for someone more experienced with the library so you can update the code if needed.

yarn.lock
0xfuje: I committed the updated lockfile!

0xfuje:
Tested with new config: on Polygon it works fine with new UD domains, but not on ETH, OP and ARB: going to find out why that's the case.

mds1:
Thanks for the updates @0xfuje — I should be able to give this a re-review on Monday!

0xfuje:
Thank you @mds1! Please keep in mind i didn't write new tests, just tested it in a browser. Hope the PR is good with new configs :)

umbra-js/test/cns.test.ts
```diff
@@ -6,11 +6,22 @@ import * as cns from '../src/utils/cns';
...
   sourceConfig: {
-    cns: {
-      provider: Eip1993Factories.fromEthersProvider(ethersProvider),
-      network: 'rinkeby',
+    uns: {
+      locations: {
+        Layer1: {
+          network: 'mainnet',
+          provider: eip1993Provider,
+          proxyReaderAddress: '0x1BDc0fD4fbABeed3E611fd6195fCd5d41dcEF393',
```
mds1, line 17: Why do the two `proxyReaderAddress` vars in this file have different addresses than the ones in `umbra-js/src/utils/utils.ts`?

umbra-js/package.json
mds1: To be honest I'm still not sure I understand the architecture and how that maps to what we need to do for forward and reverse resolution. I'm sure I can find these answers in the docs, but just laying out some unanswered questions in this thread as a reminder for myself to look into once I have the time (or if you know the answers / have links to the docs that would also help a lot).

1. What is the the process for resolving a name to an address? Based on the PR is looks like this is still `await resolution.addr(name, 'ETH')`, but I'm not sure if always passing `ETH` is correct for all networks we support.

2. What is the the process for reverse-resolving an address to a name? Previously we used a subgraph to look up all names owned by an address and just chose the first name returned in that array. I'm not sure if that's still considered the correct way.

3. Given that an address can resolve to a contract, let's confirm the current approach used for ENS of "warn the user because that contract may not exist on the current network, and even if it does, it might not have the same state/owners (e.g. if it's a multisig)" is the correct approach.

4.  Clarification around this behavior: "The domains are minted on polygon, but the library resolves to a mainnet address. The library first looks for a domain on Polygon mainnet, and if it's not found there than it will check on Ethereum mainnet." — I don't fully understand this one, and I'm not sure what this means from a developer standpoint. 
    1. Why does it always default to polygon
    2. Why does it look on mainnet if domains are minted on polygon?
    3. How does being connected to e.g. Optimism/Arbitrum affect how we should handle domain resolution?

mds1:
I'll be away most of next week and the week after, so won't have much time to pick this back up until sometime after that. Here is my current view on what needs to be done before this is ready to merge:

1. It seems the old names we used for tests were discarded or expired, so we need to register new UNS names, set their stealth keys, and use those to make the sure the four failing tests pass. https://github.com/ScopeLift/umbra-protocol/pull/349#discussion_r892914004

2. Make sure our current logic for forward and reverse resolving UNS names is accurate. https://github.com/ScopeLift/umbra-protocol/pull/349#discussion_r911533022

3. Make sure we understand the difference in `proxyReaderAddress` values. https://github.com/ScopeLift/umbra-protocol/pull/349#discussion_r911529146

4. Add comments to document everything here so the info / relevant documentation is easy to find to anyone looking at the code

0xfuje:
Hey, sorry for the late message. Communication was lacking from unstoppable and I haven't yet got paid for my previous work so I decided to not work on UD implementations anymore. All the best to you, your app is awesome!

PR #350, opened 2022-06-22
Coinbase fix

mds1:
Adds a `fallbackProvider` to the Umbra class which uses the `INFURA_ID` env var (already used elsewhere in `umbra-js` for ENS resolution, so no changes required). 

This implemented solution is a bit naive, and still needs to be tested—if the initial attempt at withdrawing the native token fails, it retries using our fallback provider.

Also ran prettier on the full codebase, which is why the quasar file was modified


mds1:
Confirmed this fixed the user's issue

apbendi (approved):
Looks good, and didn't see any regressions when testing with MM. Good to merge/deploy!

PR #351, opened 2022-06-23
Periphery contracts gas improvements

garyghayrat:
Spent some time looking at different guides, but feel like I don't have enough to show for it. Particularly could use some help/suggestions as to how to improve multiple token transfers in a loop. https://github.com/ScopeLift/umbra-protocol/pull/351#discussion_r911371285


contracts-periphery/test/gas/UmbraBatchSend.gas.t.sol
```diff
@@ -47,34 +53,34 @@ abstract contract UmbraBatchSendGasTest is DeployUmbraTest {
...
       router.batchSendEth{value: valueAmount}(toll, sendEth);
-
     } else if (_type == Send.TOKEN) {
       assertTrue(tokenAmount > 0);
       for (uint256 i = 0; i < numOfAddrs; i++) {
         valueAmount += toll;
         sendToken.push(UmbraBatchSend.SendToken(addrs[i], address(token), tokenAmount, pkx, ciphertext));
       }
-
+      router.batchSendTokens{value: valueAmount}(toll, sendToken);
```
garyghayrat, line 62: Somehow missed this critical line and line 70 when initially merging this file

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -57,36 +59,43 @@ contract UmbraBatchSend {
...
     }
   }
 
   function _batchSendTokens(uint256 _tollCommitment, SendToken[] calldata _params) internal {
-    for (uint256 i = 0; i < _params.length; i++) {
+    uint256 length = _params.length;
+    for (uint256 i = 0; i < length;) {
       IERC20 token = IERC20(address(_params[i].tokenAddr));
 
       SafeERC20.safeTransferFrom(token, msg.sender, address(this), _params[i].amount);
```
garyghayrat, line 74: @mds1 Ik you've made a couple suggestion for this, but my implementations of them weren't great in previous PRs. Any particular direction you can point me to?

garyghayrat: Specifically transferring tokens each time

garyghayrat: Implemented SwapSummary

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,133 @@ contract UmbraBatchSend {
...
         _params[i].ciphertext
       );
+      unchecked {
+        i++;
+      }
+    }
+
+    uint256 summaryLength = _transferSummary.length;
+    for (uint256 j = 0; j < summaryLength; ) {
+      IERC20 token = IERC20(address(_transferSummary[j].tokenAddr));
```
apbendi, line 129: This local variable seems to be declared but not used

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,133 @@ contract UmbraBatchSend {
     bytes32 ciphertext;
   }
 
+  mapping(address => uint256) internal TotalTransferAmountPerToken;
```
apbendi, line 30: nit: should start with a lowercase 't' `totalTransferAmountPerToken`

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,133 @@ contract UmbraBatchSend {
...
+    TransferSummary[] calldata _transferSummary
+  ) public payable {
+    if (msg.value < _tollCommitment * _params.length) revert ValueMismatch();
+
+    for (uint256 i = 0; i < _params.length; i++) {
+      // Used later to validate total amounts are correctly provided
+      TotalTransferAmountPerToken[_params[i].tokenAddr] += _params[i].amount;
+    }
+
+    uint256 summaryLength = _transferSummary.length;
```
apbendi, line 62: nit: this and all non-storage local variables should be prefixed with `_`

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,133 @@ contract UmbraBatchSend {
...
+    for (uint256 j = 0; j < summaryLength; ) {
```
apbendi, line 63: Looks like we could re-use `i` here rather than declaring a new variable.

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,133 @@ contract UmbraBatchSend {
...
+      IERC20 token = IERC20(address(_transferSummary[j].tokenAddr));
+
+      if (TotalTransferAmountPerToken[_transferSummary[j].tokenAddr] != _transferSummary[j].amount)
+        revert TransferAmmountMismatch();
```
apbendi, line 67: nit: if have a newline, I think we should include the braces

contracts-periphery/snapshot/.gas
```diff
@@ -1,99 +1,199 @@
...
-├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
-│ Function Name   ┆ min             ┆ avg   ┆ median ┆ max   ┆ # calls │
-├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
-│ balanceOf       ┆ 534             ┆ 534   ┆ 534    ┆ 534   ┆ 4       │
-├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
-│ transfer        ┆ 29962           ┆ 29962 ┆ 29962  ┆ 29962 ┆ 8       │
-├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
-│ withdraw        ┆ 7379            ┆ 7379  ┆ 7379   ┆ 7379  ┆ 4       │
-╰─────────────────┴─────────────────┴───────┴────────┴───────┴─────────╯
+Compiling 3 files with 0.8.13
```
apbendi, line 1: I'm a bit confused by this file. How does it relate to `.gas-snapshot`? Also, when I run `make snapshot-gas` on this branch, I get a big diff in this file. Is that supposed to happen? Maybe some updates to the README to explain how gas snapshotting works would be helpful.

apbendi (commented):
Looks like some nice improvements @garyghayrat. Left some questions/comments/thoughts.

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Removed it! https://github.com/ScopeLift/umbra-protocol/blob/b156045ea6b0d75c49dbb6500e5e9e6835131723/contracts-periphery/src/UmbraBatchSend.sol#L128-L132

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: gotcha https://github.com/ScopeLift/umbra-protocol/blob/b156045ea6b0d75c49dbb6500e5e9e6835131723/contracts-periphery/src/UmbraBatchSend.sol#L30

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Should be good now https://github.com/ScopeLift/umbra-protocol/blob/bf77a1765f7683b3be691c8e28a230a0711c9279/contracts-periphery/src/UmbraBatchSend.sol#L62-L64

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: like this? It seems like we can reuse the `i` name but still have to re-declare the variable? https://github.com/ScopeLift/umbra-protocol/blob/bf77a1765f7683b3be691c8e28a230a0711c9279/contracts-periphery/src/UmbraBatchSend.sol#L57-L63

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Yup, I confused myself for a sec there too https://github.com/ScopeLift/umbra-protocol/blob/bf77a1765f7683b3be691c8e28a230a0711c9279/contracts-periphery/src/UmbraBatchSend.sol#L66-L68

contracts-periphery/snapshot/.gas
apbendi: Context from @mds1:

```
* `forge snapshot` runs your tests and reports gas usage of each test, and creates a `.gas-snapshot` file at the root
* `forge test --gas-report` outputs a table of specific functions and their min/max/med/avg gas usage, but it just prints to the terminal so you need to manually dump it to a file
```

apbendi: Since we're including both files, we want our `snapshot-gas` command to update both files

apbendi (approved):
Ok! Thanks for your incredible patience here @garyghayrat. I think we're finally in good shape 😅

contracts-periphery/foundry.toml
```diff
@@ -1,4 +1,4 @@
-[default]
+[profile.default]
 verbosity = 3
 
 [ci]
```
mds1, line 4:
```suggestion
[profile.ci]
```

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -1,11 +1,14 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
 
+import "openzeppelin-contracts/contracts/access/Ownable.sol";
 import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import "src/interface/IUmbra.sol";
+import "forge-std/console2.sol";
```
mds1, line 8: Leftover import

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,131 @@ contract UmbraBatchSend {
...
-    if(msg.value != _tollCommitment * _params.length) revert ValueMismatch();
-    _batchSendTokens(_tollCommitment, _params);
+  function batchSendTokens(
+    uint256 _tollCommitment,
+    SendToken[] calldata _params,
+    TransferSummary[] calldata _transferSummary
+  ) public payable {
+    if (msg.value < _tollCommitment * _params.length) revert ValueMismatch();
+
+    for (uint256 i = 0; i < _params.length; i++) {
```
mds1, line 57: Can use unchecked when incrementing here. If you want to dedupe a bit in all of the for loops you can do:

```solidity
function uncheckedIncrement(uint256 i) internal pure returns (uint256) {
  unchecked { return i + 1; }
}

for (uint256 i = 0; i < _params.length; i = uncheckedIncrement(i)) {
  // --- snip ---
}
```

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,131 @@ contract UmbraBatchSend {
...
+
+    uint256 _summaryLength = _transferSummary.length;
+    for (uint256 i = 0; i < _summaryLength; ) {
+      IERC20 _token = IERC20(address(_transferSummary[i].tokenAddr));
+
+      if (totalTransferAmountPerToken[_transferSummary[i].tokenAddr] != _transferSummary[i].amount) {
+        revert TransferAmmountMismatch();
+      }
+
+      SafeERC20.safeTransferFrom(_token, msg.sender, address(this), _transferSummary[i].amount);
```
mds1, line 70: IMO it's a little bit cleaner/more conventional to delete the cast on L64 and change this line to:
```suggestion
      IERC20(_token).safeTransferFrom(msg.sender, address(this), _transferSummary[i].amount);
```

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,69 +27,131 @@ contract UmbraBatchSend {
...
+    SendToken[] calldata _tokenParams,
+    TransferSummary[] calldata _transferSummary
   ) external payable {
     uint256 _totalAmount = _sumValueFromEthSends(_tollCommitment, _ethParams) + (_tollCommitment * _tokenParams.length);
-    if(msg.value != _totalAmount) revert ValueMismatch();
+    if (msg.value != _totalAmount) revert ValueMismatch();
 
     _batchSendEth(_tollCommitment, _ethParams);
-    _batchSendTokens(_tollCommitment, _tokenParams);
+    batchSendTokens(_tollCommitment, _tokenParams, _transferSummary);
```
mds1, line 90: Why do we call the internal batch send ETH method but the external batch send tokens method?

contracts-periphery/foundry.toml
garyghayrat: added

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: removed.

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Agreed. Changed.

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: But actually that might make less sense the more i think about it

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,54 +26,86 @@ contract UmbraBatchSend {
...
   }
 
+  mapping(address => uint256) internal totalTransferAmountPerToken;
+  struct TransferSummary {
+    uint256 amount;
+    address tokenAddr;
+  }
+
   error ValueMismatch();
+  error TransferAmmountMismatch();
```
mds1, line 36:
```suggestion
  error TransferAmountMismatch();
```

contracts-periphery/src/UmbraBatchSend.sol
mds1: Looks like there's a loop in `batchSendTokens` that needs to be changed

contracts-periphery/src/UmbraBatchSend.sol
mds1: What if we extract the eth value checks into a method that takes two arrays: one of `SendEth[]`, one of `SendToken[]`. If we're only sending ETH, we pass an empty array of `SendToken` into the helper method. Then we can reuse the same validation logic and simplify things a bit

contracts-periphery/.gas-snapshot
```diff
@@ -0,0 +1,154 @@
+BatchSendWithTollAndWithTokenBalanceGasTest:testPostSetupState() (gas: 11857)
```
mds1, line 1: _(arbitrary comment location to start a thread)_

Nit: can we delete `fs_permissions = [{ access = "read", path = "./"}]` from the `profile.ci` section? If a key is undefined, it automatically uses the default value, so we don't need to redefine it twice

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,54 +26,86 @@ contract UmbraBatchSend {
...
+    for (uint256 i = 0; i < _params.length; i++) {
+      // Used later to validate total amounts are correctly provided
+      totalTransferAmountPerToken[_params[i].tokenAddr] += _params[i].amount;
+    }
+
+    uint256 _summaryLength = _transferSummary.length;
+    for (uint256 i = 0; i < _summaryLength; i = uncheckedIncrement(i)) {
+      IERC20 _token = IERC20(address(_transferSummary[i].tokenAddr));
+
+      if (totalTransferAmountPerToken[_transferSummary[i].tokenAddr] != _transferSummary[i].amount) {
```
mds1, line 65: We never clear this storage mapping so won't subsequent sends revert with a mismatch here?

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: good catch, fixed

contracts-periphery/.gas-snapshot
garyghayrat: gotcha

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Addressed here https://github.com/ScopeLift/umbra-protocol/pull/351/commits/11ee5e7c67649edb0601ab23fcc802dae11f56f8

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -1,11 +1,13 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
 
-import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
-import "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
+import "../lib/openzeppelin-contracts/contracts/access/Ownable.sol";
+import "../lib/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "../lib/openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
```
mds1, line 6: Why the change to relative paths?

contracts-periphery/src/UmbraBatchSend.sol
```diff
@@ -24,54 +26,83 @@ contract UmbraBatchSend {
...
+    SendToken[] calldata _tokenParams,
+    TransferSummary[] calldata _transferSummary
   ) external payable {
-    uint256 _totalAmount = _sumValueFromEthSends(_tollCommitment, _ethParams) + (_tollCommitment * _tokenParams.length);
-    if(msg.value != _totalAmount) revert ValueMismatch();
-
-    _batchSendEth(_tollCommitment, _ethParams);
-    _batchSendTokens(_tollCommitment, _tokenParams);
+    batchSendEth(_tollCommitment, _ethParams);
+    batchSendTokens(_tollCommitment, _tokenParams, _transferSummary);
```
mds1, line 83: This duplicates the `_checkValueFromSends` validation right? I think we can:
- Make both `function batchSend*` methods external instead of public
- Move the event emissions into the internal `function _batchSend*` methods
- In `batchSendTokens` method, it should mirror the `batchSendEth` method in that it only (1) checks the values, and (2) calls the internal version. So the for loops in `batchSendTokens` can be internal to `_batchSendTokens` instead
- All `_checkValueFromSends` calls are in the external methods, and the internal ones handle everything else

I think that should work?

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: Was getting
```
Compiler run failed
error[6275]: ParserError: Source "openzeppelin-contracts/contracts/access/Ownable.sol" not found: File not found. Searched the following locations: ".../umbra-protocol/contracts-periphery".
 --> src/UmbraBatchSend.sol:4:1:
  |
4 | import "openzeppelin-contracts/contracts/access/Ownable.sol";
```

contracts-periphery/src/UmbraBatchSend.sol
garyghayrat: It does. Implemented here: https://github.com/ScopeLift/umbra-protocol/pull/351/commits/fec577097bc55395bf33e0a7079845e3323a95df

contracts-periphery/src/UmbraBatchSend.sol
mds1: Huh weird, this didn't used to happen though right? Either way, IMO it's preferable to solve this with remappings since relative paths aren't as nice to work with (things break if you move around files, less clear where the file comes from)

```toml
remappings = [
  "openzeppelin-contracts/=lib/openzeppelin-contracts/contracts"
]
```

```solidity
import "openzeppelin-contracts/access/Ownable.sol";
```

mds1: Reminder that you only need to define it in the default profile and the CI profile will inherit that, so we currently have:

```toml
[profile.default]
verbosity = 3
fs_permissions = [{ access = "read", path = "./"}]
remappings = ['openzeppelin-contracts/=lib/openzeppelin-contracts/contracts']
[profile.ci]
fuzz_runs = 10000
fs_permissions = [{ access = "read", path = "./"}]
remappings = ["openzeppelin-contracts/=lib/openzeppelin-contracts/contracts"]
# See more config options https://github.com/gakonst/foundry/tree/master/config
```

But we only need:
```toml
[profile.default]
verbosity = 3
fs_permissions = [{ access = "read", path = "./"}]
remappings = ['openzeppelin-contracts/=lib/openzeppelin-contracts/contracts']

[profile.ci]
fuzz_runs = 10000
```

garyghayrat: gotcha, removed duplicates https://github.com/ScopeLift/umbra-protocol/pull/351/commits/88f80ca520f5b317c3fb77e0f0bcf0768b7d9a0e

mds1 (approved):
Thanks @garyghayrat! Can you just rebase against main, now that https://github.com/ScopeLift/umbra-protocol/pull/393 was just merged that should fix CI here

garyghayrat:
> Thanks @garyghayrat! Can you just rebase against main, now that #393 was just merged that should fix CI here

Rebased!

apbendi (approved):
Looks good to me!

PR #356, opened 2022-06-30
Stop using ethers default api keys

apbendi:
Switch to static provider with our infura id env var injected to avoid rate limiting associated with the default provider.

### DIFF
--- a/umbra-js/test/utils.test.ts
+++ b/umbra-js/test/utils.test.ts
@@ -1,11 +1,14 @@
 import { ethers } from 'hardhat';
-import { getDefaultProvider } from '@ethersproject/providers';
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import { expect } from 'chai';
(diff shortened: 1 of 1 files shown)


mds1:
We're using `getDefaultProvider`, which "connects to multiple backends and verifies their results internally, making it simple to have a high level of trust in third-party services". So even though we added an Infura key, this message shows up for other  providers.

Switching from `getDefaultProvider` to any of `JsonRpcProvider`, `StaticJsonRpcProvider`, or `InfuraProvider` should all fix this. The latter two reduce the number of RPC requests because you know the provider won't suddenly change chain ID on you (I'm going to have a PR later to use StaticJsonRpcProvider elsewhere in the app to reduce the number of requests)

apbendi:
Updated to use StaticJsonRpcProvider, ready for review @mds1

mds1 (approved):
LGTM!

PR #365, opened 2022-07-09
Full Repository CI

apbendi:
Configure CI to run the full monorepo build & test

* Updates the CI config to build and test the full monorepo
  from the root directory, including all secrets needed for testing
* Removes outdated GSN contracts and tests which caused complications
  with running testrunner


.github/workflows/ci.yaml
```diff
@@ -8,40 +8,43 @@ on:
...
-        working-directory: contracts-periphery/
+      INFURA_ID: ${{ secrets.INFURA_ID }}
+      ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+      POLYGONSCAN_API_KEY: ${{ secrets.POLYGONSCAN_API_KEY }}
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
 
+      - uses: volta-cli/action@v1
```
wildmolasses, line 39: I notice both build and test install foundry, volta, and yarn packages. The yarn install step takes 5 min on each job! I looked a little into whether portions of the workflow can be reused across jobs, but it doesn't seem like it

wildmolasses (approved):
Lgtm!

.github/workflows/ci.yaml
apbendi: Thanks for looking into this. I couldn't find a way either. It's a bummer we duplicate this but I don't see a way to change it.

PR #370, opened 2022-08-03
Add try-catch block to POAP API call

garyghayrat:
This simple solution seems to work even when the API url is wrong.

### DIFF
--- a/frontend/src/utils/address.ts
+++ b/frontend/src/utils/address.ts
@@ -147,6 +147,12 @@ const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
 
 // Returns true if the address owns any POAP tokens
 const hasPOAPs = async (address: string) => {
-  const poaps = await jsonFetch(`https://api.poap.xyz/actions/scan/${address}`);
-  return poaps.length > 0 ? true : false;
+  try {
+    const poaps = await jsonFetch(`https://api.poap.xyz/actions/scan/${address}`);
+    return poaps.length > 0 ? true : false;
+  } catch (err) {
+    console.warn('Error in hasPOAPs');
+    console.warn(err);
+    return false;
+  }
 };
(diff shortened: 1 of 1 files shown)


frontend/src/utils/address.ts
```diff
@@ -147,6 +147,12 @@ const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
...
 // Returns true if the address owns any POAP tokens
 const hasPOAPs = async (address: string) => {
-  const poaps = await jsonFetch(`https://api.poap.xyz/actions/scan/${address}`);
-  return poaps.length > 0 ? true : false;
+  try {
+    const poaps = await jsonFetch(`https://api.poap.xyz/actions/scan/${address}`);
+    return poaps.length > 0 ? true : false;
+  } catch (err) {
+    console.warn('Error in hasPOAPs');
+    console.warn(err);
```
mds1, line 155: Reminder to always use `window.logger` in the frontend

frontend/src/utils/address.ts
```diff
@@ -147,6 +147,12 @@ const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
...
+    return false;
```
mds1, line 156: Hmm are we ok with assuming `false` when this query fails? Maybe we should default to `true` to be more conservative?

apbendi (approved):
Looks and works well! Nice @garyghayrat

frontend/src/utils/address.ts
apbendi: I don't think so. A false positive could cause a lot of confusion for the user, especially if they're using an address that is totally safe.

mds1: Yea, the third options is return something other tan true/false and say "may have POAPs, not sure, the check failed" and let the user handle that information, but that feels overly complex

mds1: We're doing this "to be helpful" to users—this list is not comprehensive and there's other things that could break privacy, so I think false is ok

frontend/src/utils/address.ts
garyghayrat: Noted! Addressed it here https://github.com/ScopeLift/umbra-protocol/pull/370/commits/ff0b19c2d6e044393dee68cdf5716548d8d5f884

mds1 (approved):

PR #371, opened 2022-08-04
Add try-catch block to input validation

garyghayrat:
Fix to withdraw button locking up when it is clicked without a valid address input.


frontend/src/components/AccountReceiveTable.vue
```diff
@@ -514,18 +514,25 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
...
-      provider.value
-    );
-
-    if (safe) {
-      showConfirmationModal.value = true;
-    } else {
-      showPrivacyModal.value = true;
-      privacyModalAddressWarnings.value = reasons;
+
+    try {
```
mds1, line 518: Instead of a try/catch, wdyt about adding

```typescript
if (!destinationAddress.value) throw new Error(vm.$i18n.tc('AccountReceiveTable.missing-destination-address'));
```

beneath the other two initial condition checks (and seems the comment on L513 is out of place) with the above having a new i18n key for a custom error message?

garyghayrat: I tried that and 
1. It seems to put the withdraw button on the loading state even though I did `setIsInWithdrawFlow(false);` right before throwing the error. 
2. If destination address is filled out but isn't safe, `isAddressSafe` throws an error that isn't handled, therefore keeping the withdraw flow true, again stuck loading.

So I went with try/catch.

mds1: Ok that works, thanks. Looking at `isAddressSafe`, it probably should be refactored so it dispatches the async calls in a [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) as much as possible, instead of synchronously like the current approach. That way failed promises can be handled within `isAddressSafe` instead of having the caller handle it. 

I think it's ok to track this in a separate issue and keep the current solution for now

garyghayrat: Created the issue https://github.com/ScopeLift/umbra-protocol/issues/373

mds1 (approved):

PR #372, opened 2022-08-05
fix: clear all state on network/address change by refreshing

mds1:
Sometimes refreshing during a scan was slow (guess it needs to wait for the event loop to free up to refresh? not certain), but otherwise this seems to work as you'd expect by refreshing when you change address or network. Network changes from the app or from the wallet should both work

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -83,14 +83,14 @@ export default function useWalletStore() {
           wallet: (wallet) => {
             setProvider(wallet.provider);
           },
-          address: async (address) => {
+          address: (address) => {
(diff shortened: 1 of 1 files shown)


apbendi (commented):
Change seems simple enough, but in testing, it seems flaky and I don't understand why. Sometimes it seems to just hang.

umbra-network-switch-refresh-issues

mds1:
What's going on in your network tab when that happens, and what arbitrum RPC are you using, and does that happen right now on the main deploy? 

Mainly thinking it's just a slow RPC and all our requests are taking a while

apbendi (approved):
Working well now that my Brave/MM issues have been resolved. The reload is pretty quick, actually, not even that much of a sacrifice UX wise.

Issue #363, opened 2022-07-06
Sending max balance


garyghayrat:
On the send page, let users click a 'Max' button to send max balance of selected token.
- Might be tricky with ETH because you need to estimate amount for gas.

mds1:
We have a method called [`getEthSweepGasInfo `](https://github.com/ScopeLift/umbra-protocol/blob/72eee76c4f17f88aac2b88371bc408ff0452416a/umbra-js/src/utils/utils.ts#L338) that we use on the withdraw page to sweep all ETH from a stealth address. We should be able to use this same method on the send side.

mds1:
Closed by #375, the native token sweeps are being tracked in #378

PR #375, opened 2022-08-10
Feature/add max

alexkeating:
Description:

- Added a max button for the send page #363 

[umbra_max.webm](https://user-images.githubusercontent.com/12705146/183792644-3c78ffb9-185a-4f5f-9588-a8deb56427b8.webm)


frontend/src/pages/AccountSend.vue
```diff
@@ -398,6 +409,22 @@ function useSendForm() {
...
   }
 
+  async function setHumanAmountMax() {
+    if (!token.value?.address)  throw new Error(vm.$i18n.tc('Send.select-a-token'))
+    if (NATIVE_TOKEN.value?.address === token.value?.address) {
+        const address = await signer.value?.getAddress()
+        if (!address || !provider.value) throw new Error(vm.$i18n.tc('Send.wallet-not-connected'))
+        if (!recipientId.value) throw new Error(vm.$i18n.tc('Send.enter-a-recipient'))
+        const {ethToSend} = await umbraUtils.getEthSweepGasInfo(address || "", recipientId.value, provider.value)
+        humanAmount.value = formatUnits(ethToSend.toString(), token.value.decimals)
```
mds1, line 419: For both `humanAmount.value =` lines, let's use the `humanizeTokenAmount` helper instead. There's a lot of quirks to make the numbers look nice with varying decimal numbers/magnitudes so that method handles all of them

frontend/src/pages/AccountSend.vue
```diff
@@ -72,7 +72,18 @@
       />
 
       <!-- Amount -->
-      <div>{{ $t('Send.amount') }}</div>
+      <div class="flex justify-between items-center">
```
mds1, line 75: _(arbitrary comment location to start a thread)_

I haven't tested the actual sending yet, but one issue we'd run into is with the native token: Gas price might change from the time you fill in the max to the time you send your transaction, resulting in either (1) an insufficient balance error if gas price increases, or (2) dust leftover if gas price decreases. So for the native token we'll need a flag that indicates the user clicked the "Max" button and recompute the the actual max at the time of the send transaction. 

We also might want to throw an error if the amount changes by some threshold (+/- 1-2% maybe?) to prevent the user from sending a lot less than they expect if gas prices quickly change, but that might not be worth the added complexity for now.

I'm also ok with implementing this in a follow-up PR if we prefer, I'll defer to @garyghayrat or @apbendi on priority for that

mds1 (commented):
Thank you @alexkeating! 🙌

frontend/src/i18n/locales/zh-cn.json
```diff
@@ -71,6 +71,7 @@
     "total": "总额",
     "copy-payment-link": "复制付款链接",
     "enter-an-amount": "请输入金额",
+    "enter-a-recipient": "請輸入收件者",
```
garyghayrat, line 74:
```suggestion
    "enter-a-recipient": "请输入收款人",
```
 Hi Alex, just changed these to simplified Chinese.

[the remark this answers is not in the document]
garyghayrat:
```suggestion
    "enter-a-recipient": "请输入收款人",
```
Hi Alex, just changed these to simplified Chinese.

frontend/src/pages/AccountSend.vue
garyghayrat: I'd say let's do it in a follow-up PR. For now @alexkeating feel free to open an issue for it.

apbendi: Yep agreed @garyghayrat.

frontend/src/i18n/locales/zh-cn.json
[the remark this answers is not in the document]
alexkeating: Thanks, updated

frontend/src/pages/AccountSend.vue
```diff
@@ -72,7 +72,18 @@
...
+        {{ $t('Send.amount') }} 
+        <base-button
```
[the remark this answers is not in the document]
alexkeating: Added

frontend/src/pages/AccountSend.vue
alexkeating: Thanks, updated to use that helper. Although, I noticed a bug when the value was a 1,000. It seems like the comma is causing an error to pop up.

frontend/src/pages/AccountSend.vue
```diff
@@ -398,6 +409,22 @@ function useSendForm() {
     sendFormRef.value?.resetValidation();
   }
 
+  async function setHumanAmountMax() {
+    if (!token.value?.address)  throw new Error(vm.$i18n.tc('Send.select-a-token'))
+    if (NATIVE_TOKEN.value?.address === token.value?.address) {
+        const address = await signer.value?.getAddress()
```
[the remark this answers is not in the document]
alexkeating: Thank you, updated!

frontend/src/pages/AccountSend.vue
```diff
@@ -168,7 +179,7 @@ import useWalletStore from 'src/store/wallet';
 // --- Other ---
 import { txNotify } from 'src/utils/alerts';
 import { BigNumber, Contract, getAddress, MaxUint256, parseUnits, formatUnits, Zero } from 'src/utils/ethers';
-import { humanizeTokenAmount, humanizeMinSendAmount, humanizeArithmeticResult } from 'src/utils/utils';
+import { humanizeTokenAmount, humanizeMinSendAmount, humanizeArithmeticResult,  } from 'src/utils/utils';
```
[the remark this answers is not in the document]
alexkeating: I ran this now, and will investigate why my local isn't applying prettier on save

frontend/src/i18n/locales/zh-cn.json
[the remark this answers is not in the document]
mds1: I think this didn't get updated properly?

```javascript
"enter-a-recipient": "請輸入收件者", // original
"enter-a-recipient": "请输入收款人", // suggested
"enter-a-recipient": "请输入收件人", // current
```

frontend/src/pages/AccountSend.vue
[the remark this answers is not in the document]
mds1: Ah sorry, to clarify I don't think it's an issue with your local specifically, this has happened before

frontend/src/pages/AccountSend.vue
mds1: Oh interesting! Do you have more info / steps to reproduce? Would be great to track in an issue 🙂 

(unresolved to discuss that bug)

frontend/src/pages/AccountSend.vue
mds1: Great, tracking in https://github.com/ScopeLift/umbra-protocol/issues/378

frontend/src/pages/AccountSend.vue
```diff
@@ -398,6 +409,21 @@ function useSendForm() {
     sendFormRef.value?.resetValidation();
   }
 
+  async function setHumanAmountMax() {
+    if (!token.value?.address) throw new Error(vm.$i18n.tc('Send.select-a-token'));
+    if (NATIVE_TOKEN.value?.address === token.value?.address) {
+      const address = userAddress.value;
+      if (!address || !provider.value) throw new Error(vm.$i18n.tc('Send.wallet-not-connected'));
+      if (!recipientId.value) throw new Error(vm.$i18n.tc('Send.enter-a-recipient'));
+      const { ethToSend } = await umbraUtils.getEthSweepGasInfo(address || '', recipientId.value, provider.value);
```
mds1, line 418: I swear I left a comment about this line but apparently not, sorry! 😅 Two things:
1. We can change `address || ''` to just `address`, since we throw an error a few lines up if it's undefined
2. Using an ENS name as the recipient like `msolomon.eth` results in an error, because `getEthSweepGasInfo` wants an address but we only give it a string. `address.ts` has a `toAddress` method you can use so this should be an easy fix with just `await toAddress(recipientId.value)`

frontend/src/pages/AccountSend.vue
alexkeating: Here is a video of a reproduction
[format_bug.webm](https://user-images.githubusercontent.com/12705146/184269664-1053d1b6-0cea-4005-8fda-a545209113ba.webm)

frontend/src/i18n/locales/zh-cn.json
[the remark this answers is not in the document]
alexkeating: Ah sorry, I updated again it should match the suggestion

frontend/src/pages/AccountSend.vue
alexkeating: No worries, updated

frontend/src/pages/AccountSend.vue
mds1: Ahhh sorry yes this makes total sense, we should not have commas in the input field. Had a brain fart apparently. I'll push a commit putting back what you originally had

mds1: Pushed f31e279 which should put it back, lmk if that looks right to you

frontend/src/pages/AccountSend.vue
mds1: The current approach isn't exactly correct, sorry if I was unclear! Right now we have the below code, which I annotated with comments:

```typescript
// This line is basically a no-op, it inputs the user's address and returns it
const address = await toAddress(userAddress.value, provider.value);

// This line is "correct", but reverts because `recipientId.value` might be an ENS name.
// On mainnet, ethersjs will handle that automatically, but on L2s it won't.
const { ethToSend } = await umbraUtils.getEthSweepGasInfo(address, recipientId.value, provider.value);
```

What we want is something like this:

```typescript
// This is basically what you had before, which is correct, and I just renamed
// the variable name to make it more clear. I also added the `!` which tells
// Typescript that we are certain the address will be defined (since you have
// to connect a wallet to access this page, but TS doesn't know that)
const fromAddress = userAddress.value!;

// Now we take the recipient ID, i.e. who you are sending to, and make sure
// it's an address. This can be a string, like an ENS name, so we must resolve
// it to a standard address.
const toAddress = await toAddress(recipientId.value, provider.value);

// Now we calculate the sweep info, and the first two arguments are now
// standard addresses.
const { ethToSend } = await umbraUtils.getEthSweepGasInfo(fromAddress, toAddress, provider.value);
```

I'd recommend testing the functionality on a few networks, like mainnet and arbitrum, with both regular addresses and ENS names, to make sure it works as expected. An ENS name like `vitalik.eth` or `msolomon.eth` should work for all of them (except rinkeby—on live networks, we also resolve ENS names against mainnet, but on rinkeby (since it's a testnet) we resolve directly on rinkeby).

If you haven't already, you can also use your own ENS name to go through the full cycle of sending the max to yourself, then withdrawing to yourself, to both test the send process but also to have a better understanding of the full flow 🙂 (ideally with rinkeby to avoid losing real funds if the PR doesn't work as intended)

alexkeating: Oh sorry, I misunderstood. That makes sense, thank you for the thorough explanation. I will resolve this

alexkeating: Updated, and tested sending to your ENS on rinkeby

mds1: Hmm, I did get your rinkeby ETH, but I'm not sure how it worked since we're now not using `recipientId.value` at all—it seems the recipient address is now set to the sender's address and not whatever `recipientId.value` resolves to?

https://github.com/alexkeating/umbra-protocol/blob/1bd08cd3f4ca09efe260cfb6613e22fd342d344a/frontend/src/pages/AccountSend.vue#L416-L417

alexkeating: Yes, that is weird. I added replaced the offending line with recipientId.value

frontend/src/i18n/locales/zh-cn.json
```diff
@@ -69,8 +69,10 @@
     "fee": "Umbra 费",
     "fee-explain": "在 %{chainName} 上交易非常便宜，因此会收取少量费用以阻止垃圾信息。",
     "total": "总额",
+    "max": "最大限度",
```
mds1, line 72: @garyghayrat think I forgot to ask you to confirm this translation

frontend/src/pages/AccountSend.vue
mds1: awesome, thanks! pushed one commit to refactor the checks logic a bit to make it a little cleaner, but looks great now, thakns!

mds1 (approved):

mds1:
Tests pass when I run locally, so we can ignore the CI test failures

frontend/src/i18n/locales/zh-cn.json
garyghayrat:
```suggestion
    "max": "最大限额",
```

mds1:
ETH sends are flaky due to changing gas prices, so we're going to hide the "Max" button for the native token in this PR, then add it back in as part of #378

alexkeating:
Added hiding max, but left the native logic in setHumanAmountMax because I figured we would need it later.

alexkeating:
Added commit to handle undefined token states

garyghayrat (approved):

mds1:
Will merge/deploy monday morning to avoid any unforseen issues that may come up so we don't need to triage them over the week. Thanks @alexkeating and @garyghayrat!

Issue #135, opened 2021-03-29
After withdraw, clicking "Withdrawn" in receive table should take user to withdrawal tx on Etherscan


mds1:

apbendi:
Possible easy first pass: link only to the stealth address

apbendi:
#381 adds links to the stealth account for base-asset (ETH/MATIC) only, because it's readily available and obvious to the user what that means. We leave it out for tokens for now, and backlog the issue for future consideration.

PR #381, opened 2022-08-22
Add Etherscan links to sender and receiver addresses 

garyghayrat:
Partially addresses https://github.com/ScopeLift/umbra-protocol/issues/135
Screenshot_2022-08-31_15-24-06 Native token on top, ERC20 on the bottom


frontend/src/components/AccountReceiveTable.vue
```diff
@@ -233,15 +233,23 @@
               <!-- Sender column -->
               <div v-else-if="col.name === 'from'" class="d-inline-block">
                 <div @click="copyAddress(props.row.from, 'Sender')" class="cursor-pointer copy-icon-parent">
-                  <span>{{ col.value }}</span>
+                  <span
+                    ><a :href="getSenderOrReceiverEtherscanUrl(col.value)" target="_blank">{{
```
mds1, line 237: Looks like some formatting got messed up? Right now it has the default browser style which looks ugly IMO

image

And it also cuts off the info, whereas before it looked like this

image

You may want to add `class="hyperlink"` which is the style we usually give to hyperlinks. However, I think that would get noisy with a big table so it may be better to just keep the regular text style

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -473,7 +487,7 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
 
     // Format addresses to use ENS, CNS, or formatted address
     const fromAddresses = announcements.map((announcement) => announcement.from);
-    const formattedAddresses = await lookupOrFormatAddresses(fromAddresses, MAINNET_PROVIDER as Web3Provider);
+    const formattedAddresses = await lookupOrReturnAddresses(fromAddresses, MAINNET_PROVIDER as Web3Provider);
```
mds1, line 490: Why was this change made? It seems we added a method which is similar to `lookupOrFormatAddress` but doesn't do the formatting, e.g. it returns the full address and not a truncated version?

If we keep this change, it means we no longer use `lookupOrFormatAddress[es]` and should delete those

frontend/src/components/AccountReceiveTable.vue
garyghayrat: I actually do prefer "hyperlink" styling

frontend/src/components/AccountReceiveTable.vue
garyghayrat: I changed it because the sender's address `announcement.from` was actually being overwritten by the truncated version here https://github.com/ScopeLift/umbra-protocol/blob/34f45f3ce389d5aecc603814adfadcab5c18758a/frontend/src/components/AccountReceiveTable.vue#L476-L479 The copy address button would end up copying the truncated version of the address and the ehterscan link returned 404 because of it. So I ended up formatting inside the template before displaying. https://github.com/ScopeLift/umbra-protocol/blob/32dbe9b5356ef0b55603100c9cbdffd44233a174/frontend/src/components/AccountReceiveTable.vue#L238

To solve the ENS truncation problem tho, should I maybe create a new announcement value that stores a formatted and an unformatted version of addresses so I can link to unformatted ones but show the formatted ones?

frontend/src/components/AccountReceiveTable.vue
mds1: You mean the hyperlink class, not the default blue style? If so that's fine, though we do still need to fix the "sender" name/addr being truncated like in the first screenshot

frontend/src/components/AccountReceiveTable.vue
mds1: > I changed it because the sender's address announcement.from was actually being overwritten by the truncated version here
> 
> The copy address button copied the truncated version of the address and the ehterscan link returned 404 because of it. So I ended up formatting inside the template before displaying.

This all makes sense.

> To solve the ENS truncation problem tho, should I maybe create a new announcement value that stores a formatted and an unformatted version of addresses so I can link to unformatted ones but show the formatted ones?

The problem here is that you do `formatAddress(col.value)`, but sometimes `col.value` is an ENS name, and `formatAddress` truncates and adds `...` regardless of whether its a hex string or ENS name. So it seems you can just make that method more robust and do something like this:
- if the ethers.js `isHexString` method returns true, it's an address so use the current logic
- otherwise, it's not an address so do nothing and return the full ENS name.

Then I think we can delete the lookupOrFormat methods that we no longer use and then we should be all set

garyghayrat: Addressed it here https://github.com/ScopeLift/umbra-protocol/pull/381/commits/8cc736ac2fcdf389065d9a482f1ed854f2a04bb4

frontend/src/components/AccountReceiveTable.vue
garyghayrat: Yea, I like the golden link style. Truncation of ens names should be good now.

frontend/src/utils/address.ts
```diff
@@ -86,18 +87,13 @@ export const toAddress = utils.toAddress;
...
-export const formatAddresses = (addresses: string[]) => addresses.map(formatAddress);
+export const formatAddresses = (addresses: string[]) => addresses.map(formatNameOrAddress);
 
 export const lookupAddresses = async (addresses: string[], provider: Provider) => {
   const promises = addresses.map((address) => lookupAddress(address, provider));
   return Promise.all(promises);
 };
 
-export const lookupOrFormatAddresses = async (addresses: string[], provider: Provider) => {
-  const promises = addresses.map((address) => lookupOrFormatAddress(address, provider));
```
mds1, line 91: It seems we also don't use `lookupOrFormatAddress` and can delete that too

frontend/src/utils/utils.ts
```diff
@@ -1,14 +1,17 @@
 import { supportedChains, TokenInfo } from 'src/components/models';
 import { BigNumber, BigNumberish, hexValue, parseUnits, formatUnits } from './ethers';
-
 /**
  * @notice Generates the Etherscan URL based on the given `txHash` or `address and `chainId`
  */
 export const getEtherscanUrl = (txHashOrAddress: string, chainId: number) => {
-  const group = txHashOrAddress.length === 42 ? 'address' : 'tx';
+  const group = txHashOrAddress.length === 42 ? 'address' : txHashOrAddress.length === 66 ? 'tx' : 'ens';
```
mds1, line 7: This isn't exactly correct, since you can have a 42 or 66 character ENS name. This should use `isHexString` to identify a txHash/address or the presence of a `.` means it must be an ENS name

frontend/src/utils/address.ts
```diff
@@ -5,13 +5,14 @@
...
-import { getAddress, Web3Provider } from 'src/utils/ethers';
+import { getAddress, Web3Provider, isHexString } from 'src/utils/ethers';
 import { getChainById } from 'src/utils/utils';
 import { i18n } from '../boot/i18n';
 // ================================================== Address Helpers ==================================================
 
 // Returns an address with the following format: 0x1234...abcd
-export const formatAddress = (address: string) => `${address.slice(0, 6)}...${address.slice(38)}`;
+export const formatNameOrAddress = (address: string) =>
+  isHexString(address) ? `${address.slice(0, 6)}...${address.slice(38)}` : address;
```
mds1, line 15: Nit on naming/style:

```suggestion
export const formatNameOrAddress = (nameOrAddress: string) => {
  return isHexString(nameOrAddress) ? `${nameOrAddress.slice(0, 6)}...${nameOrAddress.slice(38)}` : nameOrAddress;
}
```

frontend/src/utils/address.ts
garyghayrat: deleted!

frontend/src/utils/utils.ts
garyghayrat: How's this https://github.com/ScopeLift/umbra-protocol/blob/5b7e7358a76fd8794ef53f8ea14f7c34accf0c05/frontend/src/utils/utils.ts#L7

mds1: LGTM!

garyghayrat: You're amazing, thanks!!

apbendi (changes requested):
This works well, and the code looks good too, but I think we want to change this a bit to avoid confusion by users.

The intent of #135 is to have the word "Withdrawn" be clickable after the withdraw occurs, and to have the link go to the Etherscan record of the withdraw txn itself. In the case of base asset (ETH, OETH, AETH, or MATIC), this will be a normal transfer by the stealth address. In the case of a token, this will be a meta-tx transaction sent through the Umbra contract by a relayer.

In both cases, it's a bit difficult to actually find this transaction quickly. We would need to use an indexing service like The Graph and do a little query-fu. As a result, a simple first pass approach is to link to the *stealth* address on Etherscan rather than the transaction itself. This is helpful in the base asset case, because the stealth address will have a single transaction (the withdrawal) and it would be easy for the user to see that. It is not helpful at all for Token withdrawals because in that case, the stealth address in question will never have been seen on chain, except as calldata. As such, I think it's actually *more* confusing to link to the stealth address in the token withdraw case than simply not including anything.

Based on all this, I think we should make two changes to this PR:

1. Only link out to the stealth address if the transaction in question used the base asset
2. Change the link to be on the word "Withdrawn" instead of on the stealth address in the table. Keep the style as-is, "Withdrawn" should just be clickable. A nice to have the little square/arrow indicator appear on hover if it's clickable, as seen here:

Screen Shot 2022-08-29 at 1 29 35 PM

garyghayrat:
Screenshot_2022-08-31_15-24-06
With the latest PR, native token withdrawals looks like the row on top, and ERC20 tokens like the one on the bottom.

mds1 (approved):
LGTM, CI failure is unrelated, looks like its due to infra providers being down during the arby upgrade yesterday. Will defer to @apbendi to review / make sure UX is good before merging

image

PR #383, opened 2022-08-22
Streamline periphery fork tests

garyghayrat:
Changes: 

- Use `vm.createSelectFork` cheatcode.
- Update foundry.toml
- Update dependencies including solmate

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "contracts-periphery/lib/openzeppelin-contracts"]
 	path = contracts-periphery/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/OpenZeppelin-contracts
-[submodule "contracts-periphery/lib/solmate"]
-	path = contracts-periphery/lib/solmate
-	url = https://github.com/Rari-capital/solmate
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "contracts-periphery/lib/forge-std"]
 	path = contracts-periphery/lib/forge-std
 	url = https://github.com/brockelmore/forge-std
+[submodule "contracts-periphery/lib/solmate"]
+	path = contracts-periphery/lib/solmate
+	url = https://github.com/transmissions11/solmate
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "contracts-periphery/lib/solmate"]
 	path = contracts-periphery/lib/solmate
 	url = https://github.com/transmissions11/solmate
+[submodule "contracts-periphery/lib/forge-std"]
+	path = contracts-periphery/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,6 @@
 [submodule "contracts-periphery/lib/openzeppelin-contracts"]
 	path = contracts-periphery/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/OpenZeppelin-contracts
-[submodule "contracts-periphery/lib/forge-std"]
-	path = contracts-periphery/lib/forge-std
-	url = https://github.com/brockelmore/forge-std
 [submodule "contracts-periphery/lib/solmate"]
 	path = contracts-periphery/lib/solmate
 	url = https://github.com/transmissions11/solmate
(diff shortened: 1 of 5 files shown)


contracts-periphery/test/utils/DSTestPlus.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.13;
 
 import "forge-std/Test.sol";
+import "forge-std/Vm.sol";
```
mds1, line 5: I don't think this import is needed since `Test.sol` has a `Vm` instance and we inherit from `Test`?

contracts-periphery/Makefile
```diff
@@ -12,11 +12,11 @@ forkBlock = 14800081
...
 test :; make test-contracts test-gas
 test-contracts :; make test-batch-send test-uniswap-hook
 test-batch-send :; forge test --match-path test/UmbraBatchSend.t.sol
-test-uniswap-hook :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path test/UniswapWithdrawHook.fork.1.t.sol
+test-uniswap-hook :; forge test --match-path test/UniswapWithdrawHook.fork.1.t.sol
 test-gas :; make test-gas-batch-send test-gas-uniswap-hook
 test-gas-batch-send :; forge test --match-path test/gas/UmbraBatchSend.gas.t.sol
-test-gas-uniswap-hook :; forge test -f ${ETH_RPC_URL} --fork-block-number ${forkBlock} --match-path test/gas/UniswapWithdrawHook.fork.1.gas.t.sol
\ No newline at end of file
+test-gas-uniswap-hook :; forge test --match-path test/gas/UniswapWithdrawHook.fork.1.gas.t.sol
```
mds1, line 22: Should we clean up this makefile a bit? We probably just need `make test` and `make test-gas` (and `make snapshot-gas`) now, since the new fork architecture means any specific test file can just be run with the standard `forge test --match-path` flag so we prob don't need special commands in the makefile for each

contracts-periphery/test/utils/DSTestPlus.sol
garyghayrat: You're right!

contracts-periphery/Makefile
garyghayrat: It's def cleeean https://github.com/ScopeLift/umbra-protocol/blob/6f3caa0080de5df43fd12b1afc35f2444539df69/contracts-periphery/Makefile#L15-L17

mds1: Love it 😍

.gitmodules
```diff
@@ -1,12 +1,12 @@
...
 	url = https://github.com/dapphub/ds-test
-[submodule "contracts-periphery/lib/forge-std"]
-	path = contracts-periphery/lib/forge-std
-	url = https://github.com/brockelmore/forge-std
 [submodule "contracts-periphery/lib/openzeppelin-contracts"]
 	path = contracts-periphery/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/OpenZeppelin-contracts
+[submodule "contracts-periphery/lib/forge-std"]
+	path = contracts-periphery/lib/forge-std
+	url = https://github.com/brockelmore/forge-std
```
mds1, line 9: We can also update this 
```suggestion
	url = https://github.com/foundry-rs/forge-std
```

garyghayrat: Updated!

mds1 (approved):

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
```diff
@@ -34,6 +35,9 @@ contract UniswapWithdrawHookTest is DeployUmbraTest {
 
   function setUp() public override {
     super.setUp();
+    uint256 mainnetForkBlock = 15181633;
```
apbendi, line 38: Is there anything special about this block number that would warrant a comment explaining why it was chosen?

apbendi (approved):
Will need to be rebased after #351 is merged, but otherwise LGTM. Thanks @garyghayrat.

contracts-periphery/test/UniswapWithdrawHook.fork.1.t.sol
garyghayrat: Nope, it was just the latest block at the time I was testing this.

PR #390, opened 2022-09-11
Send native token max without dust

### LINKED ISSUE #378: "send max" robustness improvements
> one issue we'd run into [when using the "Max" button on the send page] is with the native token: Gas price might change from the time you fill in the max to the time you send your transaction, resulting in either (1) an insufficient balance error if gas price increases, or (2) dust leftover if gas price decreases. So for the native token we'll need a flag that indicates the user clicked the "Max" button and recompute the the actual max at the time of the send transaction.
> 
> We also might want to throw an error if the amount changes by some threshold (+/- 1-2% maybe?) to prevent the user from sending a lot less than they expect if gas prices quickly change, but that might not be worth the added complexity for now.

_originally posted in https://github.com/ScopeLift/umbra-protocol/pull/375#discussion_r942698351_

davidlaprade:
Fixes #378

umbra-send-max-improvements

### DIFF
--- a/frontend/src/i18n/locales/en-US.json
+++ b/frontend/src/i18n/locales/en-US.json
@@ -68,6 +68,7 @@
     "question-circle-warning": "Don't use this feature unless you know what you're doing.",
     "learn-more": "Learn more",
     "select-token": "Select token to send",
+    "slippage-exceeded": "Actual send amount would materially differ from the estimate due to changing gas prices. Please try again.",
     "token": "Token",
     "amount": "Amount to send",
     "summary": "Summary",
--- a/frontend/src/i18n/locales/en-US.json
+++ b/frontend/src/i18n/locales/en-US.json
@@ -77,6 +77,7 @@
     "fee-explain": "Transactions on %{chainName} are very cheap, so a small fee is charged to deter spamming the protocol.",
     "total": "Total",
     "max": "Max",
+    "max-native-less-than-toll": "Cannot send max. Balance less than toll.",
     "copy-payment-link": "Copy payment link",
     "enter-an-amount": "Please enter an amount",
     "enter-a-recipient": "Please enter a recipient",
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -495,17 +495,12 @@ function useSendForm() {
             gasPrice: _sendMaxGasPrice,
             gasLimit: _sendMaxGasLimit,
             ethToSend: balanceLessGasCosts,
-          } = await umbraUtils.getEthSweepGasInfo(
-            userAddress.value!,
-            _toAddress,
-            provider.value!,
-            {
-              // We override the gasLimit here because we are sending to an
-              // address that has never been seen before, which increases gas
-              // costs and is not accounted for by getEthSweepGasInfo.
-              gasLimit: _estimatedNativeSendGasLimit,
-            }
-          );
+          } = await umbraUtils.getEthSweepGasInfo(userAddress.value!, _toAddress, provider.value!, {
+            // We override the gasLimit here because we are sending to an
+            // address that has never been seen before, which increases gas
+            // costs and is not accounted for by getEthSweepGasInfo.
+            gasLimit: _estimatedNativeSendGasLimit,
+          });
           sendMaxGasPrice = _sendMaxGasPrice;
           sendMaxGasLimit = _sendMaxGasLimit;
           tokenAmount = balanceLessGasCosts.sub(toll.value);
@@ -513,7 +508,6 @@ function useSendForm() {
           tokenAmount = currentBalance;
         }
       }
-
 
       if (tokenAddress === NATIVE_TOKEN.value.address) {
         // Throw if the tokenAmount differs from humanAmount.value by too much.
@@ -594,17 +588,12 @@ function useSendForm() {
       if (!userAddress.value || !provider.value) throw new Error(tc('Send.wallet-not-connected'));
       const fromAddress = userAddress.value;
       const recipientAddress = await toAddress(recipientId.value, provider.value);
-      const { ethToSend } = await umbraUtils.getEthSweepGasInfo(
-        fromAddress,
-        recipientAddress,
-        provider.value,
-        {
-          // We override the gasLimit here because we are sending to an
-          // address that has never been seen before, which increases gas
-          // costs and is not accounted for by getEthSweepGasInfo.
-          gasLimit: await estimateNativeSendGasLimit(),
-        }
-      );
+      const { ethToSend } = await umbraUtils.getEthSweepGasInfo(fromAddress, recipientAddress, provider.value, {
+        // We override the gasLimit here because we are sending to an
+        // address that has never been seen before, which increases gas
+        // costs and is not accounted for by getEthSweepGasInfo.
+        gasLimit: await estimateNativeSendGasLimit(),
+      });
       const sendAmount = ethToSend.sub(toll.value);
       if (sendAmount.lt('0')) throw new Error(tc('Send.max-native-less-than-toll'));
       humanAmount.value = formatUnits(sendAmount, token.value.decimals);
@@ -616,21 +605,19 @@ function useSendForm() {
 
   // Get an accurate estimate of the amount of gas needed to perform a native send.
   async function estimateNativeSendGasLimit() {
-    return (
-      await umbra.value!.umbraContract.estimateGas.sendEth(
-        // We will be sending to an address that has never been seen before which substantially
-        // increases gas costs on some networks (e.g. by 25k on mainnet). To ensure this cost is
-        // included in our gas limit estimate, we estimate using a `to` address that is randomly
-        // generated (and thus likely to have never been seen before).
-        new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0').slice(0, 42),
-        // The toll needs to be correct, otherwise the tx would revert.
-        toll.value,
-        // Fake values just to get a reasonable estimate.
-        new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0'), // pubKeyXCoordinate
-        new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0'), // ciphertext
-        // Value doesn't matter, it just needs to be more than the toll else the tx would revert.
-        { value: toll.value.add('1') }
-      )
+    return await umbra.value!.umbraContract.estimateGas.sendEth(
+      // We will be sending to an address that has never been seen before which substantially
+      // increases gas costs on some networks (e.g. by 25k on mainnet). To ensure this cost is
+      // included in our gas limit estimate, we estimate using a `to` address that is randomly
+      // generated (and thus likely to have never been seen before).
+      new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0').slice(0, 42),
+      // The toll needs to be correct, otherwise the tx would revert.
+      toll.value,
+      // Fake values just to get a reasonable estimate.
+      new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0'), // pubKeyXCoordinate
+      new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0'), // ciphertext
+      // Value doesn't matter, it just needs to be more than the toll else the tx would revert.
+      { value: toll.value.add('1') }
     );
   }
 
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -481,8 +481,6 @@ function useSendForm() {
       const currentBalance = balances.value[tokenAddress];
       const sendingNativeToken = tokenAddress === NATIVE_TOKEN.value.address;
       let tokenAmount = parseUnits(humanAmount.value, decimals);
-      // Needed for transactions that are going to sendMax.
-      let preCalculatedGasPrice;
       // Refresh the tokenAmount if the sendMax flag is set.
       if (sendMax.value) {
         if (sendingNativeToken) {
@@ -547,7 +545,6 @@ function useSendForm() {
       // Send with Umbra
       const { tx } = await umbra.value.send(signer.value, tokenAddress, tokenAmount, recipientId.value, {
         advanced: shouldUseNormalPubKey.value,
-        gasPrice: sendMax.value ? preCalculatedGasPrice : undefined,
       });
       void txNotify(tx.hash, ethersProvider);
       await tx.wait();
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -605,7 +605,9 @@ function useSendForm() {
           gasLimit: await estimateNativeSendGasLimit(),
         }
       );
-      humanAmount.value = formatUnits(ethToSend.sub(toll.value), token.value.decimals);
+      const sendAmount = ethToSend.sub(toll.value);
+      if (sendAmount.lt('0')) throw new Error(tc('Send.max-native-less-than-toll'));
+      humanAmount.value = formatUnits(sendAmount, token.value.decimals);
     } else {
       const tokenBalance = balances.value[token.value.address];
       humanAmount.value = formatUnits(tokenBalance.toString(), token.value.decimals);
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -149,8 +149,9 @@
         :disable="isSending"
         placeholder="0"
         :appendButtonDisable="!recipientId || !isValidRecipientId"
-        :appendButtonLabel="token && NATIVE_TOKEN && token.address !== NATIVE_TOKEN.address ? $t('Send.max') : ''"
+        :appendButtonLabel="$t('Send.max')"
         @click="setHumanAmountMax"
+        @input="() => (sendMax = false)"
         lazy-rules
         :rules="isValidTokenAmount"
         ref="humanAmountBaseInputRef"
@@ -248,7 +249,7 @@
 // --- External imports ---
 import { computed, defineComponent, onMounted, ref, watch } from 'vue';
 import { QForm, QInput, QSelect } from 'quasar';
-import { utils as umbraUtils } from '@umbracash/umbra-js';
+import { RandomNumber, utils as umbraUtils } from '@umbracash/umbra-js';
 // --- Components ---
 import BaseInput from 'components/BaseInput.vue';
 import BaseSelect from 'components/BaseSelect.vue';
@@ -312,6 +313,7 @@ function useSendForm() {
   const isValidForm = ref(false);
   const isValidRecipientId = ref(true); // for showing/hiding bottom space (error message div) under input field
   const toll = ref<BigNumber>(Zero);
+  const sendMax = ref(false);
 
   // Computed form parameters.
   const showAdvancedWarning = computed(() => advancedAcknowledged.value === false && useNormalPubKey.value === true);
@@ -363,6 +365,12 @@ function useSendForm() {
       // Reset acknowledgement if user changes the public key type.
       if (!useNormalPubKey) {
         advancedAcknowledged.value = false;
+      }
+
+      // Switch off the sendMax flag and clear value if we change tokens.
+      if (tokenValue != (_prevTokenValue || prevNativeTokenValue) && sendMax.value) {
+        sendMax.value = false;
+        humanAmount.value = '0.0';
       }
 
       // Perform minimal required validation based on what changed.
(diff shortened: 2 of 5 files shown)


frontend/src/i18n/locales/en-us.json
```diff
@@ -62,6 +62,7 @@
     "question-circle-warning": "Don't use this feature unless you know what you're doing.",
     "learn-more": "Learn more",
     "select-token": "Select token to send",
+    "slippage-exceeded": "Actual send amount would materially differ from estimate. Please retry.",
```
mds1, line 65: Nit, feels a little more clear, lmk what you think / feel free to resolve this comment yourself. Also @garyghayrat can you give us the translation for whatever @davidlaprade goes with? 
```suggestion
    "slippage-exceeded": "Actual send amount would significantly differ from estimate due to changing gas prices. Please try again.",
```

frontend/src/pages/AccountSend.vue
```diff
@@ -179,6 +180,7 @@ import useWalletStore from 'src/store/wallet';
 // --- Other ---
 import { txNotify } from 'src/utils/alerts';
 import { BigNumber, Contract, getAddress, MaxUint256, parseUnits, formatUnits, Zero } from 'src/utils/ethers';
+import { Wallet } from 'ethers';
```
mds1, line 183: We keep all ethers imports in `src/utils/ethers` since I _think_ it tree-shakes better if you don't import from the root `ethers` package (but that might not be true/outdated), and it just makes it clearer to track the set of packages we're dependent on. All that to say let's move `Wallet` to be exported in `src/utils/ethers` and import it from there
```suggestion
import { BigNumber, Contract, getAddress, MaxUint256, parseUnits, formatUnits, Zero, Wallet } from 'src/utils/ethers';
```

I think `export { Wallet } from '@ethersproject/wallet';` is all you need to add that that file

frontend/src/pages/AccountSend.vue
```diff
@@ -414,18 +448,55 @@ function useSendForm() {
...
-    const tokenBalance = balances.value[token.value.address];
-    humanAmount.value = formatUnits(tokenBalance.toString(), token.value.decimals);
-    return tokenBalance.toString();
+    return (
+      await umbra.value!.umbraContract.estimateGas.sendEth(
+        // We will be sending to an address that has never been seen before, which increases gas
+        // costs by 25k. To ensure this cost is included in our gas limit estimate, we estimate
+        // using a `to` address that is randomly generated (and thus likely to have never been seen
+        // before).
+        Wallet.createRandom().address,
```
mds1, line 485: We can avoid introducing dependency on the wallet class with ` new RandomNumber().asHex.replace(/0/g, 'f').replace(/^./, '0').slice(0, 42)`

mds1 (commented):
Tested out on polygon and this worked great!

frontend/src/i18n/locales/en-us.json
mds1: Eh reading this again now I get why you used "materially". My motivation for changing it was that it seemed like a less common word for users who aren't native english speakers, but I don't know if that's actually true

frontend/src/pages/AccountSend.vue
davidlaprade: Going to just remove the Wallet import in favor of the `RandomNumber` approach below

frontend/src/i18n/locales/en-us.json
davidlaprade: I think "material" is pretty common in finance-speak, but I have no idea how recognizable it is to non-native speakers. Curious what Gary thinks. For now I'll leave it but take the other suggestions, thanks!

Also, yeah, @garyghayrat if you could give me the translation for Chinese, that'd be great!

davidlaprade:
Test failure does not appear to be related to this PR. It's also failing on `master`

image

frontend/src/i18n/locales/en-us.json
garyghayrat: I like matt's suggestion overall. Like the simpler 'significantly' and the explanation of `due to changing gas prices.` Just added the translation as well.

garyghayrat:
Mainnet max estimate seems a bit high. Not enough eth left to pay for market gas price.
image

davidlaprade:
> Mainnet max estimate seems a bit high

Interesting. I'm really glad you checked on mainnet! I actually hadn't. (I needed to republish my keys and I was too lazy/cheap to do it before now. 😅) But since I fixed my mainnet setup, I'm actually seeing the opposite -- estimates are too low:

image

What fee estimate are you seeing in metamask on mainnet @garyghayrat? This is what I'm currently seeing (on 9/12/22 ~3:45pm ET):

image

I added some logging and it seems like our mainnet estimates are actually consistently 20% too low, even with the 10% scale factor applied. I'm not sure why that is.

Metamask's estimated gas limit is ~64k for a mainnet sendETH:

image

And ours is 70671 -- exactly 10% higher than metamask's (as expected, given the scale factor). So the issue seems to be with gas prices.

Sure enough, the MM max gas fee is ~11% higher than what we're getting from our provider. And based on https://www.blocknative.com/gas-estimator theirs is correct. I'll need to look into this

frontend/src/i18n/locales/en-us.json
mds1: Looks like we still have "materially"? I'm ok with that if that's the decision, but from the above thought we were going with "significantly"

frontend/src/pages/AccountSend.vue
```diff
@@ -414,18 +455,55 @@ function useSendForm() {
...
+
     if (NATIVE_TOKEN.value?.address === token.value?.address) {
       if (!userAddress.value || !provider.value) throw new Error(vm.$i18n.tc('Send.wallet-not-connected'));
       const fromAddress = userAddress.value;
       const recipientAddress = await toAddress(recipientId.value, provider.value);
-      const { ethToSend } = await umbraUtils.getEthSweepGasInfo(fromAddress, recipientAddress, provider.value);
-      humanAmount.value = formatUnits(ethToSend, token.value.decimals);
-      return ethToSend;
+      const { ethToSend } = await umbraUtils.getEthSweepGasInfo(fromAddress, recipientAddress, provider.value, {
+        gasLimit: await estimateNativeSendGasLimit(),
```
mds1, line 465: Can we add a comment about why we need to pass a gas limit in here? I know we discussed it and the reason is slipping my mind right now

apbendi (commented):
Hey @davidlaprade, thanks for trying to tackle this. I did some QA testing, and here's what I found:

* Attempting to test this on Optimism and Arbitrum, I get the following error after approving the Transaction:
* 
Screen Shot 2022-09-26 at 1 55 16 PM

Screen Shot 2022-09-26 at 1 55 53 PM

* Testing it on Mainnet, I'm unable to approve the transaction, with M*taM*sk giving me the following message:

Screen Shot 2022-09-26 at 2 04 11 PM

* It worked perfectly on Polygon

I hate to say it, but at this point, I'm starting to wonder if this feature is worth the trouble for the native asset. What do you think @davidlaprade @mds1?

frontend/src/pages/AccountSend.vue
```diff
@@ -363,17 +369,51 @@ function useSendForm() {
...
+          // Add a 5% buffer to the gasPrice to avoid: "max fee per gas less than block base fee"
+          preCalculatedGasPrice = (await provider.value!.getGasPrice()).mul('105').div('100');
+
+          // Get current balance less gas costs.
+          const { ethToSend: balanceLessGasCosts } = await umbraUtils.getEthSweepGasInfo(
+            userAddress.value!,
+            await toAddress(recipientId.value, provider.value!),
+            provider.value!,
+            {
+              gasLimit: await estimateNativeSendGasLimit(),
```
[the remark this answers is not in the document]
davidlaprade: Good call, done in 2cafa3c

frontend/src/pages/AccountSend.vue
davidlaprade: Haha, if I can just remember myself...

davidlaprade: Ah, figured it out! e0ceb38

davidlaprade:
@apbendi I was just able to reproduce the error on Arbitrum

image

I ultimately got it to work ([send tx here](https://arbiscan.io/tx/0xd2c849383a4a544d00729b9dfe925d70e0f31f50488579403cc83c64ba60a6fa)) by bumping the native send gas limit but tbh I'm not sure why it was a problem to begin with. The gas used by the tx (372,580) was well under the original estimate (461,905). It seems that this might be an issue with metamask? Unclear. But I bumped the scale factor `estimateNativeSendGasLimit` for arbitrum by an additional 5% (10% total) and that seemed to fix the issue.

I was also able to reproduce on optimism. Optimism fees are just weird, so there was a real fix needed here. I have to get more OETH before I can test that the fix works though.

I wasn't able to reproduce the issue on mainnet. I was able to [drain an account](https://etherscan.io/tx/0x7776a90b2b9dae646813344b603a5273cb0a53577ebe719095d9f6d74792a21c) without issue. Could you re-confirm you have this problem?

apbendi:
Hey @davidlaprade, I gave this another look and tested out both Arbitrum and Mainnet. The "allowance" errors on Arbitrum are gone, however now both networks show this message in MetaMask:

Screen Shot 2022-11-21 at 9 31 28 PM

This is the error I see when MetaMask is on "Market", which is the default. If I switch it to "Low", I'm able to confirm the transaction. So it seems maybe the price estimate for gas is just a bit lower than what MetaMask ends up using. Is there a way we can bump it up?

davidlaprade:
@apbendi I just rebased on top of master and retried max sends on mainnet and arbitrum, per:

> I gave this another look and tested out both Arbitrum and Mainnet. The "allowance" errors on Arbitrum are gone, however now both networks show this message in MetaMask

I didn't get any errors. Native token max sends worked without issue on both networks for me.

> This is the error I see when MetaMask is on "Market", which is the default. If I switch it to "Low", I'm able to confirm the transaction

This comment makes me wonder what your MetaMask settings are. I can't even see Market vs Low options unless I switch on an experimental feature (see screenshot). But actually even with this feature switched on I still had an arbitrum max send work without issue.

At this point I think it'd be helpful to hop on a call so I can see what you're seeing.

image

frontend/src/i18n/locales/en-US.json
```diff
@@ -68,6 +68,7 @@
     "question-circle-warning": "Don't use this feature unless you know what you're doing.",
     "learn-more": "Learn more",
     "select-token": "Select token to send",
+    "slippage-exceeded": "Actual send amount would materially differ from the estimate due to changing gas prices. Please try again.",
```
davidlaprade, line 71: Coming back to this, I still think "materially" is what we want and "significantly" is incorrect/confusing. We throw if the final amount is more than 2% off of the projected amount. For a $10 transaction, that's an error of $0.20, an insignificant amount, but it might still matter to someone (i.e. be "material")

frontend/src/i18n/locales/en-US.json
```diff
@@ -68,6 +68,7 @@
     "question-circle-warning": "Don't use this feature unless you know what you're doing.",
```
apbendi, line 68: Still getting one error when I attempt this on Optimism :cry: 

image

umbra-js/src/utils/utils.ts
```diff
@@ -350,6 +350,10 @@ export async function getEthSweepGasInfo(
     provider.getCode(to),
     provider.getNetwork(),
     provider.getBalance(from),
+    // getGasPrice is officially deprecated in favor of getFeeData, as the
+    // former uses pre-1559 gas pricing. But we use it anyway because it makes
+    // it easier to calculate a price to sweep the account (1599 factors in
+    // refunds that we don't care about).
```
mds1, line 356: I think it's ok to delete this comment, or just simplify it to say "We use `getGasPrice` instead of `getFeeData` because we don't care about refunds here, so using type 0 transactions makes it easier to calculate a price to sweep the account, and 1559 is not supported on some chains". Mainly just commenting that type 0/1 transactions (EIP-1559 txs are type 2) aren't technically deprecated, and are still a valid alternative (just for most people, 1559 is better). Also some networks we support might not support `getFeData`/1559, e.g. I don't think Arbitrum does nor Optimisim (currently, Bedrock will though)

frontend/src/pages/AccountSend.vue
```diff
@@ -365,6 +367,12 @@ function useSendForm() {
         advancedAcknowledged.value = false;
       }
 
+      // Switch off the sendMax flag and clear value if we change tokens.
+      if (tokenValue != (_prevTokenValue || prevNativeTokenValue) && sendMax.value) {
```
mds1, line 371: Let's compare the address field specifically to be a bit more robust against surrounding code changes, since this compares by reference by we just care about the values. Also in general recommend always using `!==` to avoid JS type casting for you

image

frontend/src/pages/AccountSend.vue
```diff
@@ -469,16 +477,60 @@ function useSendForm() {
...
+      let preCalculatedGasPrice;
+      // Refresh the tokenAmount if the sendMax flag is set.
+      if (sendMax.value) {
+        if (sendingNativeToken) {
+          const [_getGasPrice, _toAddress, _estimatedNativeSendGasLimit] = await Promise.all([
+            provider.value!.getGasPrice(),
+            toAddress(recipientId.value, provider.value!),
+            estimateNativeSendGasLimit(),
+          ]);
+          // Add a 5% buffer to the gasPrice to avoid: "max fee per gas less than block base fee"
```
mds1, line 494: Lots of comments seem L1-specific, such as "max fee per gas less than block base fee" and (paraphrased) "sending to a new address costs 25k extra gas". Maybe worth tweaking them

frontend/src/pages/AccountSend.vue
```diff
@@ -469,16 +477,60 @@ function useSendForm() {
...
+          const { ethToSend: balanceLessGasCosts } = await umbraUtils.getEthSweepGasInfo(
+            userAddress.value!,
+            _toAddress,
+            provider.value!,
+            {
+              // We override the gasLimit here because we are sending to an
+              // address that has never been seen before, which increases gas
+              // costs and is not accounted for by getEthSweepGasInfo.
+              gasLimit: _estimatedNativeSendGasLimit,
+              gasPrice: preCalculatedGasPrice,
```
mds1, line 506: Why do we need to fetch and use `preCalculatedGasPrice` instead of just relying on `getEthSweepGasInfo` to do it? (RR since it's a question I guess)

frontend/src/pages/AccountSend.vue
```diff
@@ -520,18 +573,63 @@ function useSendForm() {
...
+      const tokenBalance = balances.value[token.value.address];
+      humanAmount.value = formatUnits(tokenBalance.toString(), token.value.decimals);
+    }
+  }
+
+  // Get an accurate estimate of the amount of gas needed to perform a native send.
+  async function estimateNativeSendGasLimit() {
+    // Increase estimate to give us some wiggle room if network conditions are volatile.
+    // These have each been configured on a case-by-case basis; do not change
+    // unless you have tested a native token Umbra send on the network with the new value.
```
mds1, line 604: How did we come up with the current scale factors / what's the process for testing here? A single send or two isn't really sufficient since maybe you tested when gas prices were stable. I feel like we're more eyeballing these, which is fine for now

davidlaprade: Yeah, just eyeballing

frontend/src/pages/AccountSend.vue
davidlaprade: Ah yeah, good catch 👍

frontend/src/pages/AccountSend.vue
davidlaprade: Trying to remember, but seems we do this because of the "max fee per gas less than block base fee" error.

But if that's right, the padding we're doing here with `preCalculatedGasPrice` doesn't feel like the correct solution. It seems like the right approach would be to edit `getETHSweepGasInfo` so that it never returns a `gasPrice` less than the expected base fee (if relevant). And then it wouldn't be necessary to pass in a gasPrice to the function as an override.

frontend/src/pages/AccountSend.vue
davidlaprade: This is no longer necessary and the code has been removed. Resolving

frontend/src/pages/AccountSend.vue
davidlaprade: This code was removed. Resolving

frontend/src/pages/AccountSend.vue
davidlaprade: Comments tweaked!

davidlaprade:
Alright @apbendi @mds1, I think I _finally_ figured out the remaining issues we were having. I just rebased on master (again) and I think this is ready for re-review.

I have now tested:
* native token max sends (all 4 supported networks)
* native token partial sends (all 4 supported networks)
* ERC20 token max sends (polygon)
* ERC20 token partial sends (polygon)
* receive native token (all 4 supported networks)
* receive ERC20 (polygon)
* native token max sends on the [release preview](https://deploy-preview-390--jolly-shaw-20fe62.netlify.app/send) (all L2s)

Everything worked!

cc @garyghayrat as there is one new string to translate

umbra-js/src/utils/utils.ts
```diff
@@ -360,10 +360,16 @@ export async function getEthSweepGasInfo(
   const gasLimitOf21k = [1, 4, 5, 10, 137, 1337]; // networks where ETH sends cost 21000 gas
   const ignoreGasPriceOverride = [10, 42161]; // to maximize ETH sweeps, ignore uer-specified gasPrice overrides
 
-  const [toAddressCode, network, fromBalance, providerGasPrice] = await Promise.all([
+  const [toAddressCode, network, fromBalance, lastBlockData, providerGasPrice] = await Promise.all([
     provider.getCode(to),
     provider.getNetwork(),
     provider.getBalance(from),
+    provider.getBlock('latest'),
+    // getGasPrice is officially deprecated in favor of getFeeData but we use it anyway because:
```
mds1, line 368: nit, since this isn't actually deprecated
```suggestion
    // We use `getGasPrice` instead of `getFeeData` because:
```

umbra-js/src/utils/utils.ts
```diff
@@ -378,22 +384,51 @@ export async function getEthSweepGasInfo(
     : await provider.estimateGas({ gasPrice: 0, to, from, value: fromBalance });
 
   // Estimate the gas price, defaulting to the given one unless on a network where we want to use provider gas price
-  const gasPrice = ignoreGasPriceOverride.includes(chainId)
+  let gasPrice = ignoreGasPriceOverride.includes(chainId)
     ? providerGasPrice
     : BigNumber.from((await overrides.gasPrice) || providerGasPrice);
 
+  // On networks with EIP-1559 gas pricing, the provider will throw an error and refuse to submit
+  // the tx if the gas price is less than the block-specified base gas cost. The error is "max fee
```
mds1, line 392:
```suggestion
  // the tx if the gas price is less than the block-specified base fee. The error is "max fee
```

umbra-js/src/utils/utils.ts
```diff
@@ -378,22 +384,51 @@ export async function getEthSweepGasInfo(
...
+  // per gas less than block base fee". So we need to ensure that the low estimate we're using isn't
+  // *too* low.  Additionally, if the previous block exceeded the target size, the base fee will be
```
mds1, line 394:
```suggestion
  // *too* low. Additionally, if the previous block exceeded the target size, the base fee will be
```

umbra-js/src/utils/utils.ts
```diff
@@ -378,22 +384,51 @@ export async function getEthSweepGasInfo(
...
+  // increased by 12.5% for the next block, per:
+  //   https://ethereum.org/en/developers/docs/gas/#base-fee
+  // To be conservative, therefore, we assume a 12.5% increase will affect the base fee for the
+  // transaction we're about to send.
+  const baseGasFee = (lastBlockData?.baseFeePerGas || BigNumber.from('1')).mul('1125').div('1000');
+  if (gasPrice.lt(baseGasFee)) gasPrice = baseGasFee;
```
mds1, line 400: RR. I'm a bit confused here, isn't this saying that if the block had no base fee (i.e. EIP-1559 isn't supported) then use a base fee of 1.125 gwei? I think what we want is:

```suggestion
// Get basefee from block, or zero if EIP-1559 is not supported.
const baseFeePerGas = lastBlockData?.baseFeePerGas || BigNumber.from('0')
// If the gas price is smaller than the base fee, increase it to match the base fee
if (gasPrice.lt(baseGasFee)) gasPrice = baseGasFee;
// Add 12.5% to account for gas price increase
gasPrice = gasPrice.mul(1125).div(1000);
```

umbra-js/src/utils/utils.ts
```diff
@@ -378,22 +384,51 @@ export async function getEthSweepGasInfo(
...
     const gasPriceOracle = new Contract('0x420000000000000000000000000000000000000F', gasOracleAbi, provider);
-    const l1FeeInWei = await gasPriceOracle.getL1Fee(
+    let l1FeeInWei = await gasPriceOracle.getL1Fee(
       serializeTransaction({ to, value: fromBalance, data: '0x', gasLimit, gasPrice, nonce })
     );
-    txCost = txCost.add(<BigNumber>l1FeeInWei);
+
+    // We apply a 70% multiplier to the Optimism oracle quote, since it's frequently low
+    // by 45+% compared to actual L1 send fees.
+    l1FeeInWei = (l1FeeInWei as BigNumber).mul('170').div('100');
```
mds1, line 429: oof that's a lot of margin, not much we can do there I guess

mds1 (commented):
LGTM overall, a few nits on comments and one thing I'm confused about

apbendi (approved):
Tested on each network. Seems to be working great! This one took some doing. Thanks for seeing it through @davidlaprade

umbra-js/src/utils/utils.ts
davidlaprade: Ah, you're right. That should have been zero.

But I don't think we want to apply the 12.5% increase to the gasPrice if we're not using the base fee for the gasPrice. Really what I think we want is: if gasPrice < baseGasFee, then gasPrice = baseGasFee * 1.125. I'll make this change

davidlaprade: Fixed with 948cdd8adb84cf4453ac5409c70ca93d70195693

mds1: > The 12.5% increase is just applied to the base fee because we pessimistically assume that it will be increased in the next block. I

Yea, but `gasPrice` can also increase in the next block, so wouldn't we want to add margin there also? At the very least, it ensures the tx gets mined faster which is nicer UX

>  If the gasPrice that the provider predicts is above the pessimistic base fee, it should be safe to use without padding it also

What about networks without base fees? Gas price can increase in the next block still

garyghayrat (approved):

umbra-js/src/utils/utils.ts
davidlaprade: > gasPrice can also increase in the next block, so wouldn't we want to add margin there also?

Good point. And actually I'd forgotten that we are [already padding gas prices](https://github.com/ScopeLift/umbra-protocol/pull/390/files?diff=unified&w=0#diff-83ca539312493c7e5d6efd2715eef9cf3c42f8fea27cd423ea611c4a48ebbe5eR402-R412) globally by 5% or more depending on the chain right below this code. So for that reason I don't think we need to do more

davidlaprade: Tracking here: https://github.com/ScopeLift/umbra-protocol/issues/477

umbra-js/src/utils/utils.ts
davidlaprade: Resolving. This comment was updated accordingly

Issue #310, opened 2022-03-23
On withdraw modal, warn if withdrawing to a token contract


mds1:
We'd add this check here: https://github.com/ScopeLift/umbra-protocol/blob/a9ee536c46d7e28c990b536b03986ac860358a88/frontend/src/utils/address.ts#L99

Potential ways to implement:
- Have the backend maintain a static list of token addresses, and the frontend will use a new endpoint to ask if the destination address is a token
- The frontend can directly query the contract for "name", "symbol", "decimals", and  fields, and if all 3 exist it's likely a token. Alternatively just query for name and symbol which should also cover NFTs

I think I lean toward the second option to reduce reliance on the backend

mds1:
To start, perhaps we just only check against the address being withdrawn to simplify the logic

alexkeating:
I will pick this one up this week

alexkeating:
This should be resolved now

PR #396, opened 2022-10-03
Add ERC20 and ERC721 warnings for withdrawl

alexkeating:
### Issue
#310 

### Description

Added a check for an ERC721 and ERC20

[Screencast from 10-02-2022 11:58:42 PM.webm](https://user-images.githubusercontent.com/12705146/193499324-22e957f6-9d09-4d2a-9c86-4394953852fc.webm)


frontend/src/i18n/locales/zh-cn.json
```diff
@@ -306,6 +306,8 @@
       "it": "它",
       "resolves-to": "解析为",
       "is": "是",
+      "is-erc20": "一个 ERC20 代币",
+      "is-erc721": "一个 ERC721 代币",
```
alexkeating, line 310: @garyghayrat Are these translations okay?

garyghayrat: Yes!

mds1: One nit is to just add hyphens in both i18n files, e.g. "ERC-20" and "ERC-721"

frontend/src/utils/address.ts
```diff
@@ -131,6 +131,18 @@ export const isAddressSafe = async (name: string, userAddress: string, stealthAd
...
+  // Check if address is an ERC20
+  const desIsERC20 = await isERC20(destinationAddress, provider);
+  if (desIsERC20) {
+    reasons.push(`${i18n.tc('Utils.Address.it')} ${i18n.tc('Utils.Address.is')} ${i18n.tc('Utils.Address.is-erc20')}`);
+  }
+
+  // Check if address is an ERC721
+  const desIsERC721 = await isERC721(destinationAddress, provider);
+  if (!desIsERC20 && desIsERC721) {
+    reasons.push(`${i18n.tc('Utils.Address.it')} ${i18n.tc('Utils.Address.is')} ${i18n.tc('Utils.Address.is-erc721')}`);
```
mds1, line 143: Should we just push one `'Utils.Address.is-erc721'` string instead of composing? Seems that would scale better

frontend/src/utils/utils.ts
```diff
@@ -138,3 +139,28 @@ export const getGasPrice = async (gasPriceConfidence: TxPriceConfidence = 99): P
...
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+export const isERC721 = async (address: string, provider: Provider) => {
+  const tokenContract = new Contract(address, ERC721_ABI, provider);
+  const hasNamePromise = tokenContract.name();
+  const hasSymbolPromise = tokenContract.symbol();
```
mds1, line 159: We probably should look for a `tokenURI` to identify an NFT. I've seen contracts that have a `name` and `symbol` for other reasons, but aren't tokens

alexkeating: That is a good point. The tricky thing with tokenURI is it requires an id argument. If we include it should we pass in an id of 1?

frontend/src/utils/address.ts
alexkeating: I am changing to a single string, but this construction matches the way existing error messages are constructed for withdrawl

mds1: Ah, ok yea we might need to do some refactoring then if we ever support more languages, but thanks!

frontend/src/utils/utils.ts
mds1: Ahh, also a good point. Hm ok so wdyt of removing the ERC-20/721 distinction, looking for name and symbol, and if both exist say something like "this might be a token contract, such as ERC-20 or ERC-721". That way it's simpler, and other contracts that have name/symbol prob aren't ones you want to withdraw to anyway

alexkeating: I think that is a great idea, I will remove the distinction.

alexkeating:
Removed distinction

frontend/src/utils/address.ts
```diff
@@ -131,16 +131,10 @@ export const isAddressSafe = async (name: string, userAddress: string, stealthAd
   // Check if address has contributed to Gitcoin Grants
   // TODO
 
-  // Check if address is an ERC20
-  const desIsERC20 = await isERC20(destinationAddress, provider);
+  // Check if address might be an ERC-20 or ERC-721
+  const desIsERC20 = await isToken(destinationAddress, provider);
```
mds1, line 135: Nit
```suggestion
  const destIsToken = await isToken(destinationAddress, provider);
```

frontend/src/utils/constants.ts
```diff
@@ -19,7 +19,7 @@ export const ERC20_ABI = [
   'event Transfer(address indexed from, address indexed to, uint256 value)',
 ];
 
-export const ERC721_ABI = ['function name() view returns (string)', 'function symbol() view returns (string)'];
+export const TOKEN_ABI = ['function name() view returns (string)', 'function symbol() view returns (string)'];
```
mds1, line 22: We already have `ERC20_ABI` so we don't need to add `TOKEN_ABI`

mds1 (approved):
Two nits, feel free to self-resolve once addressed, otherwise LGTM. Pretty sure I tested this the other day and it worked as expected, so will wait for @apbendi as second reviewer before merging

frontend/src/utils/address.ts
alexkeating: Good catch

frontend/src/utils/constants.ts
alexkeating: True, I had that originally, but thought it might be clearer to separate out into a separate ABI even though it is captured by the ERC20. I will remove and change to the ERC-20 abi

apbendi (commented):
Great work @alexkeating, this is working great. Here's my only thought after testing it out: I feel like including it as a normal warning does not give sufficient gravity to what's actually happening. It is very likely that the user doesn't mean to do this, and will lose funds if he does.

Given that, I'm wondering: how hard would it be to show a *different* message if this criteria is met? This would supersede the other warning modal. If this is easy enough, let's do it in this PR. If it's not easy, let's pull it into an issue and merge this, as it's already an improvement.

The copy for the alternate modal could be something like:

### 🚨 Danger

It looks like you're trying to withdraw your funds to token contract. It is very likely this is not what you intend, and proceeding will likely result in a loss of funds. **Do not proceed unless you know exactly what you are doing.**

☐ I Know What I'm Doing

[ Withdraw Anyway ]

mds1:
That's a good point, all other checks are privacy related whereas this would be lost funds. FWIW this is how we spec'd it out in #310, so my suggestion would be to merge as-is since it's better than nothing like we currently have, then we can open a separate issue/PR to make this more prominent

frontend/src/i18n/locales/zh-cn.json
```diff
@@ -300,6 +300,7 @@
       "transaction-failed": "您的交易已失败(Your transaction has failed)"
     },
     "Address": {
+      "might-be-token": "它可能是一个代币合约，例如ERC-20或ERC-721",
```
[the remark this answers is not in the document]
garyghayrat: yes!

alexkeating:
This should be good, I also need to add the joi fix after rebaseing

frontend/package.json
```diff
@@ -24,12 +24,12 @@
     "@uniswap/token-lists": "^1.0.0-beta.19",
     "@unstoppabledomains/resolution": "8.3.1",
     "@vue/composition-api": "^0.6.4",
-    "@web3-onboard/coinbase": "^2.1.1",
-    "@web3-onboard/core": "^2.8.1",
-    "@web3-onboard/injected-wallets": "^2.2.0",
-    "@web3-onboard/ledger": "^2.2.1",
-    "@web3-onboard/trezor": "^2.2.1",
-    "@web3-onboard/walletconnect": "^2.1.1",
+    "@web3-onboard/coinbase": "^2.1.3",
```
mds1, line 27: _(arbitrary comment location to start a thread)_

Seems that when I cancel a withdraw to a token after the warning, the withdraw form is stuck in this loading state

image

frontend/src/i18n/locales/en-us.json
```diff
@@ -249,7 +249,10 @@
     "signer-or-provider-not-found": "Signer or provider not found",
     "fee-not-set": "Fee is not set",
     "invalid-chain-id": "Invalid chainID",
-    "receiver-tool-tip": "The stealth address which received these funds. It looks like an empty address, and it can't be tied to your account, but only you have the ability to generate its private key."
+    "receiver-tool-tip": "The stealth address which received these funds. It looks like an empty address, and it can't be tied to your account, but only you have the ability to generate its private key.",
+    "loss-warning": "It looks like you're trying to withdraw your funds to token contract. It is very likely this is not what you intend, and proceeding will likely result in a loss of funds. Do not proceed unless you know exactly what you are doing.",
```
mds1, line 253:
```suggestion
    "loss-warning": "It looks like you're trying to withdraw your funds to a token contract. It is very likely this is not what you intend, and proceeding will likely result in a loss of funds. Do not proceed unless you know exactly what you are doing.",
```

alexkeating: This should be resolved

frontend/package.json
alexkeating: This should be resolved

garyghayrat:
Getting this error when trying to connect wallet not sure why
image

garyghayrat (approved):
Two small nits, otherwise working great. Thank you!

apbendi (commented):
Hey @alexkeating, sorry for one more nit, but here's something I noticed in testing 🙈 

If the user clicks the checkbox and does the approval, the modal does not dismiss after the withdrawal. The warning modal, by comparison, dismisses before showing the subsequent fee modal.

alexkeating:
> Hey @alexkeating, sorry for one more nit, but here's something I noticed in testing see_no_evil
> 
> If the user clicks the checkbox and does the approval, the modal does not dismiss after the withdrawal. The warning modal, by comparison, dismisses before showing the subsequent fee modal.

Good catch pushed up a fix

apbendi (approved):
Looks and works great! Thanks for bearing with us and ironing out the details @alexkeating.

Issue #270, opened 2022-01-12
chore: yarn lint failing


thelostone-mc:
#### Description 

Looks like there are issues in compiler version being used.
This raises cause of the line https://github.com/ScopeLift/umbra-protocol/blob/master/contracts/.solhint.json#L6

`yarn lint`

![image](https://user-images.githubusercontent.com/5358146/149160135-654e04e1-3ba2-4a14-bef4-5ccea91b7729.png)

#### Possible Solutions

- Remove the compiler version (as it is not relevant) **OR**
- Increment  
  - the compiler version to 0.8.7
  - the contracts are updated to 0.8.7
- Do nothing and let the lint fail as contract upgrades aren't needed

alexkeating:
I can take this on

PR #401, opened 2022-10-06
goodbye rinkeby, hello goerli

mds1:

PR #402, opened 2022-10-09
Fix/fix linting

alexkeating:
Description

#270 Fix linting so it passes and add it as a step to the build. I had to restructure our lint rules so typescript rules did not apply to js files.


.github/workflows/ci.yaml
```diff
@@ -48,3 +48,27 @@ jobs:
...
         run: yarn test
 
+  lint:
+    runs-on: ubuntu-latest
+    env:
+      FOUNDRY_PROFILE: ci
+      ETH_RPC_URL: ${{ secrets.ETH_RPC_URL }}
+      INFURA_ID: ${{ secrets.INFURA_ID }}
+      ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+      POLYGONSCAN_API_KEY: ${{ secrets.POLYGONSCAN_API_KEY }}
```
mds1, line 58: Do we need these env vars for linting in CI? guessing we can remove them

.github/workflows/ci.yaml
```diff
@@ -48,3 +48,27 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - uses: volta-cli/action@v1
+
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
+        with:
+          version: nightly
```
mds1, line 69: We don't technically need this now but let's keep it since we may integrate [scopelint](https://github.com/ScopeLift/scopelint) soon (example usage in our [foundry template](https://github.com/ScopeLift/foundry-template)

.github/workflows/ci.yaml
```diff
@@ -48,3 +48,27 @@ jobs:
...
+
+      - run: yarn install --network-concurrency 1
```
mds1, line 71: Why do we need `--network-concurrency 1`?

mds1 (commented):
Thanks for doing this! LGTM overall, just tiny questions

.github/workflows/ci.yaml
alexkeating: I don't think so

.github/workflows/ci.yaml
alexkeating: I will remove

alexkeating:
Updated ci, should be good now

mds1 (approved):
CI failure unrelated and is fixed in #401

apbendi (commented):
I'm probably not the best person to review this—would love for @wildmolasses to take a look if he has the bandwidth.

That said, running `yarn lint` at the root I get the following reports from the packages:

```
@umbra/contracts-core: ✖ 9 problems (0 errors, 9 warnings)
@umbra/umbra-js: ✖ 4 problems (0 errors, 4 warnings)
@umbra/frontend: ✖ 8 problems (0 errors, 8 warnings)
```

Is this the expected output?

mds1:
Yea those warnings were already there. The config we use is pretty old and probably was just based on whatever defaults the tooling like quasar used, so if we want we can open a separate issue to decide on our lint config and make any required changes, but I don't think it should block this PR

apbendi:
> The config we use is pretty old and probably was just based on whatever defaults the tooling like quasar used

Got it. I thought based on the diff we'd changed the config, but looks like it was just re-ordered. Is that right @alexkeating?

alexkeating:
> > The config we use is pretty old and probably was just based on whatever defaults the tooling like quasar used
> 
> Got it. I thought based on the diff we'd changed the config, but looks like it was just re-ordered. Is that right @alexkeating?

Yes exactly, I basically made it so that typescript rules don't run on javascript files, but I didn't remove or add any new rules.

apbendi (approved):
Approved now that I understand we didn't change linting rules. Thanks @alexkeating!

Issue #374, opened 2022-08-09
fix: do not prompt for wallet connection if wallet is locked


mds1:
Right now if metamask is locked and I visit the app, just visiting https://app.umbra.cash/ will open MM and prompt me for my password. This is not great UX so we should lazily prompt for wallet connection

Issue #263, opened 2021-12-17
feat: encrypt send data and save it locally to (1) show a history of sends, and (2) help debugging


mds1:
Once a user sends funds to someone, there is currently no way for the sender to see who the funds were sent to. The only available information is the stealth address it what sent to. There are two main benefits to letting users see a history of where they sent funds:

1. It's very helpful for users to have a history of their sends so they can remember who they've sent funds to
2. When users are unable to receive funds, this helps debug as we can verify what the sender did

To do this, we'd need to encrypt the following data and save it in locally in the browser:
1. The recipient ID entered in the form
2. Whether advanced mode was used (i.e. recipient's pubkey or stealth key registry)
3. The random number? 

Options to do the encryption:
1. Require users to sign the message before sending, so we can use their normal Umbra-specific private key for the encryption
2. Only support this feature for wallets that support [`eth_getEncryptionPublicKey`](https://docs.metamask.io/guide/rpc-api.html#other-rpc-methods) and [`eth_decrypt`](https://docs.metamask.io/guide/rpc-api.html#eth-decrypt) RPC methods. Then we'd get their encryption pub key, generate a random private key, and save the corresponding random pubkey along with the encrypted data

For completion, I'll mention the option of storing the data unencrypted in local storage but this seems like a very bad idea. A browser extension could probably just read from local storage and de-anonymize all your transactions

alexkeating:
I can pick this up. Would we potentially want to store these messages outside of the browser in order to provide a consistent experience across browsers?

mds1:
@garyghayrat looks like you assigned yourself to this, do you still want to work on it or should we reassign to @alexkeating?

Re implementation, some thoughts:
1. Having a database store the encrypted sends would definitely be better UX across browsers, and more robust since it persists when browser storage is cleared. If we go this route, let's discuss with @apbendi  to see if there's any concerns around us having a DB with (encrypted) user info + any tooling preferences
2. Until then, we can save everything locally to start. Local storage _might_ fill up quick (I forget the size limit offhand) so we might want to use IndexedDB instead, but not all browsers support that and some use WebSQL. [localForage](https://github.com/localForage/localForage) is a wrapper around all browser storage APIs to make that easy, but it also hasn't been updated in a year
[...]

PR #407, opened 2022-10-23
Feature/add account center

### LINKED ISSUE #315: bug: Users cannot switch wallets without manually clearing local storage
Once you've connected to the app with a wallet, we remember your wallet app choice via a local storage variable. Then, whenever you return, we immediately attempt to reconnect with that wallet app. If you refuse to connect b/c you want to use another app, you just get this error:

image

The error is fine, but doesn't provide an option to change wallets. We should add a button that clears the local storage var and pops open the wallet select modal:

image

alexkeating:
## Description

I removed the the address settings layout and the network dropdown and replaced them with onboard/'s account center. 

Some things to be aware of:

1. The account center position is set to fixed and cannot be modified, but I can create an issue with onboard to resolve if this is something that must change. 
2. The mobile and desktop views are determined by the device on load and will not change with responsive changes to the screen. To view the mobile screen please reload the page when choosing a device in your browser.
3.  I didn't change the default color scheme, but it should be easy to change if we so desire.
4. The advanced emoji in address settings was removed, is this something we still want to keep? If so we may need to rethink where it is displayed.

Closes #315 #374


frontend/src/layouts/BaseLayout.vue
```diff
@@ -248,13 +226,11 @@ import BaseTooltip from 'src/components/BaseTooltip.vue';
...
 import useSettingsStore from 'src/store/settings';
 import useWalletStore from 'src/store/wallet';
-import AddressSettings from './AddressSettings.vue';
 import HeaderLinks from './HeaderLinks.vue';
-import NetworkDropdown from './NetworkDropdown.vue';
 
 export default defineComponent({
   name: 'BaseLayout',
-  components: { AddressSettings, ArgentWarningModal, BaseButton, BaseTooltip, ConnectWallet, HeaderLinks, NetworkDropdown }, // prettier-ignore
+  components: { ArgentWarningModal, BaseButton, BaseTooltip, ConnectWallet, HeaderLinks }, // prettier-ignore
```
mds1, line 233: _(arbitrary comment location to start a thread)_

1. Do we have any control over the style of this dropdown? It would be nice to have a themed dropdown here that matches the app style / previous style and shows the logos for each chain
2. The "Explore" link for smart contracts takes me to the FAQ. Is this expected? My guess was that it was going to link me to the Umbra contract on Etherscan for the currently selected chain

image

frontend/src/layouts/NetworkDropdown.vue
```diff
@@ -1,49 +0,0 @@
-<template>
```
mds1, line 1: _(arbitrary comment location to start a thread)_

When connected on Optimism / other chains, it no longer shows my ENS name, but I think it should based on how resolution is supposed to work, cc @apbendi to confirm

image

mds1 (commented):
Code LGTM and I like the account center overall, seems like there's a lot of quirks though so I wonder if this is worth using? Guess it depends on the answers to these comments 😅

frontend/src/layouts/NetworkDropdown.vue
apbendi: Yeah, ENS reverse resolution should always use mainnet. I believe we provided our own Provider to do this. Perhaps Blocknative doesn't support this—is there a way to override?

frontend/src/css/app.sass
```diff
@@ -1,4 +1,7 @@
-// app global css in Sass form
```
apbendi, line 1: *(Arbitrary comment location)*

Thanks for tackling this @alexkeating. It's nice to offload a lot of this to a well maintained third party tool, and it all seems to work.

Here are some things I noticed, though. I wonder which can actually be addressed, and if any of them will be showstoppers. Might want input from the rest of the team on the tradeoffs here.

* Ideally we would be able to hide Goerli from the list, but the dapp would still work if the user was connected to Goerli. Is this possible?
* ENS names only resolve when connected to mainnet (@mds1 mentioned this elsewhere too).
* There's not a lot of space to display the ENS name, and even relatively short ones get badly cut off

Screen Shot 2022-10-24 at 6 15 27 PM

* We don't show the user's ENS Avatar if it's set
* The floating account center can obscure parts of the UI when scrolling
Screen Shot 2022-10-24 at 6 22 17 PM

* The Loading Indicator shows up behind—or to the left— of the account center when switching networks:
umbra-loading-indicator

frontend/src/layouts/BaseLayout.vue
```diff
@@ -15,22 +15,15 @@
                     src="~assets/app-logo-128x128.png"
                     style="max-width: 50px"
                     class="q-ml-md"
+                    v-if="$q.screen.gt.sm || !network"
```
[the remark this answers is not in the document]
alexkeating: Yes, that is why it is all the way to the right. Right, now they only allow CSS customization with various CSS variables they provide. But, I think they are working on making customizatoin easier in the future with this ticket. https://github.com/blocknative/web3-onboard/issues/941

frontend/src/layouts/BaseLayout.vue
```diff
@@ -15,22 +15,15 @@
...
                   />
                   <div v-if="$q.screen.gt.sm" @click="$router.push({ name: 'home' })" class="text-h5 dark-toggle">
                     <span class="primary header-black q-ml-md">Umbra</span>
                   </div>
                   <div v-else-if="$q.screen.xs">
-                    <div v-if="isLoading" class="q-ml-md row">
+                    <div v-if="isLoading" class="row">
```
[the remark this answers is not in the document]
alexkeating: It looks like by default it truncates if the ENS name is less then 11 characters long, and it doesn't seem to be adjustable.

https://github.com/blocknative/web3-onboard/blob/860fad574b69efe62ef61c5e050f3f4cdbde64eb/packages/core/src/utils.ts#L83

frontend/src/layouts/BaseLayout.vue
```diff
@@ -57,19 +50,7 @@
...
-                  :userAddress="userAddress"
-                  :userDisplayName="userDisplayName"
-                  :advancedMode="advancedMode"
-                  class="row"
-                />
-              </div>
-              <network-dropdown />
-            </div>
-            <connect-wallet v-else-if="!isLoading">
+            <connect-wallet v-if="!isLoading && !network">
```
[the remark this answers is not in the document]
alexkeating: Yes, I think we can hide the loading spinner. I don't see the avatar being used by account center so my guess is that it isn't supported currently.

frontend/src/global.d.ts
```diff
@@ -0,0 +1,6 @@
+import { ExternalProvider } from 'ethersproject/providers';
```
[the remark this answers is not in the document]
alexkeating: It is needed to type window.ethereum which is used in order to prevent metamask from popping up when it is unlocked. I tried to originally include this in the vue-shims file, but it ended up causing build errors. I don't really know exactly why but splitting it to a separate file seemed to fix it.

frontend/src/css/app.sass
alexkeating: Based on both of your feedback I don't think account center is mature enough for our use case. I can split out the work for #374 into a separate pr, and close this one.

For #315 I can add a custom solution that looks like [Uniswap](https://github.com/ScopeLift/umbra-protocol/issues/315#issuecomment-1225944841), but I think we can omit the change button along with the transaction history. Does that work?

mds1: I agree with all of that. Though actually that "transaction history" section may be a good place to show the data in https://github.com/ScopeLift/umbra-protocol/issues/263, but we don't need to worry about that yet

apbendi: @alexkeating agree it sounds like Account Center might not work for the moment. Having not looked at the code for it at all, is there any world where maintaining a forked and modified version of account center (for now) is easier than rolling these solutions on our own? Maybe try to PR as many of our changes upstream as they'll accept, until it matures to the point we can switch back to the out of the box version? I don't have an opinion here, just asking for your assessment of that possibility vs. rolling our own.

alexkeating: That is definitely an option, but I think rolling our own will be less effort because we already have many of the pieces to recreate the UI Matt suggested.

PR #408, opened 2022-10-27
Do not prompt locked metamask

### LINKED ISSUE #374: fix: do not prompt for wallet connection if wallet is locked

alexkeating:
### Description

We do not prompt metamask if it is locked.

[Screencast from 10-26-2022 09:41:27 PM.webm](https://user-images.githubusercontent.com/12705146/198171249-f5363d56-2aff-4ebb-b120-d69acb134f86.webm)

# Closes #374

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -129,6 +129,13 @@ export default function useWalletStore() {
 
   watch(lastWallet, async () => {
     if (lastWallet.value && !userAddress.value) {
+      if (lastWallet.value === 'MetaMask') {
+        const unlocked = await window.ethereum._metamask.isUnlocked();
+        if (!unlocked) {
+          setLoading(false);
+          return;
+        }
+      }
       await connectWallet();
     }
   });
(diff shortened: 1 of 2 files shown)


apbendi (approved):
Looks good and seems to work well in testing. Thanks @alexkeating.

frontend/src/store/wallet.ts
```diff
@@ -129,6 +129,13 @@ export default function useWalletStore() {
 
   watch(lastWallet, async () => {
     if (lastWallet.value && !userAddress.value) {
+      if (lastWallet.value === 'MetaMask') {
```
mds1, line 132: Does `lastWallet` have a property we can key off of to determine if its injected web3? That way instead of special casing MetaMask we also cover Brave, Frame, Tally, etc.?

But presumably those have different ways of checking if they're unlocked, so we probably need to special case each. In which case, let's leave this as-is and we can special case others as requests for it come in

mds1: Seems like no based on the fields `WalletState` has (since `our `connectedWallet` value is of that type)

mds1 (approved):

frontend/src/store/wallet.ts
alexkeating: Yea, I looked there wasn't another way. It would definitely be nice if they had a generic implementation for this.

mds1: Cool, this LGTM then!

PR #410, opened 2022-11-06
Fix coverage and add to build

### LINKED ISSUE #323: fix code coverage
The `yarn coverage` commands in the `umbra-js/` and `contracts/` packages do not work

alexkeating:
### Description
- I fixed the existing coverage commands and added a monorepo command. I also added coverage to the github workflow

Closes #323

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,47 @@ jobs:
       - name: Run tests
         run: yarn test
 
+  coverage:
+    runs-on: ubuntu-latest
+    env:
+      FOUNDRY_PROFILE: ci
+      ETH_RPC_URL: ${{ secrets.ETH_RPC_URL }}
+      INFURA_ID: ${{ secrets.INFURA_ID }}
+      ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+      POLYGONSCAN_API_KEY: ${{ secrets.POLYGONSCAN_API_KEY }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - uses: volta-cli/action@v1
+
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - run: yarn install
+
+      - name: Run coverage
+        run: yarn coverage
+      - name: Merge directories
+        run: |
+          sudo apt update && sudo apt install -y lcov
+          lcov --add-tracefile ./contracts-periphery/lcov.info --add-tracefile ./umbra-js/coverage/lcov.info  --output-file lcov.info
+          # To ignore coverage for certain directories modify the paths in this step as needed. The
+      # below default ignores coverage results for the test and script directories. Alternatively,
+      # to include coverage in all directories, comment out this step.
+      - name: Filter directories
+        run: |
+          lcov --remove ./lcov.info 'contracts-periphery/test/*' 'contracts-periphery/script/*'
+      - name: Verify coverage level
+        uses: zgosalvez/github-actions-report-lcov@v1
+        with:
+          coverage-files: ./lcov.info
+          minimum-coverage: 82 # Set coverage threshold.
+          github-token: ${{ secrets.GITHUB_TOKEN }} # Adds a coverage summary comment to the PR.
+
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -47,6 +47,30 @@ jobs:
...
+
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - run: yarn install
+
+      - name: Run coverage
+        run: yarn coverage
```
alexkeating, line 73: Build is failing because of an invalid mnemonic phrase is there something I need to do to setup the CI correctly? Things seem to be working locally

mds1: From a high level on scope here:
- Coverage for the `umbra-js` and `contracts-periphery` packages would be great
- We can skip `contracts-core` since the core contracts aren't changing anymore and that package isn't really touched these days
- Frontend coverage is roughly zero so lets skip that too

Then when testing the current state  of things it seems:
- Running `yarn coverage` in the `umbra-js` directory worked
- Running `forge coverage` in the `contracts-periphery` directory worked, so we probably need to add an alias into the package.json there

The mnemonic issue you mentioned above seems to be for the `contracts-core` directory but we can skip that for coverage.

`forge coverage --report lcov` creates a coverage report at `./lcov.info`. I'm sure `nyc` can gnerate LCOV too since that's a standard format. So then maybe what we do is generate the two LCOV reports, filter out certain directories (such as `test/` in `contracts-periphery`), merge the two files then check for a minimum overall coverage threshold in CI? Here is how we filter out directories and check for minimum coverage threshold for a pure forge repo: https://github.com/ScopeLift/foundry-template/blob/f2b8c5f72a4c85533957537e804e6816d9744bc8/.github/workflows/ci.yml#L42-L68

mds1: Ah wait but build is failing because of the missing mnemonic, so I just added one to github CI. I used `test test test test test test test test test test silent junk` based on this https://github.com/ScopeLift/umbra-protocol/blob/727d678ee30a42f656705b13bc7bb7904b3c36a3/contracts-core/scripts/deploy-registry.js#L58-L62

alexkeating: I think this should be working with lcov now.

contracts-periphery/Makefile
```diff
@@ -13,3 +13,4 @@ install :; $(INSTALL_CMD)
 test :; forge test
 test-gas :; forge test --match-path *.gas.t.sol
 snapshot-gas :; forge test --match-path *.gas.t.sol --gas-report > snapshot/.gas
+coverage :; forge coverage --report lcov && sed -i 's/SF:/SF:contracts-periphery\//gI' ./lcov.info
```
mds1, line 16: If I run `yarn coverage` from the repo root, this part errors:

```
@umbra/contracts-periphery: Wrote LCOV report.
@umbra/contracts-periphery: sed: 1: "./lcov.info": invalid command code .
@umbra/contracts-periphery: make: *** [coverage] Error 1
@umbra/contracts-periphery: error Command failed with exit code 2.
@umbra/contracts-periphery: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run coverage exited 2 in '@umbra/contracts-periphery'
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
error Command failed with exit code 2.
```

IIRC sed behaves differently on macOS (which I'm using) vs linux, wonder if that's the error? Seems like it thinks the `.` part of `./lcov.info` is a command?

.github/workflows/ci.yaml
mds1: Awesome, yea all the CI checks now run so marking this resolved!

contracts-periphery/Makefile
alexkeating: Ah that is probably the case, I tested on linux and I will resolve

alexkeating: I think this should be resolved. I tested on mac and linux

mds1: confirmed, thanks!

mds1 (approved):
LGTM, will defer to @apbendi for a second review before merging

alexkeating:
@apbendi Bumping in case it was missed

apbendi (approved):
LGTM!

PR #411, opened 2022-11-07
best practices + style guide updates

mds1:
Ref https://book.getfoundry.sh/tutorials/best-practices and https://github.com/ScopeLift/foundry-template/

See commit messages for more info on some of the changes

### DIFF
--- a/contracts-periphery/script/Deploy.s.sol
+++ b/contracts-periphery/script/Deploy.s.sol
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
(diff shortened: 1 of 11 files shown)


contracts-periphery/script/Deploy.s.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
```
mds1, line 1: scopelint currently errors if the `script` dir is missing, so this is just a placeholder script until we write the actual deploy scripts

apbendi (approved):
Looks good to me 👍

garyghayrat (approved):

PR #414, opened 2022-11-14
Add batchSend deploy script

garyghayrat:
To test, run: 
1. `anvil --fork-url <ETH-RPC-URL> --fork-block-number 15971478`
2. `forge script script/Deploy.s.sol --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --broadcast --slow --fork-url http://127.0.0.1:8545`

Goerli deployed contract address `0xDf4B5B2C08A77078c05176D8dFAb6274D54215e9`

### DIFF
--- a/contracts-periphery/script/Deploy.s.sol
+++ b/contracts-periphery/script/Deploy.s.sol
@@ -10,7 +10,7 @@ contract Deploy is Script {
 
   function run() public {
     vm.broadcast();
-    UmbraBatchSend batchSend = new UmbraBatchSend(IUmbra(umbraAddress));
+    UmbraBatchSend batchSend = new UmbraBatchSend(IUmbra(UMBRA));
     console2.log("batchSend Address: ", address(batchSend));
   }
 }
--- a/contracts-periphery/script/Deploy.s.sol
+++ b/contracts-periphery/script/Deploy.s.sol
@@ -2,9 +2,15 @@
 pragma solidity ^0.8.16;
 
 import "forge-std/Script.sol";
+import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
+import {IUmbra} from "src/interface/IUmbra.sol";
 
 contract Deploy is Script {
+  address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+
   function run() public {
     vm.broadcast();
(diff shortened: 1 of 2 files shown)


contracts-periphery/script/Deploy.s.sol
```diff
@@ -2,9 +2,15 @@
...
+import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
+import {IUmbra} from "src/interface/IUmbra.sol";
 
 contract Deploy is Script {
+  address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+
   function run() public {
     vm.broadcast();
+    UmbraBatchSend batchSend = new UmbraBatchSend(IUmbra(UMBRA));
+    console2.log("batchSend Address: ", address(batchSend));
```
mds1, line 14: Probably can remove this line, deploys are logged by default. You can see the address once in the logs section and once in the spot where the tx was sent. Otherwise PR LGTM!

image

garyghayrat: Thanks for the review! Removed console log. And was able to deploy to Goerli for test purposes 0xDf4B5B2C08A77078c05176D8dFAb6274D54215e9

mds1 (approved):
LGTM!

apbendi (approved):
Looks good to me! Only note is that after rebasing, the (new) coverage check fails this PR because the Deploy script is not tested. Should we just remove the scripts file from the coverage report @mds1?

mds1:
Seems the CI is already set to filter those, but it's not working properly presumably due to the monorepo setup. Took a few shots at fixing but you can see I didn't get it yet. Whoever continues this, just check out what we currently do in CI, check out the available workflow settings from [the action](https://github.com/zgosalvez/github-actions-report-lcov) we use, and just get those to match up

mds1:
I will temporarily disable the contracts CI but keep umbra-js CI, merge, then fix the contracts CI in a separate PR

mds1:
Decided to just set threshold to zero for now in case the prior idea proved to be harder than I expected, once CI passes I'll merge and get CI working again in a separate branch

mds1:
CI failed because an umbra-js test is failing, merging then fixing

PR #416, opened 2022-11-27
Feature/sign sent messages

### LINKED ISSUE #263: feat: encrypt send data and save it locally to (1) show a history of sends, and (2) help debugging

alexkeating:
#### Description

When a user sends a payment through umbra we now save the record to the users browser. I also followed the receive page pattern and added a sent table that can be accessed from the homepage.

Also, I have a couple clarifying comments/questions...
1. Does the homepage look okay with 4 boxes? It looks a little wonky to me, but I am wondering what other people think.
2. I used aes encryption because it was more straightforward for me compared to xor, but am open to changing it if XOR makes sense.
3. Should we surface stealth address? Should I add whether it was sent in advance mode or add any other data potentially missing?

Closes #263 

##### Walkthrough of new feature
[Screencast from 11-26-2022 09:26:41 PM.webm](https://user-images.githubusercontent.com/12705146/204116667-478c426b-37ee-47b1-b001-3cdc98651006.webm)

#### Mobile table
Screenshot from 2022-11-26 21-29-06

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -32,7 +32,9 @@
     "@web3-onboard/walletconnect": "^2.1.3",
     "bnc-notify": "^1.5.1",
     "core-js": "^3.6.5",
+    "ethereum-cryptography": "^1.1.2",
     "ethers": "^5.6.8",
+    "localforage": "^1.10.0",
     "quasar": "^1.0.0",
     "vue-i18n": "8.27.1",
     "yaml-loader": "0.6.0"
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -87,22 +87,22 @@
         <!-- Card Layout for grid option -->
         <template v-slot:item="props">
           <div :key="props.row.id" class="col-12">
-            <q-card class="card-border cursor-pointer q-pt-md col justify-center items-center ">
-              <q-card-section class="q-px-sm row justify-center items-center">
+            <q-card class="card-border cursor-pointer q-pt-md col justify-center items-center">
+              <q-card-section class="row justify-center items-center">
                 <img class="q-mr-md" :src="getTokenLogoUri(props.row.token)" style="width: 1.2rem" />
                 <div class="text-primary text-h6 header-black q-pb-none">
                   {{ formatAmount(props.row.amount, props.row.token) }} {{ getTokenSymbol(props.row.token) }}
                 </div>
               </q-card-section>
               <q-card-section>
-                <div class="row justify-between items-center q-px-sm">
+                <div class="row justify-between items-center">
                   <div>{{ $t('AccountReceiveTable.sender') }}</div>
                   <div @click="copyAddress(props.row.from, 'Sender')" class="cursor-pointer copy-icon-parent">
                     <span>{{ props.row.from }}</span>
                     <q-icon color="primary" class="q-ml-sm" name="far fa-copy" />
                   </div>
                 </div>
-                <div class="row justify-between items-center q-px-sm">
+                <div class="row justify-between items-center">
                   <div>
                     <span class="q-mr-xs">{{ $t('AccountReceiveTable.stealth-receiver') }}</span>
                     <base-tooltip icon="fas fa-question-circle">
@@ -123,7 +123,7 @@
                     <q-icon color="primary" class="q-ml-sm" name="far fa-copy" />
                   </div>
                 </div>
-                <div class="row justify-between items-center text-caption text-grey q-px-sm">
+                <div class="row justify-between items-center text-caption text-grey">
                   <div>{{ $t('AccountReceiveTable.received') }}</div>
                   <div>
                     {{ formatDate(props.row.timestamp * 1000) }}
@@ -337,7 +337,7 @@
 
 <script lang="ts">
 import { computed, getCurrentInstance, defineComponent, onMounted, PropType, ref } from '@vue/composition-api';
-import { date, copyToClipboard } from 'quasar';
+import { copyToClipboard } from 'quasar';
 import { BigNumber, Block, joinSignature, formatUnits, TransactionResponse, Web3Provider } from 'src/utils/ethers';
 import { Umbra, UserAnnouncement, KeyPair, utils } from '@umbra/umbra-js';
 import useSettingsStore from 'src/store/settings';
@@ -353,6 +353,7 @@ import { FeeEstimateResponse } from 'components/models';
 import { formatNameOrAddress, lookupOrReturnAddresses, toAddress, isAddressSafe } from 'src/utils/address';
 import { MAINNET_PROVIDER } from 'src/utils/constants';
 import { getEtherscanUrl, isToken } from 'src/utils/utils';
+import { formatDate, formatAmount, formatTime, getTokenSymbol, getTokenLogoUri } from 'src/utils/utils';
 
 function useAdvancedFeatures(spendingKeyPair: KeyPair) {
   const vm = getCurrentInstance()!;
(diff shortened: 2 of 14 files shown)


frontend/src/utils/account-send.ts
```diff
@@ -0,0 +1,89 @@
+import { encrypt, decrypt } from 'ethereum-cryptography/aes';
+import { hexToBytes, utf8ToBytes, bytesToUtf8 } from 'ethereum-cryptography/utils';
+import { getRandomBytesSync } from 'ethereum-cryptography/random';
+import localforage from 'localforage';
```
mds1, line 4: Thoughts on using the [config](https://github.com/localForage/localForage#configuration) to avoid using Web SQL here? Seems we may be able to force using indexed DB, since Web SQL is being [deprecated](https://developer.chrome.com/blog/deprecating-web-sql/)

frontend/package.json
```diff
@@ -32,7 +32,9 @@
     "@web3-onboard/walletconnect": "^2.1.3",
     "bnc-notify": "^1.5.1",
     "core-js": "^3.6.5",
+    "ethereum-cryptography": "^1.1.2",
```
mds1, line 35: _(arbitrary comment location to start a thread)_

> Does the homepage look okay with 4 boxes? It looks a little wonky to me, but I am wondering what other people think.

I agree it looks a bit wonky. Wdyt about putting the send history in the account info box, similar to what uniswap does? We can basically put the table component there, may need some formatting changes though if it doesn't fit nicely

image

image

image

frontend/package.json
```diff
@@ -32,7 +32,9 @@
...
     "ethers": "^5.6.8",
+    "localforage": "^1.10.0",
```
mds1, line 37: _(arbitrary comment location to start a thread)_

> I used aes encryption because it was more straightforward for me compared to xor, but am open to changing it if XOR makes sense.

Sounds good, let me check on this and get back to you

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -337,7 +337,7 @@
 
 <script lang="ts">
 import { computed, getCurrentInstance, defineComponent, onMounted, PropType, ref } from '@vue/composition-api';
-import { date, copyToClipboard } from 'quasar';
+import { copyToClipboard } from 'quasar';
```
mds1, line 340: _(arbitrary comment location to start a thread)_

> Should we surface stealth address? Should I add whether it was sent in advance mode or add any other data potentially missing?

Good call, it would be nice to also track:
1. ENS/CNS name AND address that resolved to, or just address if no ENS was given (since a user may enter an ENS name, and what it resolves to at the time of their tx may be different than what it resolves to in the future
2. Was advanced mode enabled (as a boolean)
3. Was the "Send using recipient's standard public key" box checked (as a boolean)

Items 2 and 3 I don't think we need to show the user, but they are good to have available for debugging, etc.

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -472,40 +473,28 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
   };
 
   // Table formatters and helpers
-  const formatDate = (timestamp: number) => date.formatDate(timestamp, 'YYYY-MM-DD');
-  const formatTime = (timestamp: number) => date.formatDate(timestamp, 'h:mm A');
   const isNativeToken = (tokenAddress: string) => tokenAddress === '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
-  const getTokenInfo = (tokenAddress: string) => tokens.value.filter((token) => token.address === tokenAddress)[0];
+  const getTokenInfo = (tokenAddress: string) => tokens.value.filter(token => token.address === tokenAddress)[0];
```
mds1, line 477: Can you regenerate the lockfile (mine differs after running `yarn`) and run `yarn prettier` (some changes in here seem to have been made by your editor on save)

frontend/src/components/AccountSendTable.vue
```diff
@@ -0,0 +1,184 @@
...
+                {{ props.row.tokenSymbol }}
+              </div>
+            </div>
+          </div>
+
+          <!-- Sender column -->
+          <div v-else-if="col.name === 'from'" class="d-inline-block">
+            <div @click="copyAddress(props.row.address)" class="cursor-pointer copy-icon-parent">
+              <span>{{ props.row.addressShortened }}</span>
+              <q-icon class="copy-icon" name="far fa-copy" right />
```
mds1, line 78: If we have an ENS name, let's show that as the main content here instead, and we can show the shortened address underneath in smaller/grey text

frontend/src/components/AccountSendTable.vue
```diff
@@ -0,0 +1,184 @@
...
+      type: (undefined as unknown) as PropType<SendTableMetdataRow[]>,
+      required: true,
+    },
+  },
+
+  setup(props, context) {
+    const vm = getCurrentInstance()!;
+    const { provider, chainId } = useWalletStore();
+    const paginationConfig = { rowsPerPage: 25 };
+    async function copyAddress(address: string) {
```
mds1, line 124: Can we dedupe this? This function is defined/implemented three times

frontend/src/components/AccountSendTable.vue
```diff
@@ -0,0 +1,184 @@
...
+      if (!provider.value) throw new Error(vm.$i18n.tc('AccountSendTable.wallet-not-connected'));
+      const mainAddress = await toAddress(address, provider.value);
+      await copyToClipboard(mainAddress);
+      notifyUser('success', `${vm.$i18n.tc('AccountSendTable.address-copied')}`);
+    }
+    function openInEtherscan(hash: string) {
```
mds1, line 130: This one is similarly defined twice, but maybe ok since `getEtherscanUrl` is the heavy lifting

frontend/src/i18n/locales/zh-cn.json
```diff
@@ -276,6 +276,29 @@
     "withdraw-in-progress": "正在提款...",
     "view-transaction": "查看交易"
   },
+  "AccountSendTable": {
```
mds1, line 279: @garyghayrat can you review all the translations here?

frontend/src/pages/AccountSend.vue
```diff
@@ -475,12 +478,28 @@ function useSendForm() {
         }
       }
 
-      // Send with Umbra
+      if (!viewingKeyPair.value?.privateKeyHex) {
```
mds1, line 481: (note to self: continue reviewing from here)

mds1 (commented):
Didn't get to finish reviewing yet, but left some comments so far!

frontend/src/utils/account-send.ts
```diff
@@ -0,0 +1,89 @@
...
+  amount: string;
+  address: string;
+  dateSent: Date;
+  hash: string;
+  tokenAddress: string;
+};
+
+const encryptAddress = async (recipientAddress: string, privateKey: string) => {
+  const iv = getRandomBytesSync(16);
+  const encrypted = await encrypt(utf8ToBytes(recipientAddress), hexToBytes(privateKey), iv, 'aes-256-cbc');
```
mds1, line 28: From a conversation with @seresistvanandras, we should be able to encrypt with xor as follows:
- Start encryption counter `c = 0`
- On first send, encryption key `k` is `keccak256(viewingPrivateKey, c)`
- Take 160 bit recipient address, append one bit indicating whether or not advanced mode was enabled, and a second bit indicated whether the checkbox was checked, giving 162 bits. Call this data `d`
- Pad `d` with zeros so it's length is 256 bits. Padding direction probably doesn't matter, so let's right pad since that's how `bytesNN` behaves in solidity and I think ethers.js has methods to do that (hexZeroPad)
- Encrypted address `a = k ^ d` where `^` is xor
- Increment `c` to `1`, save `a` and `c` in browser storage (we of course only need to store the new (current) value of `c` since it increments, but should save all `a` values)
- Decrypt with `d = a ^ k` 

This should be safe because:
- With AES or this approach, compromised viewing private key means compromised send history
- Even if you have `k`, you can't back out the viewing key from it even if you know the counter values, thanks to the hashing

Nice benefits of the xor + counter approach:
- Supports random access into the history at any point, i.e. you can decrypt just the 10th tx sent. This doesn't really matter currently but may be helpful one day for people with hundreds/thousands of sends
- If local storage is cleared, and you know with certainty that tx `i` was sent to recipient `r`, you can figure out the required counter value to decrypt that send correctly and then decrypt the rest from there
- No new encryption deps needed

mds1: > encrypted address `a`, is also written to the blockchain or not? Or is it only stored in the browser of the user?

`a` would only be stored locally in the user's browser.

The rest of this makes sense to me though, thanks for confirming things!

frontend/src/utils/account-send.ts
alexkeating: Totally agree

frontend/package.json
alexkeating: Yes, totally I can give that a shot

frontend/package.json
alexkeating: Any updates here?

frontend/src/components/AccountReceiveTable.vue
alexkeating: Okay yes I can add those checkboxes

alexkeating:
Going to close this in favor of #465 and a separate cryptography pr.

PR #421, opened 2022-12-01
ci/fixes

mds1:
## Fixes flaky umbra-js test

Previously we used public RPCs in tests, but the arbitrum one was flaky and not finding a tx that did exist. As a result we got this failure:

```
  1) Utilities
       Recipient identifier lookups
         looks up transaction history on arbitrum one:
     Error: Transaction hash not found. Are the provider and transaction hash on the same network?
      at getTransactionByHash (src/utils/utils.ts:397:11)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at Object.getSentTransaction (src/utils/utils.ts:148:24)
      at Context.<anonymous> (test/utils.test.ts:198:22)
```

Replacing the public RPC with Infura resolved it, so I also made that change for the other RPCs

## CI Coverage Fixes

- Added `--rc lcov_branch_coverage=1` to the `lcov` filtering command. This keeps branch coverage information in the output lcov file (for some reason the default is to remove it), and this seems to have fixed the CI action
- I used the results to set the threshold to 70%
- Modified the `contracts-periphery` coverage command to `forge coverage --report lcov --report summary`. This means it will both (1) output an lcov file, and (2) print the summary table. This is useful because lcov just gives a high level summary, so printing the table too means we can view the CI job logs for more granular info

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,10 +72,11 @@ jobs:
 
       - name: Run coverage
         run: yarn coverage
+
       - name: Merge directories
         run: |
           sudo apt update && sudo apt install -y lcov
-          lcov --add-tracefile ./contracts-periphery/lcov.info --add-tracefile ./umbra-js/coverage/lcov.info  --output-file lcov.info
+          lcov --add-tracefile ./contracts-periphery/lcov.info --add-tracefile ./umbra-js/coverage/lcov.info  --output-file lcov.info --rc lcov_branch_coverage=1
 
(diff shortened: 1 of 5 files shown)


apbendi (approved):
Looks good to me. Ran tests multiple times locally and they passed without issue. Seemed to run faster too—wonder if Infura responds faster than the public RPCs. Good to merge!

mds1:
> wonder if Infura responds faster than the public RPCs

almost certain it does, I think public ones are typically throttled/rate-limited

PR #423, opened 2022-12-13
chore: update snapshots

mds1:
Updating snapshots, since I have an idea on how to make batch sends simpler/cheaper and want to diff the snapshots

Also fixed solc/scopelint issues that arose in CI for this issue


apbendi (approved):
LGTM

Issue #275, opened 2022-01-18
Honor startBlock & endBlock parameters when fetching announcements from The Graph


apbendi:
When fetching payment announcements via subgraph, in the `Umbra` method inside `umbra-js`, we currently do not honor the `startBlock` and `endBlock` parameters, and instead always fetch all announcements. We should update the queries to respect these parameters.

alexkeating:
I think #520 handled this

PR #425, opened 2022-12-14
Quasar v2 / Vue 3

mds1:
To run this branch, delete all `node_modules` folders before installing with `rm -rf node_modules **/node_modules`

Known issues:

1. ~~network selector doesn't prompt you to change networks (it did work the other day though, so seems I broke this recently and the fix should be easy)~~ fixed in af1256f
2. ~~language selector at the bottom doesn't populate by default~~ fixed in 87c3539
3. ~~faq page works, but throws errors in the console due to some i18n interpolation changes (we need to update strings to escape certain characters, see [here](https://github.com/intlify/vue-i18n-next/issues/127))~~ fixed in 7caac17
4. ~~sending does not work, details below, which I'm assuming affects all sending of txs~~ fixed in 971aa53
5. ~~receiving always shows "nothing found" when it's done scanning, which I'm guessing is related to some web worker change we need~~ fixed in 8520d1c (not web worker related)
6. ~~some form validations are broken due to quasar changes. I think [this](https://quasar.dev/start/upgrade-guide#-new-connecting-to-qform) is the fix but haven't investigated yet~~ fixed in 9a36d86
7. ~~ENS avatars don't work usually? Lots of requests and they 429~~ fixed in aef7e00
8. ~~Warnings such as `shared.esm-bundler.js?d074:54 [intlify] Detected HTML in 'Use the <span class="text-bold">Setup</span> page to configure your account' message. Recommend not using HTML messages to avoid XSS.` — This is from our use of `v-html`, which we use for i18n stuff. So these warnings can be safely ignored since it's coming from a local, trusted source and not external input. TBD about turning off this warning or ignoring all/some of them somehow~~ fixed in 2544383
9. ~~Benchmark scan times with web workers on this branch vs production across networks. If scan times are the same or this branch is faster, we _can_ consider leaving leave the implementation alone for now. Otherwise, we probably have fixes to make. Regardless, look into removing the `worker-loader` dependency used by `worker.ts` since it's [no longer needed](https://webpack.js.org/guides/web-workers/) now that we're upgrading from webpack v4 to v5.~~
    - Benchmarked using my account on polygon with the live site vs. deploy preview for ba552dc. Live site was 24s to fetch announcements, 69s to scan. Deploy preview was 25s to fetch announcements, 67 to scan. This is close enough to conclude performance is unaffected (though, small sample size). Also updated to remove `worker-loader` in 93f8acc
11. ~~When withdrawing you get the following error (with both ENS and regular address): `runtime-core.esm-bundler.js?69b0:4178 Uncaught (in promise) TypeError: Cannot convert object to primitive value`~~ fixed in c7e9124

TODO before deploy:
- [ ] @apbendi to look into doing a phased rollout via Netlify
- [ ] @mds1 and @davidlaprade to ensure we're familiar with how to rollback deploys on Netlify


stratus4:
So far just found this page button delay bug
[Screencast from 2023-01-16 15-53-22.webm](https://user-images.githubusercontent.com/67454987/212764317-ff387884-387c-41c7-afe6-828a939548e7.webm)

.eslintrc.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
         'eslint:recommended',
         'plugin:@typescript-eslint/recommended',
         'plugin:@typescript-eslint/recommended-requiring-type-checking',
-        'plugin:vue/essential',
+        'plugin:vue/vue3-essential',
```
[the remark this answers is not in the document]
mds1: Should be resolved by 533633f

davidlaprade:
Just manually tested and found two small issues:
* on arbitrum, [this payment link](http://localhost:8080/send?to=vitalik.eth&token=AETH&amount=0.01) didn't populate the amount
* on polygon, with advanced mode enabled, on the receive page, I specified a from-block to begin scanning but it still scanned blocks before the specified number

Otherwise things looked great!

davidlaprade:
Just manually tested and found this issue:
* on polygon, with advanced mode enabled, on the receive page, I specified a from-block to begin scanning but it still scanned blocks before the specified number

frontend/src/pages/AccountReceive.vue
```diff
@@ -81,7 +81,7 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent, onMounted, ref, watch } from '@vue/composition-api';
+import { computed, defineComponent, onMounted, ref, watch } from 'vue';
```
davidlaprade, line 84: _mostly arbitrary place_

Just manually tested and found this issue:
* on polygon, with advanced mode enabled, on the receive page, I specified a from-block to begin scanning but it still scanned blocks before the specified number

frontend/src/utils/payment-links.ts
```diff
@@ -1,7 +1,7 @@
 import { copyToClipboard } from 'quasar';
 import { TokenInfoExtended } from 'components/models';
 import { utils as umbraUtils } from '@umbra/umbra-js';
-import useWalletStore from 'src/store/wallet';
+import { providerExport as provider, relayerExport as relayer, tokensExport as tokens } from 'src/store/wallet';
```
davidlaprade, line 4: _(mostly arbitrary place to stark a thread)_

Just manually tested and found this issue:
* on arbitrum, [this payment link](http://localhost:8080/send?to=vitalik.eth&token=AETH&amount=0.01) didn't populate the amount

frontend/src/pages/AccountReceive.vue
mds1: This is tracked in https://github.com/ScopeLift/umbra-protocol/issues/275, so going to mark this resolved. Nice job catching it in the QA testing, appreciate the thoroughness 👌

frontend/src/utils/payment-links.ts
mds1: Ah interesting, seems it only populates on mainnet. The token in the link is AETH, I wonder if the difference in naming native tokens has to do with it. Going to test using ERC20 tokens too, let's see...

mds1: Ok here's (part of?) the bug, haven't implemented a fix yet:

In `AccountSend.vue`, the payment link is properly parsed and the component's state is properly. This happens immediately when the component mounts. In that component, we watch various state and update accordingly:

https://github.com/ScopeLift/umbra-protocol/blob/533633f7971695afe4b224ec5bef26780780ec5e/frontend/src/pages/AccountSend.vue#L333-L342

Every time any of that state changes, we verify that the selected token exists on the selected network, and if not we clear the token and amount selection:

https://github.com/ScopeLift/umbra-protocol/blob/533633f7971695afe4b224ec5bef26780780ec5e/frontend/src/pages/AccountSend.vue#L366-L373

This is the source of a bug: The data populates, then the relayer hasn't finished loading so the only token in the list is ETH, therefore our tokens are not found and we reset the token selection and amount. This line of code was required when we didn't refresh the page on network change. Now that we do that, I'm pretty sure this block can be deleted. From a quick test, deleting it caused some errors, so need to look a bit more but just documenting this for now

mds1: I also pushed 4aa2b44 which renames all native-ETH tokens to just ETH, instead of AETH for arbitrum and OETH for optimism. When L2s were newer they tried that naming convention to help users distinguish. Now, looking at the official arbitrum and optimism bridges, it seems they no longer make that distinction.

mds1: Just pushed 9e88061 to make payment link parsing more robust. @davidlaprade mind testing that out to confirm it fixes the issue?

davidlaprade: Looks great! Thanks for the fix

This is unrelated to this PR -- more like a feature request than anything -- but it'd be nice if a payment link could prompt a user to change networks, if necessary. If you're requesting payment from someone with a link, network matters. I don't want to receive (say) DAI on Polygon if I don't have Matic there. Currently, someone could be on mainnet, create the payment link to request DAI, then the recipient could be logged in on Polygon, and paste the link thinking it's fine to send on Polygon when actually it's not.

mds1: I was thinking about that too as I was fixing this, tracking in https://github.com/ScopeLift/umbra-protocol/issues/439

.eslintrc.js
apbendi, line 24: *(Arbitary comment location)*

On the send page, if there is an error in the "recipient" field, and the user tries to edit the "amount" field, the "recipient" field steals the cursor focus every time a change is made to the "amount" field.

recipient-error-steals-focus

contracts-core/.prettierignore
```diff
@@ -10,3 +10,4 @@ typechain/
 
 # files
 coverage.json
+*.sol
```
apbendi, line 13: *(Arbitrary comment location)*

If I go to the receive page and perform a scan with an address that has *not yet configured Umbra* I get the following error(s) after scanning completes.

Screen Shot 2023-01-19 at 1 33 30 PM

apbendi: This still happens even when scanning with a private key, if the account connected has not configured Umbra.

.eslintrc.js
mds1: Ha nice find. Here's why it happens:
- We have a watcher to watch for changes in a few form-related parameters
- One is the amount. When amount changes, we validate the new amount AND validate the full form
- So when you type the first character in the amount, the form gets validated. The recipient is validated, and validation fails, resulting in the focus on the send component

Interestingly, it seems this doesn't exist on master, so I'm guessing I introduced it in 9e88061 when modifying that watcher to fix payment link parsing. This isn't _terrible_ since you'll need to fix the recipient anyway, and I think this whole watcher setup we have should be cleaned up and refactored since a lot of it was to make sure things worked properly when we didn't refresh on network change, so I pulled this out in https://github.com/ScopeLift/umbra-protocol/issues/441 to track.

Still open to getting the fix in during this PR, lmk what you think

contracts-core/.prettierignore
mds1: Should now be fixed, c774dea

.eslintrc.js
apbendi: If there's an easy fix (even a hacky one) I think it would be better to fix it if possible. Obviously it's poor UX, but also there's a risk it could lead to a user mangling the recipient address, changing one character by accident or something because they don't know what's going on, and sending funds to the wrong place. Relatively low probability but still. That said if it's a big pain to fix, then I'm ok merging to get this PR in, as long as we can come back to this relatively quickly to get it fixed.

contracts-core/.prettierignore
apbendi: Yep! Seems to be fixed.

.eslintrc.js
mds1: Ok this _should_ be good now. Tested it a bit and it seems to work. Needs some more testing, both in normal flow (visit send page and use it) and payment link flow (load a payment link and maybe modify it). Also pushed the banner changes

PR #426, opened 2022-12-15
Frontend-batchSend: Umbra-js

garyghayrat:
Adds `batchSend` functionality to Umrba.js


contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
```diff
@@ -0,0 +1,4 @@
+{
+  "_format": "hh-sol-dbg-1",
```
mds1, line 2: What are all these new files in `contracts-core`? I'd think we don't need to modify this folder, especially since it seems there's BatchSend artifacts here?

contracts-core/periphery-typechain/UmbraBatchSend.ts
```diff
@@ -0,0 +1,228 @@
+/* Autogenerated file. Do not edit manually. */
```
mds1, line 1: We should avoid committing these autogenerating files, and they should be generated by the build command

umbra-js/src/classes/Umbra.ts
```diff
@@ -27,14 +27,19 @@ import { KeyPair } from './KeyPair';
 import { RandomNumber } from './RandomNumber';
 import { blockedStealthAddresses, getEthSweepGasInfo, lookupRecipient, assertSupportedAddress } from '../utils/utils';
 import { Umbra as UmbraContract, Erc20 as ERC20 } from '@umbra/contracts-core/typechain';
+import { UmbraBatchSend as BatchSendContract } from '@umbra/contracts-core/periphery-typechain';
 import { ERC20_ABI } from '../utils/constants';
-import type { Announcement, ChainConfig, EthersProvider, ScanOverrides, SendOverrides, SubgraphAnnouncement, UserAnnouncement, AnnouncementDetail } from '../types'; // prettier-ignore
+import type { Announcement, ChainConfig, EthersProvider, ScanOverrides, SendOverrides, SubgraphAnnouncement, UserAnnouncement, AnnouncementDetail} from '../types'; // prettier-ignore
 
 // Umbra.sol ABI
 const { abi } = require('@umbra/contracts-core/artifacts/contracts/Umbra.sol/Umbra.json');
```
mds1, line 35: Let's make this one unambiguous too, since now we have multiple ABIs imported
```suggestion
const { abi: umbraAbi } = require('@umbra/contracts-core/artifacts/contracts/Umbra.sol/Umbra.json');
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -129,6 +149,7 @@ export class Umbra {
   constructor(readonly provider: EthersProvider, chainConfig: ChainConfig | number) {
     this.chainConfig = parseChainConfig(chainConfig);
     this.umbraContract = new Contract(this.chainConfig.umbraAddress, abi, provider) as UmbraContract;
+    this.batchSendContract = new Contract(batchSendAddress, batchSendAbi, provider) as BatchSendContract;
```
mds1, line 152: Let's add this to the `chainConfig`
```suggestion
    this.batchSendContract = new Contract(this.chainConfig.batchSendAddress, batchSendAbi, provider) as BatchSendContract;
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -43,6 +48,20 @@ const subgraphs = {
...
+  address: string;
+}
+
+interface SendData {
+  receiver: string;
+  tokenAddr: string;
+  amount: BigNumberish;
+  pkx: string;
+  ciphertext: string;
+}
```
mds1, line 63: Let's define these in `types.ts` instead (seems `SendBatch` is already there)

umbra-js/src/types.ts
```diff
@@ -115,3 +115,9 @@ export interface UserAnnouncement {
   token: string; // token address
   txHash: string;
 }
+
+export interface SendBatch {
+  address: string;
+  token: string;
+  amount: string;
```
mds1, line 122: When you have a type that expects user-provided inputs (e.g. token amounts), you typically want to take a `BigNumberish` so the API is more flexible and easier to use, then the function that consumes this `amount` should `BigNumber.from(amount)` to parse it
```suggestion
  amount: BigNumberish;
```

OTOH, when returning a value from a function, `BigNumber` is what you want (or `bigint` in a more modern codebase)

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
     return { tx, stealthKeyPair };
   }
 
+  async batchSend(signer: JsonRpcSigner | Wallet, batch: SendBatch[], overrides: SendOverrides = {}) {
+    let tokenSum = new Map();
+    let tokens: string[] = [];
+    for (let i = 0; i < batch.length; i++) {
```
mds1, line 262: This loop and the next loop are trying to do a lot of things: validate recipient address, compute total sums of tokens, get an array of unique tokens, and check balances. We can make things a bit cleaner here, and _a lot_ more performant by taking the `await`s out of the loops

The code below renames the `batch` input to something plural like `sends`. The code should be considered pseduocode, may not work but conveys the idea. Lmk what you think of something like this:

```typescript
async batchSend(signer: JsonRpcSigner | Wallet, sends: SendBatch[], overrides: SendOverrides = {}) {
  // Get all unique recipients and validate them
  const recipients = new Set(sends.map(send => send.recipient))
  await Promise.all(recipients.map(assertSupportedAddress))

  // Get a unique list of tokens
  const tokens = new Set(sends.map(send => send.token))

  // Validate balances
  //    - pull out the validation logic re-used by both these functions into a helper method, say
  //      it's called `assertSufficientBalance`
  //    - this line assumes you created a `balances` mapping somewhere above, perhaps
  //      using `for (const send of sends) {` 
  await Promise.all(tokens.map(token => assertSufficientBalance(balances.get(token)))

  // --- snip ---
}
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
...
+        }
+      }
+    }
+
+    const txSigner = this.getConnectedSigner(signer); // signer input validated
+
+    // Get toll amount from contract
+    const toll = await this.umbraContract.toll();
+
+    // Parse provided overrides
```
mds1, line 294: This also seems like a good helper function to create to avoid duplicating code

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
...
+      const randomNumber = new RandomNumber();
+
+      // Encrypt random number with recipient's public key
+      const encrypted = viewingKeyPair.encrypt(randomNumber);
+
+      // Get x,y coordinates of ephemeral private key
+      const { pubKeyXCoordinate } = KeyPair.compressPublicKey(encrypted.ephemeralPublicKey);
+
+      // Compute stealth address
+      const stealthKeyPair = spendingKeyPair.mulPublicKey(randomNumber);
```
mds1, line 336: We also talked about a helper method to do all of this logic, i.e. take the recipient, token address, and amount and return the stealth key pair, ciphertext, and pkx

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
...
+      stealthKeyPairs.push(stealthKeyPair);
+
+      // Ensure that the stealthKeyPair's address is not on the block list
+      if (blockedStealthAddresses.includes(stealthKeyPair.address)) throw new Error('Invalid stealth address');
```
mds1, line 340: Perhaps an `assertValidStealthAddress` helper would be nice here. Let's also specify the stealth address in the error string to help debug

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
...
+
+      // Sum valueAmount
+      if (isEth(token)) {
+        valueAmount = valueAmount.add(toll.add(amount));
+      } else {
+        valueAmount = toll.add(valueAmount);
+      }
```
mds1, line 347: Nit
```suggestion
      // Sum valueAmount
      valueAmount = valueAmount.add(toll);
      if (isEth(token)) {
        valueAmount = valueAmount.add(amount);
      }
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +256,119 @@ export class Umbra {
...
+    }
+
+    // Sort by token addresses
+    const sortedData: SendData[] = sendData.sort((n1, n2) => {
+      return n1.tokenAddr.localeCompare(n2.tokenAddr);
+    });
+
+    // Send transaction
+    let tx: ContractTransaction;
+    const txOverrides = { ...localOverrides, value: valueAmount, gasLimit: 5000000 };
```
mds1, line 364: Why do we hardcode a gas limit here? We should be able to remove that
```suggestion
    const txOverrides = { ...localOverrides, value: valueAmount };
```

umbra-js/src/types.ts
```diff
@@ -115,3 +115,9 @@ export interface UserAnnouncement {
...
+}
```
mds1, line 123: _(arbitrary comment location to start a thread)_

We should also add tests for all new methods (both public and helper methods). Check out `Umbra.test.ts` to see how we test the existing `send` method

contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
garyghayrat: I was able to use `/out` as the artifact but I think I'm having `ethers` version problem with the typechain file. Because if I change
```
import { UmbraBatchSend as BatchSendContract } from '@umbra/contracts-core/periphery-typechain/UmbraBatchSend';
```
to 
```
...from '@umbra/contracts-periphery/periphery-typechain/UmbraBatchSend';
```
I'm getting type errors. Also getting type errors when autogenerating the files using `yarn run typechain --target ethers-v5 --out-dir ./typechain './out/UmbraBatchSend.sol/UmbraBatchSend.json'` and when using
```
...from '@umbra/contracts-periphery/typechain/UmbraBatchSend';
```
Tried to play around with the `contract-periphery` folder's `package.json` by installing the `^5.6.8` version of `ethers` and other few older versions of `typechain` and `typechain/ethers-v5` to no avail

mds1: What if we just blow away all the typechain stuff for now since it's more of a nice to have, get things working, and we can improve the typings later? I'm getting a bit confused by all the autogenerated files, and I think @wildmolasses may have setup some lerna build order config for ensuring typechain gets generated in `contracts-core` before the umbra-js package is built, so might need to ask him for help here too

contracts-periphery/script/ApproveToken.s.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import "forge-std/Script.sol";
+import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
+import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
+
+contract ApproveTokenScript is Script {
+  UmbraBatchSend batchSend = UmbraBatchSend(0x0d81Df222BB44b883265538586829715CF157163);
+  address token = 0x57Cf115b0Abdce35dC9A8A8D9DfF1f586eA63ec8; //USDC
```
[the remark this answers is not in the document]
garyghayrat: Removed it! 313b0cc125b23afc4f86c91c803ec53560396e76

umbra-js/src/classes/Umbra.ts
garyghayrat: Addressed

umbra-js/src/classes/Umbra.ts
garyghayrat: Added in b5ed2dde8d6d175dc42203ca8d5a04106ecc5777

umbra-js/src/classes/Umbra.ts
garyghayrat: Moved them in 4618fab3d65ea08e70613c8af301ae13e03e0b66

umbra-js/src/classes/Umbra.ts
garyghayrat: Definitely cleaner, but my implementation still uses for loops. Using `Promise.all` seem to just add an extra step but maybe I'm wrong? 082acd9876681cd3c0488d5f4d824b8af1507f88

umbra-js/src/classes/Umbra.ts
garyghayrat: Added paseOverrides helper 20d4e6f4518e7236c5467c398a8186cb17b7f319

umbra-js/src/classes/Umbra.ts
garyghayrat: Checkout prepareSend helper method here 1b968b7f98e2a8dc3a68e91dbe806bde65b17568

umbra-js/src/classes/Umbra.ts
garyghayrat: Added assertValidStealthAddress helper in 1b968b7f98e2a8dc3a68e91dbe806bde65b17568

umbra-js/src/classes/Umbra.ts
garyghayrat: Addressed!

umbra-js/src/classes/Umbra.ts
garyghayrat: I was getting gas errors when testing along with the frontend, not sure why. But removed it for now!

contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
garyghayrat: Oh I didn't know they were nice to haves. I was looking at the implementation for Umbra's `send` and thought they were essential.

mds1: To expand on why typechain is just a nice to have, and not essential: By default if you create a new contract in ethers with `myContract = new Contract(...)` you get an object of the `Contract` type, i.e. an instance of the `Contract` class. So let's say you pass in an ERC-20 ABI. Then you can call `myContract.name()` as expected. But one downside here is that you have to tell typescript that you expect a `string` to be returned, it doesn't know what that method is supposed to return.

You may accidentally call `myContract.nameeee()`, which of course doesn't exist in an ERC-20. Your code will compile, then you'll get a runtime error since that method doesn't exist on the called contract.

Using Typechain solves these problems by giving you classes such as `UmbraBatchSend` which extend the default ethers `Contract` class based on the ABI. So now (1) your TS code knows `name` returns a string, and (2) your TS code knows all available methods and won't compile if you try calling a method that doesn't exist.

Probably some more info at https://github.com/dethcrypto/TypeChain

umbra-js/src/classes/Umbra.ts
mds1: https://github.com/ScopeLift/umbra-protocol/blob/1b968b7f98e2a8dc3a68e91dbe806bde65b17568/umbra-js/src/classes/Umbra.ts#L218-L221

Like you said, the above still uses a for loop. This isn't great because, say you have 10 addresses. Each iteration of the for loop `await`s a response, and only after 10 serial awaits can we move on. This is slow and therefore bad UX (`await` in a for loop is typically a code smell that something is suboptimal and should be refactored)

If you replace the for loop with:

```typescript
const assertionPromises = recipients.map(assertSupportedAddress);
await Promise.all(assertionPromises);

// Or, identical, just merged into one line
await Promise.all(recipients.map(assertSupportedAddress))

// Also identical, maybe more clear if not super familiar with callbacks
const assertionPromises = recipients.map(recipient => assertSupportedAddress(recipient));
await Promise.all(assertionPromises);
```

then the `map` call dispatches the async calls for all promises at (roughly) the same time, instead of sending each one and waiting for a response one by one in a loop. Then we just wait for all of them to resolve with `Promise.all`. Because all async calls happen in parallel, and we wait for them all in parallel, it's much faster. 

--------

Same concept for the below block. Let's remove all `await`s in a loop here. The main reason for having `await`s in a loop is when e.g. iteration 2 needs the result of iteration 1, but that's not the case here

https://github.com/ScopeLift/umbra-protocol/blob/1b968b7f98e2a8dc3a68e91dbe806bde65b17568/umbra-js/src/classes/Umbra.ts#L231-L234

umbra-js/src/classes/Umbra.ts
```diff
@@ -560,6 +633,26 @@ export class Umbra {
...
   }
 
+  /**
+   * @notice Helper method to return parsed localOverrides and LookupOverrides
+   * @param overrides Override the gas limit, gas price, nonce, or advanced mode.
+   * When `advanced` is false it looks for public keys in StealthKeyRegistry, and when true it recovers
+   * them from on-chain transaction when true
+   */
+  static parseOverrides(overrides: SendOverrides = {}) {
+    const localOverrides = { ...overrides }; // avoid mutating the object passed in
```
mds1, line 643: We should add some tests for this method to ensure that things are not mutated

mds1: Hmm although then do we need to export this method? That's kind of annoying, since it would be nice to keep this not exported 🤔

umbra-js/src/classes/Umbra.ts
mds1: Do you still get errors now that it's removed? If so what is the specific error / how can I reproduce?

umbra-js/src/classes/Umbra.ts
garyghayrat: Awesome! Thanks so much for the explanation, I'm learning so much here. Does this look good? 66133c10cc602fd3cced862974f3b89b05211f18

mds1: Yes! Looks really good. Just one more `await` in a loop when you call `this.prepareSend`

garyghayrat: How's this looking? https://github.com/ScopeLift/umbra-protocol/pull/426/commits/fe2a711959650aa19f34b01312093f022dbcf740

umbra-js/src/classes/Umbra.ts
garyghayrat: It seems to be working without the manual gas limit now, on branch `frontend-batch-send-demo`

contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
garyghayrat: Awesome, I'm starting to become a fan of type safety. But removed them for now!

contracts-core/periphery-typechain/UmbraBatchSend.ts
garyghayrat: Removed for now!

contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
mds1: Type safety is 1000% better than no type safety 😛 but yea if it's causing problems or being weird let's get the core part working first and we can always follow up in a separate PR. Leaving this unresolved until we either decide to implement it in this PR at the end or pull out into a separate issue

umbra-js/src/classes/Umbra.ts
mds1: Ok going to mark this resolved then

contracts-periphery/package.json
```diff
@@ -2,11 +2,20 @@
...
+  ],
   "scripts": {
     "install": "make install",
-    "build": "make build",
+    "build": "make build && yarn run typechain --target ethers-v5 --out-dir ./typechain './out/UmbraBatchSend.sol/UmbraBatchSend.json'",
     "test": "make test",
     "lint": "forge fmt --check",
-    "coverage": "make coverage"
+    "coverage": "make coverage",
+    "prepare": "yarn run typechain --target ethers-v5 --out-dir ./typechain './out/UmbraBatchSend.sol/UmbraBatchSend.json'"
```
mds1, line 14: If we do postpone typechain integration, let's delete all the new typechain stuff added to this file

umbra-js/src/classes/Umbra.ts
```diff
@@ -235,6 +208,81 @@ export class Umbra {
...
+      }
+
+      sendData.push({
+        receiver: receiver,
+        tokenAddr: token,
+        amount: amount,
+        pkx: info.pubKeyXCoordinate,
+        ciphertext: info.encrypted.ciphertext,
+      });
+    });
```
mds1, line 269: The use case of `map` is to return a new array generated from the old array, typically without side effects. In this case you're not using the return value, and have side effects by calling `sendData.push`.

If you want to keep the side effect approach, using `forEach` instead of `map` would be more semantically correct.

However, a better approach is to directly create and assign `sendData` using the return value from `map`:

```typescript
    const sendData: SendData[] = sendInfo.map((info, index) => {
      const receiver = info.stealthKeyPair.address;
      const token = sends[index].token;
      const amount = BigNumber.from(sends[index].amount);
      assertValidStealthAddress(receiver);

      // Sum valueAmount
      valueAmount = valueAmount.add(toll);
      if (isEth(token)) {
        valueAmount = valueAmount.add(amount);
      }

      return {
        receiver: receiver,
        tokenAddr: token,
        amount: amount,
        pkx: info.pubKeyXCoordinate,
        ciphertext: info.encrypted.ciphertext,
      };
    });
```

Additionally, I don't think we need to iteratively compute `valueAmount`, so that side effect can be removed: our `tokenSum` Map will have a key for the ETH address (or not, in which case start from zero), then sum that value with `toll * sendInfo.length`

mds1: And actually, why do we need one `.map` to generate `sendInfo` and another one here? Seems those can just be merged into a single `.map` call. And you can generate the `stealthKeyPairs` array afterwards to return it

contracts-periphery/package.json
```diff
@@ -2,11 +2,20 @@
   "name": "@umbra/contracts-periphery",
   "version": "0.0.1",
   "license": "MIT",
+  "files": [
```
mds1, line 5: _(arbitrary comment location to start a thread)_

Make sure to resolve conflicts and get CI passing too 🙂

umbra-js/src/classes/Umbra.ts
garyghayrat: @mds1 So your refactoring comment made sense, but I don't see an obvious way to combine the `sendInfo`' `.map` with `sendData`' `.map` here without losing `Promise.all`, which leads me to something like this:
```
    let stealthKeyPairs: KeyPair[] = [];
    let sendData: SendData[] = [];

    sends.map(async (send) => {
      const { stealthKeyPair, ...info } = await this.prepareSend(send.address, lookupOverrides);
      assertValidStealthAddress(stealthKeyPair.address);

      stealthKeyPairs.push(stealthKeyPair);
      sendData.push({
        receiver: stealthKeyPair.address,
        tokenAddr: send.token,
        amount: BigNumber.from(send.amount),
        pkx: info.pubKeyXCoordinate,
        ciphertext: info.encrypted.ciphertext,
      });
    });
```
Is there a way to combine the two `.maps` without losing `Promise.all`?

contracts-periphery/package.json
garyghayrat: Removed https://github.com/ScopeLift/umbra-protocol/pull/426/commits/b68763f63027e9a50cb4e4d4f4c112a7d72268fe

umbra-js/src/classes/Umbra.ts
mds1: Does something like this work? (Been a while since I looked at this so a bit hazy on all the discussion here 😅)

```typescript
const sendData = await Promise.all(
  sends.map(async (send) => {
    const sendInfo = await this.prepareSend(send.address, lookupOverrides);
    const receiver = sendInfo.stealthKeyPair.address;
    const token = send.token;
    const amount = BigNumber.from(send.amount);
    assertValidStealthAddress(receiver);

    return {
      receiver: receiver,
      tokenAddr: token,
      amount: amount,
      pkx: sendInfo.pubKeyXCoordinate,
      ciphertext: sendInfo.encrypted.ciphertext,
    };
  })
);
```

contracts-core/periphery-artifacts/UmbraBatchSend.sol/UmbraBatchSend.dbg.json
garyghayrat: Created a separate issue for it! https://github.com/ScopeLift/umbra-protocol/issues/457

umbra-js/src/classes/Umbra.ts
garyghayrat: Exported and tested here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/4003eedc816f957152059f7adf6c81ae1665fd65 Maybe once you think the test is solid we can remove the test and the export.

contracts-periphery/package.json
garyghayrat: CI should be passing now

umbra-js/src/abi/UmbraBatchSend.sol/UmbraBatchSend.json
```diff
@@ -0,0 +1,4223 @@
+{
+  "abi": [
```
mds1, line 2: This is the full output artifact, which includes wayyy more data than we need which is why it's so big. Let's only include the `abi` section.

Even better would be to take the abi and convert it to "human readable form" in `constants.ts`, like we do for other ABIs in that file. You can do this by creating an ethers.js interface from the ABI and and then using the interface's `format` method. Check out the [docs](https://docs.ethers.org/v5/single-page/#/v5/api/utils/abi/interface/) for more info

garyghayrat: Awesome! Addressed here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/f09f9b371633493016d6577c2fea0b2e02ab8d16

mds1: Can we move it to `constants.ts` where the other non-Umbra.sol ABIs are? In a separate PR we should move the Umbra ABI declaration there too.

Also, CI is failing which it seems is because we still try importing the old (non-existent) ABI file in tests instead of using the new ABI declaration

umbra-js/test/Umbra.test.ts
```diff
@@ -343,6 +372,379 @@ describe('Umbra class', () => {
     });
   });
 
+  describe('BatchSend, scan, and withdraw funds', () => {
```
mds1, line 375: It looks like this is *a lot* of duplication from the `describe('Send, scan, and withdraw funds'` block of tests. But we don't need to re-test scanning and withdrawing, so let's simplify things significantly and only test the new functionality here: the `batchSend` method. Duplicating this much code makes it hard to know where the important new code/tests are vs. the duplicated code/tests that aren't really adding value

umbra-js/test/Umbra.test.ts
```diff
@@ -446,4 +848,49 @@ describe('Umbra class', () => {
       );
     });
   });
+
+  describe('ParseOverrides without mutation', () => {
```
mds1, line 852: We can shorten the name here to just be the method under test, then the `it` statements function as its spec
```suggestion
  describe('parseOverrides', () => {
```

umbra-js/src/classes/Umbra.ts
mds1: Ok actually we do not want to do this, because it'd require returning `stealthKeyPair` from this map, but then we'd need to remove that item later to get the right the for the `batchSend` call, so going to mark this resolved

umbra-js/src/abi/UmbraBatchSend.sol/UmbraBatchSend.json
garyghayrat: Addressed by your PR: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/8dae74b4730f11ba3d34204707f81ed9742bafa7#diff-4425693d92f00c46b6f05b597ec8880c030e8575db36e43e23d26e956e573901

umbra-js/test/Umbra.test.ts
garyghayrat: Addressed here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/d84ba7d45dcfcd86862d083ce96644cd079dbbd7

umbra-js/test/Umbra.test.ts
garyghayrat: Added batchSend tests this way:
https://github.com/ScopeLift/umbra-protocol/pull/426/commits/e9a1c6dce86c129374bc9161bae7dfc219f17e7d
https://github.com/ScopeLift/umbra-protocol/pull/426/commits/18829bbc7123331418d3ab4ca33733d129f850d3
Does this work?

umbra-js/src/classes/Umbra.ts
```diff
@@ -8,7 +8,6 @@ import {
   BigNumber,
```
garyghayrat, line 8: (_random placement_)

CI coverage is now super low for `Umbra.ts`, not sure why?

umbra-js/test/Umbra.test.ts
```diff
@@ -446,4 +552,97 @@ describe('Umbra class', () => {
       );
     });
   });
+
+  describe('prepareSend', () => {
+    it('should return a valid result when public keys are found', async () => {
```
garyghayrat, line 557: is it sufficient to just check if the properties exist?

umbra-js/src/classes/Umbra.ts
mds1: > Maybe once you think the test is solid we can remove the test and the export.

I think it's ok to keep the export and just add a comment explaining why (i.e. for tests only, not intended to be removed). If you remove the test/export after testing, then you need to re-write a test and do that process any time that method is changed, which partly defeats the purpose of having tests

umbra-js/test/Umbra.test.ts
mds1: Yea since we have e2e tests that ensure you can withdraw sent funds, that implicitly tests that the return data here is correct, since otherwise it's hard to unit test that, so this is ok. Though I'd suggest adding that previous sentence as a comment explaining why the tests for this method are so bare

umbra-js/test/Umbra.test.ts
```diff
@@ -1,20 +1,23 @@
...
   Umbra as UmbraContract,
   TestTokenFactory as ERC20__factory,
   UmbraFactory as Umbra__factory,
 } from '@umbra/contracts-core/typechain';
-
+import { parseOverrides, assertSufficientBalance, assertValidStealthAddress } from '../src/classes/Umbra';
+import { UMBRA_BATCH_SEND_ABI } from '../src/utils/constants';
+import { KeyPair } from '../src';
+import { blockedStealthAddresses } from '../src/utils/utils';
 const { parseEther } = ethers.utils;
```
mds1, line 21: Nit, let's separate imports from variables with a line break
```suggestion
import { blockedStealthAddresses } from '../src/utils/utils';

const { parseEther } = ethers.utils;
```

umbra-js/test/Umbra.test.ts
```diff
@@ -54,6 +58,11 @@ describe('Umbra class', () => {
     sender.connect(ethers.provider);
     receiver = ethers.Wallet.fromMnemonic(mnemonic as string, `${path as string}/${receiverIndex}`);
     receiver.connect(ethers.provider);
+    receivers.push(receiver);
+    for (let i = 4; i < 10; i++) {
```
mds1, line 62: Let's just pick a large number to ensure it never conflicts with the lower indicies, in case we ever bump those or add another account. These integers don't matter, so can just make the loop go from 100 to 106 or something

umbra-js/test/Umbra.test.ts
```diff
@@ -54,6 +58,11 @@ describe('Umbra class', () => {
...
+      receivers.push(ethers.Wallet.fromMnemonic(mnemonic as string, `${path as string}/${i}`));
+      receivers[i - 3].connect(ethers.provider);
```
mds1, line 64: Actually you can just start the loop at 0 to make this indexing simpler, then in the derivation path just use `i + 100`, which is probably the most readable IMO

umbra-js/test/Umbra.test.ts
```diff
@@ -27,9 +30,10 @@ const overrides = { supportPubKey: true }; // we directly enter a pubkey in thes
 const senderIndex = 2;
 const receiverIndex = 3;
 
-describe('Umbra class', () => {
+describe.only('Umbra class', () => {
   let sender: Wallet;
   let receiver: Wallet;
+  let receivers: Wallet[] = [];
```
mds1, line 36: Let's add comments explaining why we have two variables here. Intuitively you'd think these are redundant, so if they're not let's just document why

umbra-js/test/Umbra.test.ts
```diff
@@ -73,11 +82,23 @@ describe('Umbra class', () => {
...
+      deployer
+    );
+    const batchSendContract = await batchSendFactory.deploy(umbraContract.address);
+    await batchSendContract.deployTransaction.wait();
+    // Approve DAI token
```
mds1, line 93:
```suggestion
    await batchSendContract.deployTransaction.wait();

    // Approve DAI token
```

umbra-js/test/Umbra.test.ts
```diff
@@ -176,172 +206,248 @@ describe('Umbra class', () => {
...
+        if (test.id === 'send') {
+          const result = await umbra.send(sender, dai.address, quantity, receiver!.publicKey, overrides);
+          tx = result.tx;
+          stealthKeyPairs = [result.stealthKeyPair];
+          usedReceivers = receivers.slice(0, 1);
+        } else if (test.id === 'batchSend') {
+          const sends: SendBatch[] = [];
+          for (let i = 0; i < 5; i++) {
+            sends.push({ token: dai.address, amount: quantity, address: receivers[i].publicKey });
+          }
```
mds1, line 270: Can we get some more entropy in here? Maybe make the quantity a function of `i` or something (just ensure it's never zero). Would be nice to mix up the tokens too if that's doable. You can imagine a bug in Umbra.ts where we accidentally overwrite data with a different array item's data, and this would not catch that since all entries are the same

umbra-js/test/Umbra.test.ts
```diff
@@ -176,172 +206,248 @@ describe('Umbra class', () => {
...
+          const result = await umbra.batchSend(sender, sends, overrides);
+          tx = result.tx;
+          stealthKeyPairs = result.stealthKeyPairs;
+          usedReceivers = receivers.slice(0, 5);
+        }
+        if (tx) await tx.wait();
```
mds1, line 276: We know `tx` will always be defined, so semantically this is more accurate

```suggestion
        await tx!.wait();
```

umbra-js/test/Umbra.test.ts
```diff
@@ -176,172 +206,248 @@ describe('Umbra class', () => {
...
+        // Send funds with Umbra
+        let tx: ContractTransaction | null = null;
+        let stealthKeyPairs: KeyPair[] = [];
+        let usedReceivers: Wallet[] = [];
+
+        if (test.id === 'send') {
+          const result = await umbra.send(sender, dai.address, quantity, receiver!.publicKey, overrides);
+          tx = result.tx;
+          stealthKeyPairs = [result.stealthKeyPair];
+          usedReceivers = receivers.slice(0, 1);
```
mds1, line 265: Hmm, I don't think this is correc? We send to `receiver` which uses the account at `receiverIndex`, but the account we save off in `usedReceivers` is from the `receivers` array which is a different index

umbra-js/test/Umbra.test.ts
```diff
@@ -176,172 +206,248 @@ describe('Umbra class', () => {
...
+
+        let tx: ContractTransaction | null = null;
+        let stealthKeyPairs: KeyPair[] = [];
+        let usedReceivers: Wallet[] = [];
+
+        if (test.id === 'send') {
+          const result = await umbra.send(sender, dai.address, quantity, receiver!.publicKey, overrides);
+          tx = result.tx;
+          stealthKeyPairs = [result.stealthKeyPair];
+          usedReceivers = receivers.slice(0, 1);
```
mds1, line 330: Similar comments to above for this test

umbra-js/test/Umbra.test.ts
mds1: Definitely looking better! Marking this resolved in favor of the next round of comments, we're almost finished here!

umbra-js/test/Umbra.test.ts
garyghayrat: I pushed `receiver` into the receivers index way above so `receivers[0]` is `receiver`, but I see it's confusing. So,

1. I've removed the push code above, so `receivers` are a completely different set of wallets.
2. Made `usedReceivers = [receiver]`, so it's more clear.

umbra-js/test/Umbra.test.ts
garyghayrat: Answered here: https://github.com/ScopeLift/umbra-protocol/pull/426#discussion_r1158327028

umbra-js/test/Umbra.test.ts
garyghayrat: Code : https://github.com/ScopeLift/umbra-protocol/pull/426/commits/ddc768e7504ccb601877993210dbfa445f644277

umbra-js/test/Umbra.test.ts
garyghayrat: Addressed here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/ddc768e7504ccb601877993210dbfa445f644277

umbra-js/test/Umbra.test.ts
garyghayrat: Made the changes here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/ddc768e7504ccb601877993210dbfa445f644277

umbra-js/test/Umbra.test.ts
garyghayrat: Addressed according the latest comment below: https://github.com/ScopeLift/umbra-protocol/pull/426#discussion_r1158332053

umbra-js/test/Umbra.test.ts
garyghayrat: Added the comment https://github.com/ScopeLift/umbra-protocol/commit/ddc768e7504ccb601877993210dbfa445f644277

umbra-js/test/Umbra.test.ts
garyghayrat: Also here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/3d3234a5ad319ef3bfa08cc8d8951239baa963d2

umbra-js/test/Umbra.test.ts
garyghayrat: Also noticed our `Send ETH, scan for it, withdraw it (direct withdraw)` is exactly the same as `Send ETH, scan for it, withdraw it (relayer withdraw)`. This is because we can't relay eth withdrawals I'm guessing? So should I remove the relayer withdraw test for eth?

mds1: > So should I remove the relayer withdraw test for eth?

Let's track in a separate issue, this PR is already pretty big and has been open for a while, so I don't want to increase it's scope further

umbra-js/src/classes/Umbra.ts
mds1: Hmm let's see, here's what I get on master:
image
image


And on this branch:
image
image

That's odd, it seems coverage went down for things that weren't even changed on this PR? 🤨

umbra-js/test/Umbra.test.ts
garyghayrat: Opened an issue https://github.com/ScopeLift/umbra-protocol/issues/485

umbra-js/src/classes/Umbra.ts
mds1: This is now resolved, the issue was a `.only` was accidentally committed which was removed in https://github.com/ScopeLift/umbra-protocol/pull/426/commits/93f3b34677c2160030cc414a5405d6d1a201f0a4

apbendi (approved):
:shipit:

umbra-js/src/classes/Umbra.ts
```diff
@@ -67,12 +43,12 @@ const subgraphs = {
...
-  10: { chainId: 10, umbraAddress, startBlock: 4069556, subgraphUrl: subgraphs[10] }, // Optimism
-  137: { chainId: 137, umbraAddress, startBlock: 20717318, subgraphUrl: subgraphs[137] }, // Polygon
-  1337: { chainId: 1337, umbraAddress, startBlock: 8505089, subgraphUrl: false }, // Local
-  42161: { chainId: 42161, umbraAddress, startBlock: 7285883, subgraphUrl: subgraphs[42161] }, // Arbitrum
+  1: { chainId: 1, umbraAddress, batchSendAddress, startBlock: 12343914, subgraphUrl: subgraphs[1] }, // Mainnet
+  5: { chainId: 5, umbraAddress, batchSendAddress, startBlock: 7718444, subgraphUrl: subgraphs[5] }, // Goerli
+  10: { chainId: 10, umbraAddress, batchSendAddress, startBlock: 4069556, subgraphUrl: subgraphs[10] }, // Optimism
+  137: { chainId: 137, umbraAddress, batchSendAddress, startBlock: 20717318, subgraphUrl: subgraphs[137] }, // Polygon
+  1337: { chainId: 1337, umbraAddress, batchSendAddress, startBlock: 8505089, subgraphUrl: false }, // Local
+  42161: { chainId: 42161, umbraAddress, batchSendAddress, startBlock: 7285883, subgraphUrl: subgraphs[42161] }, // Arbitrum
```
mds1, line 51: We define `batchSendAddress = '0x0d81Df222BB44b883265538586829715CF157163'`, and it's only deployed on goerli, but we're using it as the address for all chains. To avoid sending transactions that won't revert to an address with no code (and potentially losing funds if ETH is sent), we should put `null` for the chains that don't have this contract deployed

umbra-js/src/classes/Umbra.ts
```diff
@@ -151,7 +134,8 @@ export class Umbra {
    */
   constructor(readonly provider: EthersProvider, chainConfig: ChainConfig | number) {
     this.chainConfig = parseChainConfig(chainConfig);
-    this.umbraContract = new Contract(this.chainConfig.umbraAddress, umbraAbi, provider) as UmbraContract;
+    this.umbraContract = new Contract(this.chainConfig.umbraAddress, UMBRA_ABI, provider) as UmbraContract;
+    this.batchSendContract = new Contract(this.chainConfig.batchSendAddress, UMBRA_BATCH_SEND_ABI, provider);
```
mds1, line 138: Similarly, this should be not initialized/undefined for chains that don't have the contract deployed

umbra-js/src/classes/Umbra.ts
```diff
@@ -721,3 +750,46 @@ async function tryEthWithdraw(
...
+export function assertValidStealthAddress(stealthAddress: string) {
+  // Ensure that the stealthKeyPair's address is not on the block list
+  if (blockedStealthAddresses.includes(stealthAddress)) throw new Error(`Invalid stealth address: ${stealthAddress}`);
+}
+
+/**
+ * @notice Helper method to return parsed localOverrides and LookupOverrides
+ * @param overrides Override the gas limit, gas price, nonce, or advanced mode.
+ * When `advanced` is false it looks for public keys in StealthKeyRegistry, and when true it recovers
+ * them from on-chain transaction when true
```
mds1, line 778: I think we said we were going to add a comment here explaining why this function is exported (for testing only) and that it should only be used within this file

garyghayrat: Added the comment https://github.com/ScopeLift/umbra-protocol/pull/426/commits/c6ae57a55c8671619e38b1dd202e691fe4730098

umbra-js/src/classes/Umbra.ts
garyghayrat: Initialize only if `batchSendAddress` exists on `chainConfig` https://github.com/ScopeLift/umbra-protocol/pull/426/commits/c6ae57a55c8671619e38b1dd202e691fe4730098

umbra-js/src/classes/Umbra.ts
garyghayrat: Removed the address from all other chains https://github.com/ScopeLift/umbra-protocol/pull/426/commits/c6ae57a55c8671619e38b1dd202e691fe4730098

umbra-js/src/classes/Umbra.ts
mds1: It looks like our `batchSend` method now just silently becomes a no-op that doesn't actually send funds if the batch send contract doesn't exist? That feels like confusing behavior. We should instead throw an error if you try to use `batchSend` on a chain that doesn't have it
image

mds1: Similarly in the `parseChainConfig`, we don't need two whole if blocks like this:

image


Just do this:

```typescript
return {
      umbraAddress: getAddress(umbraAddress),
      batchSendAddress: batchSendAddress ? getAddress(batchSendAddress) : null,
      startBlock,
      chainId,
      subgraphUrl,
    };
```

garyghayrat: Aww man, I did it in such a roundabout way... Thank you! Addressed here: https://github.com/ScopeLift/umbra-protocol/pull/426/commits/961a34a4c697c5ca6dfce7c231990a6c95e2547f

apbendi:
I've tested this extensively looking for regressions and things seem to be working well. What do you think of moving forward using a phased rollout @mds1?

mds1 (approved):

mds1:
@apbendi phased rollout sounds perfect 👌

PR #428, opened 2022-12-20
Upgrade i18n syntax

garyghayrat:
- Fixes the errors
-  But getting `Detected HTML in <i18n key message> Recommend not using HTML messages to avoid XSS.` Which should be fine because we're hard-coding the content and not taking user inputs. And it will be a bit tedious to extract HTML code from i18n message to templates. 
- So we probably should merge this fix, then later open another if we want to get rid of the HTML.


frontend/src/components/AccountReceiveTableWarning.vue
```diff
@@ -11,10 +11,10 @@
     </q-card-section>
 
     <q-card-section>
-      <i18n path="AccountReceiveTableWarning.withdrawal-warning" tag="p">
+      <i18n-t keypath="AccountReceiveTableWarning.withdrawal-warning" tag="p">
```
mds1, line 14: How does this fix work, i.e. (1) what is the difference between the `i18n` and `i18n-t` components, and (2) what is the `path` vs. `keypath` change? Are these just breaking changes from upgrading, or something different?

frontend/src/i18n/locales/en-US.json
```diff
@@ -168,24 +167,24 @@
     "audit-answer-contract": "The {contracts} have been audited by ConsenSys Diligence, and the audit report can be found {here}.",
     "audit-contracts": "contracts",
     "audit-here": "here",
-    "audit-answer-umrba-js": "The {umbra-js} {library} responsible for handling the required off-chain logic and elliptic curve operations—has been audited by Least Authority, and the audit report can be found {here}.",
+    "audit-answer-umrba-js": "The {umbrajs} {library} responsible for handling the required off-chain logic and elliptic curve operations—has been audited by Least Authority, and the audit report can be found {here}.",
```
mds1, line 170: Probably a few other places to fix this typo too
```suggestion
    "audit-answer-umbra-js": "The {umbrajs} {library} responsible for handling the required off-chain logic and elliptic curve operations—has been audited by Least Authority, and the audit report can be found {here}.",
```

mds1 (commented):
Thanks!

frontend/src/components/AccountReceiveTableWarning.vue
garyghayrat: Yes, `i18n-t` and `keypath` are breaking change renames for i18n v9 or later https://vue-i18n.intlify.dev/guide/migration/breaking.html#translation-component

frontend/src/i18n/locales/en-US.json
garyghayrat: From testing it seems like dashes aren't allowed in key's contents like the 

> "The {umbrajs} {library} responsible...

But dashes in key names are still allowed.

> "audit-answer-umbra-js"

So left the rest as is.

mds1: Yea that's fine! The suggestion is that they key name is "audit-answer-**umrba**-js" instead of "audit-answer-**umbra**-js" 😅

garyghayrat: oh shoot, totally missed that. My bad.

garyghayrat: Fixed it. Thank you!

mds1: Thanks!

mds1 (approved):
LGTM, thanks!

PR #430, opened 2022-12-20
Fix scanning with private key issue

garyghayrat:
https://user-images.githubusercontent.com/61768337/208781387-0895ae5d-2555-4612-9d0f-ed387e355163.mov

### DIFF
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -74,7 +74,7 @@
 
       <!-- Scanning complete -->
       <div v-else-if="scanStatus === 'complete'" class="text-center">
-        <account-receive-table :announcements="userAnnouncements" @reset="setFormStatus('waiting', '')" />
+        <account-receive-table :announcements="userAnnouncements" @reset="setFormStatus('waiting')" />
       </div>
     </div>
   </q-page>
@@ -124,11 +124,9 @@ function useScan() {
   };
   const isValidStartBlock = (val: string) => !val || Number(val) > 0 || 'Please enter a valid start block';
   const isValidEndBlock = (val: string) => !val || Number(val) > 0 || 'Please enter a valid start block';
-  const setFormStatus = (scanStatusVal: ScanStatus, scanPrivateKey: string) => {
+  const setFormStatus = (scanStatusVal: ScanStatus) => {
     scanStatus.value = scanStatusVal;
-    if (scanPrivateKey.length === 64) scanPrivateKey = `0x${scanPrivateKey}`;
-    setScanPrivateKey(scanPrivateKey);
-    scanPrivateKeyLocal.value = scanPrivateKey;
+    setScanPrivateKey('');
   };
 
   onMounted(async () => {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/AccountReceive.vue
```diff
@@ -164,6 +164,7 @@ function useScan() {
 
     // Save off the scanPrivateKeyLocal to memory if it exists, then scan
     if (scanPrivateKeyLocal.value) {
+      if (scanPrivateKeyLocal.value.length === 64) scanPrivateKeyLocal.value = `0x${scanPrivateKeyLocal.value}`;
```
mds1, line 167: This feature (supporting private keys with and without 0x prefix) was originally supported, but we broke it when adding the below `await utils.assertSupportedAddress(computeAddress(scanPrivateKeyLocal.value));` check. This breaks it because `computeAddress` is an ethers.js method that wants the private key to have the 0x prefix.

Previously, we made this 0x change 2 lines below within the `setScanPrivateKey`.

I'd suggest leaving this line as-is, then also:
- Update `setFormStatus` to remove the private key input, since it seems we only call setFormStatus to clear the custom private key anyway
- Update `setScanPrivateKey` to remove the current ``if (key.length === 64) key = `0x${key}` `` check since we now handle that here
- Update `setScanPrivateKey` to validate that the string is either (1) empty, i.e. `''`, or (2) a 66 character hex string

garyghayrat: > Update setScanPrivateKey to validate that the string is either (1) **empty**

Why do we allow it to be empty?

mds1: For when you no longer want to use a custom private key!

garyghayrat: Gotcha, thank you for the thorough explanation!

frontend/src/store/settings.ts
```diff
@@ -83,7 +83,7 @@ export default function useSettingsStore() {
   }
 
   function setScanPrivateKey(key: string) {
-    if (key.length === 64) key = `0x${key}`;
+    if (key !== '' && key.length !== 66) throw new Error(`Invalid private key '${key}' ${key.length}`);
```
mds1, line 86: If key length is 66, let's make sure `isHexString(key)` returns true (we should have already been doing this). Also I'd suggest tweaking the error message
```suggestion
    if (key !== '' && key.length !== 66) throw new Error(`Invalid private key '${key}'`);
```

garyghayrat: Added!

garyghayrat: Oh actually it seems like we already check for `isHexString` in `isValidPrivateKey`

garyghayrat: Guess it doesn't hurt to double check if we ever decide to use `setScanPrivateKey` elsewhere.

mds1: Good point—it's probably redundant, but can't hurt in case the QInput validation fails / is accidentally broken in the future

I just pushed 2fd25a9 to fix the check / make it more readable (from my quick test the old version seemed incorrect). Give that a quick review/UI test, and if it works we can merge!

garyghayrat: Definitely more readable. The change is working as expected!

mds1 (approved):

apbendi (approved):
lgtm nice work @garyghayrat

PR #406, opened 2022-10-18
Properly package umbra-js and make it easy to publish

wildmolasses:
resolves #134 

todo:
- make it easier to publish (.env token?)
- squash commits (v0.1.4?? huh?)
- try to maybe get the src folder out of the build folder (but keep the ./src folder where it is... yeah)

PR #432, opened 2023-01-04
chore: remove ABI import from contracts-core directory

mds1:
Facilitates https://github.com/ScopeLift/umbra-protocol/pull/406, since the import from the contracts-core directory was causing some issues and the `Umbra.json` file wasn't published as part of the build

cc @wildmolasses @jferas @apbendi


jferas (approved):
LGTM

PR #433, opened 2023-01-04
Quasar v2 Fix Sends and Setup

davidlaprade:
Note: I think there's one more place we need makeRaw -- namely [here](https://github.com/ScopeLift/umbra-protocol/blob/1fe4ef016b687d6b6c45e7487e41abca105c5e59/frontend/src/store/wallet.ts#L344-L345) -- but I haven't been able to verify yet

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -1,4 +1,4 @@
-import { computed, onMounted, ref, watch } from 'vue';
+import { computed, onMounted, markRaw, ref, watch } from 'vue';
 
 import Onboard, { OnboardAPI } from '@web3-onboard/core';
 import injectedModule from '@web3-onboard/injected-wallets';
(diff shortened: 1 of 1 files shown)


frontend/src/store/wallet.ts
```diff
@@ -265,8 +265,11 @@ export default function useWalletStore() {
       }
 
       // Set Umbra and StealthKeyRegistry classes
-      umbra.value = new Umbra(provider.value, newChainId);
-      stealthKeyRegistry.value = new StealthKeyRegistry(signer.value);
+      // We have to use markRaw here to prevent Vue from making our ethers objects into proxies:
+      //   https://github.com/vuejs/core/issues/3024
+      //   https://github.com/dcgtc/dgrants/blob/2732f3107fd497b8c7f6ba7a13fcfb9f9f0e89e9/app/src/store/wallet.ts#L8-L20
```
mds1, line 270: Nit, but let's just copy that good dgrants comment into the top of this file the same way (or tweak it however you see fit). But since that repo is dead, just want to make sure the info is preserved here in case that repo disappears one day

frontend/src/store/wallet.ts
```diff
@@ -1,4 +1,4 @@
-import { computed, onMounted, ref, watch } from 'vue';
+import { computed, onMounted, markRaw, ref, watch } from 'vue';
```
mds1, line 1: _(arbitrary comment location to start a thread)_

> Note: I think there's one more place we need makeRaw -- namely [here](https://github.com/ScopeLift/umbra-protocol/blob/1fe4ef016b687d6b6c45e7487e41abca105c5e59/frontend/src/store/wallet.ts#L344-L345) -- but I haven't been able to verify yet

About to look into this, will resolve this conversation afterwards

mds1: Ok, I was able to execute the linked code successfully without `markRaw`, so I guess we don't need it there. Let's merge this PR as the only `markRaw` for now, then add more as-needed when that error pops up. Leaving this comment unresolved so @davidlaprade sees it

frontend/src/store/wallet.ts
davidlaprade: Good idea, done in a58bf34

mds1 (approved):

PR #435, opened 2023-01-09
Remover worker-loader package for quasar-v2

mds1:
Quasar v2 uses webpack 5, which does need the worker loader package: https://webpack.js.org/guides/web-workers/

Scan times remained unchanged


frontend/tsconfig.json
```diff
@@ -8,6 +8,8 @@
     "outDir": "build",
     "resolveJsonModule": true,
     "skipLibCheck": true,
+    "module": "es2020",
```
mds1, line 11: This is needed to use `import.meta`, which is what's used in the [webpack docs](https://webpack.js.org/guides/web-workers/). Without this you get an error of `TS1343: The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node12', or 'nodenext'.`

wildmolasses: sounds good to me!

frontend/tsconfig.json
```diff
@@ -8,6 +8,8 @@
...
+    "moduleResolution": "node",
```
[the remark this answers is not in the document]
wildmolasses: hmm, maybe it has to do with the fact that `module` was added as a config prop? Anyway, since the errors are resolved after you added the `moduleResolution` prop, i'm feeling good about this diff!

PR #437, opened 2023-01-13
Batchsend Umbra.js

stratus4:


umbra-js/src/classes/Umbra.ts
```diff
@@ -244,28 +244,23 @@ export class Umbra {
...
 
     const stealthKeyPairs = sendInfo.map((info) => info.stealthKeyPair);
-    const sendData: SendData[] = [];
-    let valueAmount = BigNumber.from('0');
+    const valueAmount = toll
+      .mul(sendInfo.length)
+      .add(tokenSum.get('0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE') || BigNumber.from('0'));
 
-    sendInfo.map((info, index) => {
+    const sendData: SendData[] = sendInfo.map((info, index) => {
```
stratus4, line 251: @mds1 So your refactoring comment made sense, but I don't see an obvious way to combine the `sendInfo`' `.map` with `sendData`' `.map` here without losing `Promise.all`, which leads me to something like this:
```
    let stealthKeyPairs: KeyPair[] = [];
    let sendData: SendData[] = [];

    sends.map(async (send) => {
      const { stealthKeyPair, ...info } = await this.prepareSend(send.address, lookupOverrides);
      assertValidStealthAddress(stealthKeyPair.address);

      stealthKeyPairs.push(stealthKeyPair);
      sendData.push({
        receiver: stealthKeyPair.address,
        tokenAddr: send.token,
        amount: BigNumber.from(send.amount),
        pkx: info.pubKeyXCoordinate,
        ciphertext: info.encrypted.ciphertext,
      });
    });
```
Is there a way to combine the two `.maps` without losing `Promise.all`?

PR #438, opened 2023-01-18
feat: add hardcoded donation banner

mds1:
Recommend turning off whitespace for a cleaner diff, since everything was indented by being wrapped in a new div.

For now turning off this banner needs a new deploy, since we have auto deploys on merge to master setup this seems fine for now and we can improve it later if we need


apbendi (approved):
LGTM :shipit:

PR #450, opened 2023-02-07
chore/update noble secp256k1

### LINKED ISSUE #436: chore: update noble-secp256k1
We're still on v1.2.14 via [`noble-secp256k1`](https://www.npmjs.com/package/noble-secp256k1), which was released in October 2021. That version is now deprecated in favor of [`@noble/secp256k1`](https://www.npmjs.com/package/@noble/secp256k1) which is on version 1.7.1 as of a few days ago.

There are breaking changes to interfaces, so once #425 is merged we can do this next. Additionally. let's try out `utils.precompute` and see if that method can help us speed up any calculations. 

Will look into this more during the implementation, but my initial reaction is:
- Use bigint's for the [`PrivKey`](https://github.com/paulmillr/noble-secp256k1/blob/e125abdd2f42b2ad4cf5f4a1b7927d7737b7becf/index.ts#L89) type
- Use `Point` for the [`PubKey`](https://github.com/paulmillr/noble-secp256k1/blob/e125abdd2f42b2ad4cf5f4a1b7927d7737b7becf/index.ts#LL91) type

These seem like the safest options, and `Point` instead of hex strings avoids potential issues with a `0x` and/or `04` prefix. We can always use `Point.fromHex` to convert to strings as-needed

mds1:
Closes #436

Only changes the internals of umbra-js methods to accommodate the changes to the noble API. 

At first I started making lots of changes to tighten up our interfaces to be more strongly typed than just `string` everywhere by using the noble types. This ends up snowballing into a massive change by changing interfaces to lots of functions, which would also require frontend changes. That didn't seem necessary right now, so I bailed on that and kept it simple

Because no interfaces have changed and all tests pass I did not do any frontend testing, but perhaps we should to be safe.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,7 @@ module.exports = {
         '@typescript-eslint/no-var-requires': 'off',
         '@typescript-eslint/unbound-method': ['error', { ignoreStatic: true }],
         '@typescript-eslint/no-unused-vars': ['error', { ignoreRestSiblings: true }],
+        '@typescript-eslint/restrict-template-expressions': 'error',
         'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
       },
     },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@umbra/frontend",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Send and receive stealth payments with the Umbra protocol",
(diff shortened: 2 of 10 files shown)


garyghayrat (approved):
Tested Eth/token send/receive on Goerli and Arbitrum. 
Eth sends on mainnet also working as expected.

apbendi (approved):
Looks good! No bug reports from feature flag rollout.

PR #452, opened 2023-02-07
fix: input field button is always inside input field

### LINKED ISSUE #398: Max Button Should Be Inside Input On Mobile
In normal desktop mode, the "Max" button on the send screen appears inside the Amount input:

Screen Shot 2022-10-04 at 2 34 48 PM

However in the mobile layout, it appears as a separate button outside the input:

Screen Shot 2022-10-04 at 2 34 42 PM

I think the Max button should remain inside the input even in mobile mode. It's a small button and should be easy enough to tap even in a smaller context. Having a it as a separate button could create some UX confusion between "Max" and "Send".

mds1:
Closes #398

### DIFF
--- a/frontend/src/components/BaseInput.vue
+++ b/frontend/src/components/BaseInput.vue
@@ -33,8 +33,8 @@
       If we have a button, never show the loading slot because it makes the button jump left and right when the
       loading slot is shown / hidden
       -->
-      <template v-if="appendButtonLabel && !$q.screen.xs" v-slot:loading></template>
-      <template v-if="appendButtonLabel && !$q.screen.xs" v-slot:append>
+      <template v-if="appendButtonLabel" v-slot:loading></template>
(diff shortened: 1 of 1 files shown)


frontend/src/components/BaseInput.vue
```diff
@@ -120,12 +90,6 @@ export default defineComponent({
       default: undefined,
     },
 
-    counter: {
```
mds1, line 123: This `counter` prop was used way back when we only generated a 16 byte random number and allowed a 16 byte memo. The counter was used to show how much more text you can fit in the memo. Since that was removed because it reduced security a lot, we no longer use this `counter` prop anywhere, so I deleted it

garyghayrat (approved):
lgtm!

apbendi (approved):
Looks good and works well!

PR #453, opened 2023-02-07
feat: show token balances on send page dropdown

### LINKED ISSUE #294: On send page, show balance of each token in the dropdown
Each row in this dropdown should should the user's balance of that token

image

mds1:
Closes #294 

Style loosely mirrors what Uniswap does to show balances

image

### DIFF
--- a/frontend/src/components/BaseSelect.vue
+++ b/frontend/src/components/BaseSelect.vue
@@ -39,8 +39,12 @@
             <img class="horizontal-center" :src="scope.opt.logoURI" style="height: 1.5rem" />
           </q-item-section>
           <q-item-section>
-            <!-- eslint-disable-next-line vue/no-v-text-v-html-on-component -->
-            <q-item-label v-html="scope.opt[optionLabel]" />
+            <q-item-label>{{ scope.opt[optionLabel] }}</q-item-label>
+          </q-item-section>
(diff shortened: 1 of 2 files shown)


davidlaprade (commented):
Just tried it out locally and it worked well! Awesome 🎉 

This is a simpler design than we were going for [~6mo ago](https://github.com/ScopeLift/umbra-protocol/compare/ScopeLift:50bb314...ScopeLift:e92e8f2#), but you're right this is basically what Uniswap does. So it's fine with me.

One thing Uniswap does that we don't is sort tokens based on balance. I believe this is where the [difficulty/race conditions](https://github.com/ScopeLift/umbra-protocol/compare/ScopeLift:50bb314...ScopeLift:e92e8f2#) entered the picture before. So maybe it's not worth the effort.

But if it's not hard, I think it'd be nice to sort the `tokenList` such that:
* the native token is always first no matter what its balance is
* tokens are otherwise reverse-sorted by balance/10^decimals

mds1:
Sorting by balance is a nice idea. I think it's more important to sort that way in Uniswap where they support dozens/hundreds of tokens by default, whereas we only support <5 for now. Either way, tracking in https://github.com/ScopeLift/umbra-protocol/issues/454

apbendi (approved):
Looks good to me. Nice :+1:

PR #455, opened 2023-02-07
fix: native token always comes first

mds1:


frontend/src/store/wallet.ts
```diff
@@ -477,21 +477,20 @@ export default function useWalletStore() {
...
-      tokensArray = sortedTokens.sort((tok) => Number(tok.address != NATIVE_TOKEN_ADDRESS)); // move native token to front
+      const nativeToken = tokensArray.splice(nativeTokenIndex, 1)[0];
+      tokensArray = [nativeToken, ...tokensArray];
     } else {
-      // add native token to the front of the array
-      tokensArray = [NATIVE_TOKEN.value, ...sortedTokens];
+      tokensArray = [NATIVE_TOKEN.value, ...tokensArray];
     }
 
+    // Add native token to beginning of array.
```
apbendi, line 493: This seems like this comment is in the wrong place?

frontend/src/store/wallet.ts
```diff
@@ -477,21 +477,20 @@ export default function useWalletStore() {
   });
 
   const tokens = computed((): TokenInfoExtended[] => {
-    let tokensArray: TokenInfoExtended[] = [];
+    // Sort alphabetically.
+    const initialTokens = relayer.value?.tokens || [];
+    let tokensArray = initialTokens.sort((a, b) => a.symbol.localeCompare(b.symbol));
```
apbendi, line 482: RR: If I'm reading this PR right, we're sorting alphabetically, then splicing the native token out, then putting the native token back in at the front of the array. Could we simplify this by customizing the token sorting function here to return the Native Token ahead of other tokens?

frontend/src/store/wallet.ts
mds1: Oops, fixed in cdbe036

frontend/src/store/wallet.ts
mds1: > If I'm reading this PR right, we're sorting alphabetically, then splicing the native token out, then putting the native token back in at the front of the array.

Exactly

> Could we simplify this by customizing the token sorting function here to return the Native Token ahead of other tokens?

You definitely can do that, I'm not sure if it's actually simpler though, since then the flow is: check for native token, add ours if it's missing, then sort. So it's still 3 similar steps just in a different order

mds1: Clearer code introduced in 16927f5

apbendi: Much cleaner! Thank you for humoring my suggestion :sweat_smile:

apbendi (approved):
:shipit:

PR #462, opened 2023-02-16
Fix web3 onboard translation

### LINKED ISSUE #449: bug: translations are not working for web3-onboard connection modal
> Our i18n translations for Onboard are no longer working when Chinese is selected. Not sure if they dropped support or an update borked it. Mind taking a look? @alexkeating
> 
image

_Originally posted by @garyghayrat in https://github.com/ScopeLift/umbra-protocol/issues/418#issuecomment-1337887554_

@garyghayrat @alexkeating not sure which of you is a better fit for this one, feel free to self-assign

alexkeating:
#### Description

* We need to manually setLocale in the wallet when we want to change the translations.

Closes #449

### DIFF
--- a/frontend/src/layouts/BaseLayout.vue
+++ b/frontend/src/layouts/BaseLayout.vue
@@ -193,7 +193,7 @@
             <base-select
               class="language-selector"
               :modelValue="currentLanguage"
-              @update:modelValue="setLanguage"
+              @update:modelValue="changeLanguage"
               dense
               options-dense="true"
               :filled="false"
@@ -278,13 +278,27 @@ export default defineComponent({
   setup() {
     const { advancedMode, isDark, language, supportedLanguages, toggleAdvancedMode, setLanguage, toggleDarkMode } =
       useSettingsStore();
-    const { avatar, isAccountSetup, isAccountSetupLegacy, isArgent, isLoading, network, userAddress, userDisplayName } =
-      useWalletStore();
+    const {
+      avatar,
+      isAccountSetup,
+      isAccountSetupLegacy,
+      isArgent,
+      isLoading,
+      network,
+      userAddress,
+      userDisplayName,
+      setLanguage: setWalletLanguage,
+    } = useWalletStore();
 
     const currentLanguage = ref({ label: language.value.label, value: language.value.value });
     const argentModalDismissed = ref(false);
     const showArgentModal = computed(() => isArgent.value && !argentModalDismissed.value);
     const version = window.logger.version;
+
+    const changeLanguage = (language: { label: string; value: string }) => {
+      setLanguage(language);
+      setWalletLanguage(language.value);
+    };
     return {
       advancedMode,
       argentModalDismissed,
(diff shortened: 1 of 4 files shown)


frontend/src/layouts/BaseLayout.vue
```diff
@@ -278,13 +278,27 @@ export default defineComponent({
...
+      userDisplayName,
+      setLanguage: setWalletLanguage,
+    } = useWalletStore();
 
     const currentLanguage = ref({ label: language.value.label, value: language.value.value });
     const argentModalDismissed = ref(false);
     const showArgentModal = computed(() => isArgent.value && !argentModalDismissed.value);
     const version = window.logger.version;
+
+    const changeLanguage = (language: { label: string; value: string }) => {
```
alexkeating, line 298: Is there a place to reuse types? This type exists in settings.ts

frontend/src/store/wallet.ts
```diff
@@ -358,6 +358,11 @@ export default function useWalletStore() {
     setLastWallet(null);
   }
 
+  function setLanguage(language: string) {
+    if (!onboard.value) return;
+    onboard.value.state.actions.setLocale(language);
+  }
```
alexkeating, line 364: Are there better ways to compose this functionality with the settings setLanguage function? Are composables able to share state with other composables?

frontend/src/layouts/BaseLayout.vue
mds1: `models.ts` is where we have type definitions

frontend/src/store/wallet.ts
mds1: I'm not sure exactly what you mean by "composables", but I think what you're asking is how we can have this wallet.ts `setLanguage` method called as part of the settings.ts `setLanguage` method? I _think_ you might be able to just import this method in the settings store and call it there, or does vue given an error with that?

mds1: Oh wow I did not realize this was an actual term in the composition API world haha https://vuejs.org/guide/reusability/composables.html

alexkeating: When I tried calling useWalletStore in useSettings `setLanguage` I was getting a few errors. It seemed like there was a weird recursion, but I didn't dig too deep. So, I ended up moving away from it.

alexkeating:
@mds1 Is this good to merge now?

frontend/src/store/wallet.ts
mds1: This approach is probably fine for now, maybe just add a comment to the wallet `setLanguage` method documenting that it should never be called on it's own, and should always be called along with the settings.ts version

mds1:
@alexkeating sorry I wasn't working much the last few days. let's just add a clarifying comment per https://github.com/ScopeLift/umbra-protocol/pull/462#discussion_r1113275731 since that's not intuitive, then rebase against the latest master to fix CI, then we'll merge

frontend/src/store/wallet.ts
mds1: Done in e7a01d1, so will merge

mds1:
Added the comment in e7a01d1, will merge once CI passes

PR #465, opened 2023-02-19
Feature/send table UI

alexkeating:
#### Description

The non-cryptography changes for #263. We add a table with mocked send data and a link on the wallet pop out. Both these will be behind a `v-if` after approval.

[Screencast from 2023-02-19 13-42-24.webm](https://user-images.githubusercontent.com/12705146/219968413-6a215d7b-bfcc-4dcd-9ab8-b710e9a210c6.webm)


frontend/src/store/wallet.ts
```diff
@@ -522,6 +527,22 @@ export default function useWalletStore() {
...
+   * @notice Copies the address of type to the clipboard
+   */
+  async function copyAddress(address: string) {
+    if (!provider.value) return;
+    const mainAddress = await toAddress(address, provider.value);
+    await copyToClipboard(mainAddress);
+    notifyUser('success', `${tc('WalletRow.address-copied')}`);
+  }
+
+  function openInEtherscan(hash: string) {
```
mds1, line 540: This method and `copyAddress` above feel like they should be in `utils.ts` and take a `provider` input, since they aren't otherwise reliant on the wallet store

frontend/src/pages/AccountSent.vue
```diff
@@ -0,0 +1,79 @@
...
+    <div v-else class="q-mx-auto" style="max-width: 800px">
+      <account-send-table :sendMetadata="sendMetadata" />
+    </div>
+  </q-page>
+</template>
+
+<script lang="ts">
+import { defineComponent, ref } from 'vue';
+import useWallet from 'src/store/wallet';
+import useWalletStore from 'src/store/wallet';
```
mds1, line 22: These are the same import with different names, let's only keep the `useWalletStore` one

frontend/src/pages/AccountSent.vue
```diff
@@ -0,0 +1,79 @@
...
+        amount: '1000000000000000000',
+        tokenAddress: NATIVE_TOKEN_ADDRESS,
+        dateSent: new Date(),
+        address: '0x2436012a54c81f2F03e6E3D83090f3F5967bF1B5',
+        hash: '0xabeea5720640859db6dc46caae44fe34449469bfae8e3e930d9f59abbf50ed50',
+      },
+    ];
+
+    for (const row of data) {
+      formattedRows.push({
```
mds1, line 49: Wdyt about generating the `formattedRows` array with `data.map` and using that to initialize `sendMetadata`, and removing the try/catch since I don't think we really need that here? Would make this a bit more concise and remove the need to initialize an empty array

frontend/src/i18n/locales/en-US.json
```diff
@@ -286,6 +286,19 @@
...
     "acknowledge-risks": "I acknowledge the risks"
   },
+  "AccountSendTable": {
+    "account-empty": "No funds sent in this browser",
+    "address": "Recipient address",
+    "amount": "Amount",
+    "date-sent": "Date sent",
+    "receiver": "Receiver",
+    "received": "Received",
```
mds1, line 295: This key can be deleted and replaced with "date-sent" (we currently use the received key on mobile view)

frontend/src/i18n/locales/en-US.json
```diff
@@ -295,6 +308,7 @@
     "address-copied": "Address copied to clipboard",
     "connected-with": "Connected with",
     "copy-address": "Copy address",
+    "send-history": "Send history",
```
mds1, line 311:
```suggestion
    "view-send-history": "View send history",
```

cc @garyghayrat for translation

frontend/src/components/WalletRow.vue
```diff
@@ -42,32 +42,24 @@
         </div>
       </div>
     </div>
+    <div class="q-py-sm q-mt-md row justify-center items-center bg-grey-2 cursor">
+      <router-link :class="{ 'no-text-decoration': true, 'dark-toggle': true }" :to="{ name: 'sent' }">{{
+        $t('WalletRow.send-history')
+      }}</router-link>
+    </div>
```
mds1, line 49: A few comments here:
1. The whole thing looks like the button, but I can only click the text. Let's make the whole thing clickable
2. Let's add some change on hover, maybe the gold outline like we do when hovering the "Send" and "Receive" boxes

frontend/src/components/WalletRow.vue
```diff
@@ -160,10 +152,10 @@ export default defineComponent({
   margin-top: 40px
   position: absolute !important
   width: 290px
-  height: 150px
+  height: auto
   z-index: 1
   border-radius: 15px
-  padding: 15px
+
```
mds1, line 158: Nit, extra line break. Maybe would be removed by `yarn prettier`?

frontend/src/components/AccountSendTable.vue
```diff
@@ -0,0 +1,150 @@
...
+</template>
+
+<script lang="ts">
+import { defineComponent, PropType } from 'vue';
+import { SendTableMetadataRow } from 'components/models';
+import useWalletStore from 'src/store/wallet';
+import { tc } from 'src/boot/i18n';
+
+export default defineComponent({
+  name: 'AccountSendTable',
```
mds1, line 96: This file name and this should be `AccountSentTable` to be consistent with the `AccountSent` page (makes more sense too since the sends already occurred)

frontend/src/components/AccountSendTable.vue
```diff
@@ -0,0 +1,150 @@
+<template>
```
mds1, line 1: How much of this is similar to the `AccountReceiveTable`? Does it make sense to extract a `BaseSendsTable` component and reuse that in the `AccountReceiveTable` and `AccountSentTable` components?

mds1 (commented):
looks great overall!

frontend/src/i18n/locales/en-US.json
garyghayrat: 查看发送历史

garyghayrat: Fix here https://github.com/ScopeLift/umbra-protocol/pull/465/commits/9121397c9a0f7060b5265addbdd9fcb5d0cd9f41

frontend/src/i18n/locales/en-US.json
garyghayrat: deleted here https://github.com/ScopeLift/umbra-protocol/pull/465/commits/9121397c9a0f7060b5265addbdd9fcb5d0cd9f41

frontend/src/components/AccountSendTable.vue
alexkeating: I think it is similar but a pared down version. There are probably some things that can be abstracted

mds1: Should we track that in a separate issue or do it here? Hmm or maybe it's not worth abstracting since it's only 2 instances 🤔

alexkeating: I think a separate issue would make sense. And we abstract when the next use case comes

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -97,7 +97,7 @@
               <q-card-section>
                 <div class="row justify-between items-center">
                   <div>{{ $t('AccountReceiveTable.sender') }}</div>
-                  <div @click="copyAddress(props.row.from, 'Sender')" class="cursor-pointer copy-icon-parent">
+                  <div @click="copyAddress(props.row.from, provider, 'Sender')" class="cursor-pointer copy-icon-parent">
```
mds1, line 100: _(arbitrary comment location to start a thread)_

> I like the link in the wallet area, only thing I would say is it might also be good to include a link to it on the send form itself for better discoverability

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -118,7 +118,10 @@
                       </router-link>
                     </base-tooltip>
                   </div>
-                  <div @click="copyAddress(props.row.receiver, 'Receiver')" class="cursor-pointer copy-icon-parent">
+                  <div
```
mds1, line 121: _(arbitrary comment location to start a thread)_

> I think we should include some copy somewhere (maybe below the title, or maybe in an (i)-button mouseover tooltip thing) that explains the send history is only for this device, is stored locally, and can't be found by looking on-chain

alexkeating: Added an i tootip

frontend/src/components/AccountReceiveTable.vue
alexkeating: Added link to sent page on the send page

mds1: LGTM, might do them as two outline buttons side by side, but ok with doing that in a follow up PR to avoid churn/blocking this, as long as @apbendi approves

frontend/src/components/AccountReceiveTable.vue
mds1: Can we put it next to the "Send Funds" table header instead of the page header, i.e. the highlighted text here? 
image

alexkeating: Yes will move it now

alexkeating:
This should be good now, pending a successful build

mds1 (approved):
LGTM, pending @apbendi's review + feature flag to disable this before merging

apbendi (approved):
Looks good to me! Nice work @alexkeating

alexkeating:
Added feature flags

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -97,7 +97,7 @@
               <q-card-section>
```
garyghayrat, line 97: Getting this on the Send page 
image

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -97,7 +97,7 @@
...
                 <div class="row justify-between items-center">
```
garyghayrat, line 98: Getting these on the Receive page
image

garyghayrat: Above image is from Chrome on Goerli, this one is from Firefox on Goerli:
image

garyghayrat: I must be doing something wrong cause I don't even see the Send History button under my account 
image

frontend/src/components/AccountReceiveTable.vue
garyghayrat: image

frontend/src/components/AccountReceiveTable.vue
alexkeating: Yea, the routes change seems to be the issue. I am going to drop a message in Element

frontend/src/components/AccountReceiveTable.vue
garyghayrat: I'm no longer encountering this issue

frontend/src/components/AccountReceiveTable.vue
garyghayrat: This is still happening to me @alexkeating on the Receive page while on Chrome, using Metamask Goerli network.

garyghayrat: image

alexkeating: I think I fixed this

garyghayrat (approved):

PR #474, opened 2023-03-15
Include chain id as a param in the payment link

### LINKED ISSUE #439: feat: allow payment links to be network specific
> This is unrelated to this PR -- more like a feature request than anything -- but it'd be nice if a payment link could prompt a user to change networks, if necessary. If you're requesting payment from someone with a link, network matters. I don't want to receive (say) DAI on Polygon if I don't have Matic there. Currently, someone could be on mainnet, create the payment link to request DAI, then the recipient could be logged in on Polygon, and paste the link thinking it's fine to send on Polygon when actually it's not.

_Originally posted by @davidlaprade in https://github.com/ScopeLift/umbra-protocol/pull/425#discussion_r1081482876_

alexkeating:
#### Description

- Adds chainId as a param

[Screencast from 2023-03-15 00-32-55.webm](https://user-images.githubusercontent.com/12705146/225207327-868a2510-34da-424a-8234-b937d5fa55b2.webm)

Closes #439


apbendi (changes requested):
Hey @alexkeating, thanks for picking this one up! Overall it seems to be working, however when I use a payment link in a browser that has not yet connected the wallet, it seems to break. The spinner just runs forever, instead of asking for the wallet connection as it does normally. Check it out:

network-id-link-breaks-wallet-connect

alexkeating:
@apbendi Good catch, fixed and also fixed another bug which didn't allow the user to cancel the chain switch.

frontend/src/pages/AccountSend.vue
```diff
@@ -397,6 +399,18 @@ function useSendForm() {
 
     // Validate the form
     await sendFormRef.value?.validate();
+
+    // Switch chain
+    const chain = supportedChains.filter((chain) => chain.chainId === `0x${Number(linkChainId)?.toString(16)}`);
```
mds1, line 404: This is probably a bit more robust/safe, I don't really like trusting native JS type/format conversions for important things 😅
```suggestion
    const chain = supportedChains.filter((chain) => chain.chainId === BigNumber.from(linkChainId).toHexString;
```

frontend/src/pages/AccountSend.vue
```diff
@@ -397,6 +399,18 @@ function useSendForm() {
...
+    if (
+      chain.length === 1 &&
+      !isLoading.value &&
+      chainId.value !== Number(linkChainId) &&
+      !attemptedNetworkChange.value
+    ) {
+      attemptedNetworkChange.value = true;
+      await setNetwork(chain[0]);
```
mds1, line 412: Should we reset `attemptedNetworkChange` to false after the `await setNetwork` line? Also why exactly do we need the `attemptedNetworkChange` var?

frontend/src/utils/payment-links.ts
```diff
@@ -56,6 +58,7 @@ export async function generatePaymentLink({
   if (to) url.searchParams.set('to', to);
   if (token) url.searchParams.set('token', token.symbol);
   if (amount) url.searchParams.set('amount', amount);
+  if (chainId) url.searchParams.set('chainId', chainId.toString());
```
mds1, line 61: Similar to the other BigNumber comment
```suggestion
  if (chainId) url.searchParams.set('chainId', BigNumber.from(chainId).toString());
```

frontend/src/pages/AccountSend.vue
alexkeating: I ended up using hexValue instead of toHexString because it does not have leading 0s.

frontend/src/pages/AccountSend.vue
mds1: I see, thanks!

mds1 (approved):
LGTM, pending @apbendi's approval. Thanks!

apbendi (commented):
Hey @alexkeating, definitely working way better overall now. Nice fixes! One edge case I'm still seeing: if the wallet is not *unlocked* yet, the app seems to lose the link's query params when MetaMask prompts the user for a password. This means after the user unlocks the wallet, they won't see the populated send form as expected by whomever shared the link, but just the main app.

loses-link-context-when-wallet-not-unlocked

alexkeating:
I think I fixed that issue @apbendi

apbendi:
Hey @alexkeating, I'm still seeing the same behavior. When the wallet extension is not yet unlocked, when I visit a payment link, then click the "Connect" button, the URL params are cleared and as a result, after unlocking and connecting,  I don't end up on the send form. Same behavior as in the Gif above.

alexkeating:
> Hey @alexkeating, I'm still seeing the same behavior. When the wallet extension is not yet unlocked, when I visit a payment link, then click the "Connect" button, the URL params are cleared and as a result, after unlocking and connecting, I don't end up on the send form. Same behavior as in the Gif above.

Sorry for that, now the global connect should no longer redirect on the send page. Also, it may be good to create a ticket to explore other situations where we don't want to redirect to the home screen as the current default behavior is to always redirect to the home screen.

BaseLayout connect:

[Screencast from 2023-03-31 22-01-36.webm](https://user-images.githubusercontent.com/12705146/229260626-14a6b1b8-a7d5-402f-a0cd-532277ea10b3.webm)

Account send connect:

[Screencast from 2023-03-31 22-02-16.webm](https://user-images.githubusercontent.com/12705146/229260705-954e2f98-3f08-43ba-a4ce-24dde0127cac.webm)

apbendi (approved):
Looking good to me @alexkeating! Thanks for working through all those edge cases. I created issue #483 as you suggested [here](https://github.com/ScopeLift/umbra-protocol/pull/474#issuecomment-1492792726)

Issue #479, opened 2023-03-27
Sending Via Recipient's Pub Key Errors On Arbitrum


apbendi:
Sending via recipient's standard public key produces error on Arbitrum network. Must be in advanced mode, with "TO" field filled in, and check box selected as below:

<img width="1173" alt="Screen Shot 2023-03-27 at 12 02 29 PM" src="https://user-images.githubusercontent.com/1479802/227998147-120880b7-de99-45c7-bc31-39124f5d8c91.png">

PR #480, opened 2023-03-27
fix: handle tx type 120 on arbitrum (pre nitro tx)

mds1:
Closes https://github.com/ScopeLift/umbra-protocol/issues/479


apbendi (commented):
Hey @mds1, I'm still seeing a (slightly different?) error here:

image

apbendi (approved):
Looks good and working well for me :+1:

PR #488, opened 2023-04-07
fix: legacy pubkey lookup

mds1:
See commit messages for details

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -329,8 +329,8 @@ export default function useWalletStore() {
       const [_hasEnsKeys, _hasCnsKeys] = _isAccountSetup
         ? [false, false]
         : await Promise.all([
-            Boolean(_userEns) && (await hasSetPublicKeysLegacy(_userEns as string, provider.value)),
-            Boolean(_userCns) && (await hasSetPublicKeysLegacy(_userCns as string, provider.value)),
+            Boolean(_userEns) && (await hasSetPublicKeysLegacy(_userEns as string, MAINNET_PROVIDER as Web3Provider)),
+            Boolean(_userCns) && (await hasSetPublicKeysLegacy(_userCns as string, MAINNET_PROVIDER as Web3Provider)),
           ]);
 
       // Now we save the user's info to the store. We don't do this earlier because the UI is reactive based on these
--- a/frontend/src/utils/address.ts
+++ b/frontend/src/utils/address.ts
@@ -24,12 +24,6 @@ export const lookupAddress = async (address: string, provider: Provider) => {
   return domainName ? domainName : address;
 };
 
-// Returns an ENS or CNS name if found, otherwise returns the address
-export const lookupOrReturnAddress = async (address: string, provider: Provider) => {
-  const domainName = await lookupAddress(address, provider);
-  return domainName !== address ? domainName : address;
-};
-
 // Returns ENS name that address resolves to, or null if not found
 export const lookupEnsName = async (address: string, provider: Provider) => {
   try {


frontend/src/store/wallet.ts
```diff
@@ -329,8 +329,8 @@ export default function useWalletStore() {
       const [_hasEnsKeys, _hasCnsKeys] = _isAccountSetup
```
apbendi, line 329: *arbitrary comment location*

Was the intent here to remove the warning for legacy setup? I didn't think so, but that seems to be the effect:

Production:

image

Deploy Preview:

image

frontend/src/utils/address.ts
```diff
@@ -24,12 +24,6 @@ export const lookupAddress = async (address: string, provider: Provider) => {
   return domainName ? domainName : address;
 };
 
-// Returns an ENS or CNS name if found, otherwise returns the address
-export const lookupOrReturnAddress = async (address: string, provider: Provider) => {
```
apbendi, line 28: Was this just dead code?

mds1: Yep!

frontend/src/store/wallet.ts
mds1: Hmm, interesting. What's the status of the address you're logged in with? It has legacy keys setup on an ENS resolver but never registered in the StealthKeyRegistry?

mds1: Ok, investigating this. On main:
- _isAccountSetup: false
- _hasEnsKeys: true
- _hasCnsKeys: false

On this branch:
- isAccountSetup: false
- _hasEnsKeys: false
- _hasCnsKeys: false

Seems the ENS lookup throws on this branch with the below error, not yet sure why:

```
v2.0.1 WARNING - Error: processing response error (body="{\"jsonrpc\":\"2.0\",\"id\":48,\"error\":{\"code\":-32000,\"message\":\"the method has been deprecated: eth_accounts\"}}\n", error={"code":-32000}, requestBody="{\"method\":\"eth_accounts\",\"params\":[],\"id\":48,\"jsonrpc\":\"2.0\"}", requestMethod="POST", url="<REDACTED_BY_ME>", code=SERVER_ERROR, version=web/5.7.1)
    at Logger.makeError (index.js:224:1)
    at Logger.throwError (index.js:233:1)
    at eval (index.js:261:1)
    at Generator.next (<anonymous>)
    at fulfilled (index.js:5:1)
```

umbra-js/src/utils/utils.ts
```diff
@@ -242,18 +241,6 @@ export async function lookupRecipient(
...
-/**
- * @notice Creates and returns a contract instance
- * @param address Contract address
- * @param abi Contract ABI
- * @param provider ethers provider instance
- */
-export function createContract(address: string, abi: ContractInterface, provider: EthersProvider) {
-  // Use signer if available, otherwise use provider
-  const signer = provider.getSigner();
-  return new Contract(address, abi, signer || provider);
```
mds1, line 254: @apbendi the problem you were seeing is because previously the provider that was passed in to this method was the provider from the user's wallet, so it had a valid signer attached. Now we use a regular RPC URL, and `getSigner` always returns a non-null value even though there's not actually a valid signer attached to e.g. an infura URL, and I couldn't find a good way to distinguish a valid from invalid signer on the returned signer object.

We don't need this method anymore anyway, so just got rid of it and replaced it with direct contract instantiations

frontend/src/store/wallet.ts
mds1: Ok should be fixed per bca969c, details about the fix in https://github.com/ScopeLift/umbra-protocol/pull/488/files#r1163363489,

mds1: Fixed in 1c3177b

apbendi (approved):
LGTM thanks!

PR #490, opened 2023-04-14
chore: bump ethers.js version

mds1:
Enables support for ENS names with emojis, ref https://github.com/ethers-io/ethers.js/issues/2376


apbendi (approved):
LGTM

PR #491, opened 2023-04-17
Batch send frontend

garyghayrat:
https://user-images.githubusercontent.com/61768337/232454824-006bb9df-a716-4348-8e7b-03160d76a567.mp4


mds1 (commented):
good start!

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+            <div class="row items-center justify-center q-pa-xs link-container">
+              {{ $t('Send.send-history') }}
+            </div>
+          </router-link>
+        </div>
+      </div>
+      <base-button
+        v-if="!sendAdvancedButton"
+        :full-width="true"
+        label="Batch Send"
```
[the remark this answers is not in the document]
garyghayrat: Changed to Send in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d86b40a2ff886d978fdd5b0d6ca16703a4d94d69

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+            v-if="sendAdvancedButton"
+            :disable="!isValidForm || isSending"
+            :full-width="true"
+            :label="$t('Send.send')"
+            :loading="isSending"
+            @click="showAdvancedSendWarning = true"
+          />
+
+          <base-button
+            @click="generatePaymentLink({ to: recipientId, token, amount: humanAmount, chainId: chainId })"
```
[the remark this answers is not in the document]
garyghayrat: Removed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d86b40a2ff886d978fdd5b0d6ca16703a4d94d69

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+        <!-- Amount -->
+        <div>
+          {{ $t('Send.amount') }}
+        </div>
+        <base-input
+          v-model="Send.amount"
+          :disable="isSending"
+          placeholder="0"
+          :appendButtonDisable="!Send.receiver || !isValidRecipientId"
+          :appendButtonLabel="$t('Send.max')"
```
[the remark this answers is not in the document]
garyghayrat: For now i think we can definitely remove. Removed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d86b40a2ff886d978fdd5b0d6ca16703a4d94d69

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+      class="dark-toggle form-max-wide text-center text-bold q-pa-md"
+      style="border-radius: 15px"
+      :style="isDark ? 'color: #FFEEEE; background-color: #780A0A' : 'color: #610404; background-color: #FACDCD'"
+    >
+      {{ $t('Send.sending-disabled') }}
+    </div>
+  </q-page>
+
+  <q-page v-else padding>
+    <q-dialog v-model="showAdvancedWarning">
```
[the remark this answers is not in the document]
garyghayrat: Removed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d86b40a2ff886d978fdd5b0d6ca16703a4d94d69

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+
+  // Form parameters.
+  const Sends = ref<SendData[]>([{ receiver: '', token: null, amount: '', recipientIdBaseInputRef: null }]);
+  const recipientId = ref<string>();
+  const useNormalPubKey = ref(false);
+  const token = ref<TokenInfoExtended | null>();
+  const humanAmount = ref<string>();
+  const isValidForm = ref(false);
+  const isValidRecipientId = ref(true); // for showing/hiding bottom space (error message div) under input field
+  const toll = ref<BigNumber>(Zero);
```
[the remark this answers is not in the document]
garyghayrat: Yes, most of the code is taken from the `Send` page. Sorry this was meant to be just a quick review of the UI shown in the video, but it was very helpful to have you look at the code early. 

I've made the PR target branch the `Batch Send Umbra-js` branch so it has cleared out the umbra-js related commits. I haven't made any changes to compared with the existing PR.

I'll work making this page as simple as it can be and see where we could maybe refactor/modularize, and if we should split it up.

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
mds1: Looks like we still have a lot of related variables in the script section, like showAdvancedWarning, sendAdvancedButton, shouldUseNormalPubKey

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
mds1: Looks like we still have related variables in the script section, like `sendMax`

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
mds1: Looks like we still have a lot of related variables in the script section, like `paymentLinkParams` and `generatePaymentLink`. I'd recommend reviewing the script section and making sure we actually use all the data there

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,538 @@
...
+      class="dark-toggle form-max-wide text-center text-bold q-pa-md"
+      style="border-radius: 15px"
+      :style="isDark ? 'color: #FFEEEE; background-color: #780A0A' : 'color: #610404; background-color: #FACDCD'"
+    >
+      {{ $t('Send.sending-disabled') }}
+    </div>
+  </q-page>
+
+  <q-page v-else padding>
+    <h2 class="page-title">Batch Send</h2>
```
mds1, line 13: We'll need translations for all of these, do you want to do that in this PR or track separately?

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,538 @@
...
+    tokenList,
+    toll,
+    useNormalPubKey,
+    userAddress,
+    Sends,
+  };
+}
+
+export default defineComponent({
+  name: 'BatchSend',
```
mds1, line 514: Let's also rename the file to `AccountBatchSend`. The naming convention that I like for components that we use in Umbra is namespacing them. So we have `AccountSend` and `AccountReceive` pages because send and receive are actions an account can take. Then we have `AccountReceiveTable` because that's a table meant to be used with `AccountReceive`, etc.
```suggestion
  name: 'PageBatchSend',
```

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
apbendi: Yeah I like the table style idea as well, even if we just take the existing fields, put them one on each line, and make each one a little less wide so the UI is still manageable.

umbra-js/src/classes/Umbra.ts
```diff
@@ -8,7 +8,6 @@ import {
   BigNumber,
```
apbendi, line 8: *Arbitrary comment location*

The form doesn't currently default to a token. It should default to the base asset as the main send form does.

umbra-js/src/classes/Umbra.ts
```diff
@@ -8,7 +8,6 @@ import {
...
   BigNumberish,
```
apbendi, line 9: *Arbitrary comment location*

I think we should just remove the "Copy payment link" function from the page completely. Not sure it translates well to a batch send context. If we want to consider adding it back later, we can create an issue for that, but I don't think it should be required to ship.

umbra-js/src/classes/Umbra.ts
```diff
@@ -8,7 +8,6 @@ import {
...
   Contract,
```
apbendi, line 10: *Arbitrary comment location*

There should be an "x" on each group of send fields after the first one that allows the user to remove it from the list.

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
garyghayrat: Screen Shot 2023-04-25 at 4 35 25 PM
How's this look?

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
garyghayrat: Hmm, removed the ones that were mentioned https://github.com/ScopeLift/umbra-protocol/pull/491/commits/a272b78852f4f44e4ef42125ef50bf1fffb8e7df

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
garyghayrat: Removed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/a272b78852f4f44e4ef42125ef50bf1fffb8e7df

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
garyghayrat: Removed the mentioned ones in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/a272b78852f4f44e4ef42125ef50bf1fffb8e7df 

Gonna look through again before pulling out of draft.

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+
+interface SendData {
+  receiver: string | undefined;
+  token: TokenInfoExtended | null;
+  amount: string;
+  recipientIdBaseInputRef: any;
+}
+
+// Umbra.js is expected types
+interface newSendBatch {
```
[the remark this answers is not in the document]
garyghayrat: Removed and imported from umbra-js https://github.com/ScopeLift/umbra-protocol/pull/491/commits/3b00ccdf98ed2ba5787bdad8b00d6ea23a5d5468

frontend/src/pages/BatchSend.vue
```diff
@@ -0,0 +1,729 @@
...
+          :appendButtonDisable="!Send.receiver || !isValidRecipientId"
+          :appendButtonLabel="$t('Send.max')"
+          @click="setHumanAmountMax"
+          @input="() => (sendMax = false)"
+          lazy-rules
+          :rules="isValidTokenAmount"
+          ref="humanAmountBaseInputRef"
+        />
+
+        <!-- Toll + summary details -->
```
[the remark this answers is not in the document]
garyghayrat: Screen Shot 2023-04-27 at 12 46 33 PM
Def need to style it more but have it working in [74d02cc](https://github.com/ScopeLift/umbra-protocol/pull/491/commits/74d02cc252a11fd71d8598e27bd06ba25804d509)

umbra-js/src/classes/Umbra.ts
```diff
@@ -567,6 +572,30 @@ export class Umbra {
     return { spendingKeyPair, viewingKeyPair };
   }
 
+  async prepareSend(recipientId: string, lookupOverrides: SendOverrides = {}) {
```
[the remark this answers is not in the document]
garyghayrat: No changes made to `Umbra-js`, made the target [frontend-batch-send-umbra-js](https://github.com/ScopeLift/umbra-protocol/tree/frontend-batch-send-umbra-js) so it shows less in diff. But once we merge the umbra-js, I'll rebase and target master.

frontend/src/pages/BatchSend.vue
garyghayrat: I'll do it in this PR, but haven't addressed it yet. Leaving this open for now

frontend/src/pages/BatchSend.vue
garyghayrat: Renamed in [914b7cc](https://github.com/ScopeLift/umbra-protocol/pull/491/commits/914b7cc6b1d4d871ceb4913d27baf46eb97610f1)

umbra-js/src/classes/Umbra.ts
garyghayrat: Added here: [335d2b8](https://github.com/ScopeLift/umbra-protocol/pull/491/commits/335d2b857b8fa898c365c1288aca95a4877131b4)

umbra-js/src/classes/Umbra.ts
garyghayrat: Agreed, removed relevant lines

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
mds1: Great, thanks. I'll leave this unresolved as a reminder to check for unused vars

umbra-js/src/classes/Umbra.ts
apbendi: Nice! Looking good and working well. One improvement we need: it shouldn't be possible to delete the last row.

delete-last-row

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
apbendi: Love it! Looking great. One issue @mds1 noticed is the summary table formatting is looking a little messed up:


Screen Shot 2023-05-01 at 3 45 15 PM

frontend/src/css/app.sass
```diff
@@ -100,6 +100,42 @@ p, div
   max-width: 50%
```
apbendi, line 100: *arbitrary comment location*

Thinking out loud here @garyghayrat and @mds1, but what if—instead of a separate page and route for batch send—the `/send` page had something like two tabs "Single Send" (the current UI) and "Batch Send" (the new batch send UI) defaulting to "Single Send"? Curious what you guys think of this and how hard/easy it would be to implement.

garyghayrat: I think it makes complete sense. I don't think we even need an extra tab, just add an `add another send` button right below the usual `Send` button. And depending on the `sendData` length we can either call the Umbra and do usual send or call batchSend contract for batch send.

garyghayrat: Question above using visuals, pick one:

image

mds1: If we want tabs, we can do this easily using: https://quasar.dev/vue-components/tab-panels#qtabpanels-api

One nice side effect is that tabs (or more generally, just one page/route and same way to toggle) should make it easier to dedupe logic and code.

Regardless of one page vs. separate pages, I still think the UI should be:
- form style on both desktop/mobile for a single send
- table style on desktop / card style on mobile for batch send

garyghayrat: So the left one?

mds1: Ah sorry, seems you left that comment with the pics as I was writing my comment, so I did not see it. Yea, IMO the left one is better, since there's no reason to change the existing code UI (it already works and people are familiar with it)

umbra-js/src/classes/Umbra.ts
garyghayrat: Fixed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/7d836b1a1d9b24d50624e191bf1ce1cede053541

umbra-js/src/classes/Umbra.ts
garyghayrat: Fixed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d51628fb4803b9a2f67a3c8cd373760d2069234b

garyghayrat: [Screencast from 2023-05-04 11-10-10.webm](https://user-images.githubusercontent.com/61768337/236250568-2fcfa479-9bdd-4a8f-9ebf-9fc599566a4b.webm)

frontend/src/pages/BatchSend.vue
garyghayrat: Fixed them here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bb9b4b6f027fb57c4008bda8d1d2783d0d87029a

frontend/src/pages/BatchSend.vue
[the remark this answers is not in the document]
garyghayrat: Fixed them here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bb9b4b6f027fb57c4008bda8d1d2783d0d87029a

frontend/src/pages/BatchSend.vue
mds1: I've noticed you marked this ready for re-review last week so just bumping this as a reminder

garyghayrat: Added here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/136d2be8b555bb1c501646e75e6c2f1494a9261e

frontend/src/i18n/locales/zh-CN.json
```diff
@@ -89,9 +91,10 @@
     "wallet-not-connected": "钱包未连接",
     "umbra-fee-exceeds-balance": "需要的 Umbra 费用超过钱包金额",
     "advanced-send-warning": "接收者必须将其帐户的私钥输入此应用程序才能提取资金。除非您知道自己在做什么，不要使用此功能",
-    "learn-more": "了解更多信息",
     "acknowledge-risks": "我承认风险",
-    "send-history": "发送历史记录"
+    "send-history": "发送历史记录",
+    "add-send": "+ 添加发送",
+    "receiver-addr-ens": "收件人 0x 地址或 ENS 名称"
```
mds1, line 97: (or equivalent)
```suggestion
    "receiver-addr-ens": "收件人 地址或 ENS 名称"
```

garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/56aa1f6e8bf76c3767220d46da2ead15f4981b52

frontend/src/i18n/locales/en-US.json
```diff
@@ -90,7 +92,9 @@
     "umbra-fee-exceeds-balance": "required for Umbra fee exceeds wallet balance",
     "advanced-send-warning": "The receiver will have to enter their account's private key into this app to withdraw the funds. Don't use this feature unless you know what you're doing.",
     "acknowledge-risks": "I acknowledge the risks",
-    "send-history": "Send history"
+    "send-history": "Send history",
+    "add-send": "+ Add Send",
+    "receiver-addr-ens": "Receiver's 0x address or ENS name"
```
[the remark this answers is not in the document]
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/56aa1f6e8bf76c3767220d46da2ead15f4981b52

frontend/src/css/app.sass
apbendi: The update is looking great @garyghayrat! This is really starting to come together.

A couple quick questions:

1. Is it possible to make the tabbed nature of the interface a little more explicit? As it is, it's not particularly clear—especially on desktop mode—that the "Batch Send" text (which is greyed out and floating on the right) is even something I can click. Here are a couple of examples of interfaces that I think make this a bit clearer. How hard would it be to make style changes in this direction?

image

Screen Shot 2023-05-10 at 4 14 58 PM

2. How difficult would it be to completely remove the tabs on a network where the functionality is not available yet? If it's difficult I think we can live with the approach you have here, but I think simply showing the current page as-is for flagged networks would be ideal.

mds1: Looking good! A few pieces of feedback:

--------

1. For the mobile view, we should either (1) have a single row with amount on the left and token dropdown on the right, or (2) keep the look the same as the send form. Looks like currently we're in a middle ground where the fields aren't full width but are on their own line. Can shorten the amount field placeholder to just `Amount` so it fits better

image

--------

2. Let's add top and bottom margin here so the cards aren't touching. Also it seems they get a gold outline when you touch the card (prob hover is the actual thing), but we don't need that since you can't actually click the card (i.e. the card itself is not actionable, unlike the home page cards where I think that border is from)

image

--------

3. From the above screenshots and the one below, you can see that when I enter an amount, the field's placeholder text moves to the top. But this doesn't happen with the recipient field. We should make these consistent

image

--------

4. The "delete" icon in the above screenshot is just the letter X. Let's use the same close icon we use elsewhere in the codebase. Search `q-icon` usage of the fontawesome `fa-times` logo for examples

--------

5. The tabs are configured to be full width and take up too much space IMO, making them a bit hard to see. Can we center them to both take up less space, so they're closer together? Might look better if we move the header to the top and call it "Send", keep that constant, and below it are tabs called "Single Send" and "Batch Send"

image

frontend/src/css/app.sass
```diff
@@ -100,6 +100,46 @@ p, div
...
   text-align: center
 
+.batch-send
```
apbendi, line 103: *Arbitrary comment location*

When I add multiple sends with amounts >1000 DAI I get a bunch of errors on input:

multi-dai-batch-send-error

frontend/src/css/app.sass
garyghayrat: 1. Added delete icon instead of "X" https://github.com/ScopeLift/umbra-protocol/pull/491/commits/f91f44bf054b3828fdbfed4cbbcd7a08ccc87a30
2. Changed mobile match send view to match the Send page style and remove golden outline/highlight https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bd3bcdc635be649ca2742ea788fb9d7674d02f22
3. Removed tabs completely when batch send isn't supported https://github.com/ScopeLift/umbra-protocol/pull/491/commits/01c779b7811e9f712c798cbac20878d8822f5c5a
4. Used smaller numbers for desktop view https://github.com/ScopeLift/umbra-protocol/pull/491/commits/ff6469554e0f7f156e26e745347d66d5de7a61d3
5. Moved tabs under title and brought them closer to each other https://github.com/ScopeLift/umbra-protocol/pull/491/commits/749a4c043f2a9b778f7153a8c1f8774e54dd6b6a
6. Made input filed styles uniform on desktop https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bac053ac2343b5984d1b2a7e1af316d7e118ccba
7. Fixed summary erroring out when sending 1k+ tokens https://github.com/ScopeLift/umbra-protocol/pull/491/commits/16d401071503c3b1fd70d820b2f16f0cb2a2bea8
8. Fixed flaky feature flag https://github.com/ScopeLift/umbra-protocol/pull/491/commits/b73e29003e80fc37162fd5193c445417561b70bc

frontend/src/css/app.sass
garyghayrat: Fixed in https://github.com/ScopeLift/umbra-protocol/commit/16d401071503c3b1fd70d820b2f16f0cb2a2bea8

frontend/src/css/app.sass
```diff
@@ -96,10 +96,66 @@ p, div
   @media (max-width: $breakpoint-sm-max)
```
apbendi, line 96: *arbitrary comment location*

On the single send page, when you enter an invalid value for the amount, it simply gives you a message on the form validation. On the batch send page, every time I type an invalid character it throws an error and shows a toast:

batch-amount-errors

frontend/src/css/app.sass
```diff
@@ -96,10 +96,66 @@ p, div
...
     font-size: 3rem
```
apbendi, line 97: *arbitrary comment location*

Nit: seems we lost a space before the little icon:

Prod:

Screen Shot 2023-05-16 at 11 35 23 AM

Branch:

Screen Shot 2023-05-16 at 11 35 48 AM

frontend/src/css/app.sass
apbendi: Great work @garyghayrat, on my first pass through it seems all these issues are resolved. I think the tabs look fine with the spacing fixed. I'll let @mds1 give everything a look to see if he agrees. This is really looking great!

mds1: Yea tabs def look better now, thanks! From my numbering:
1. Resolved
2. Seems the gold outline is still there when I click using mobile view in chrome desktop, maybe that's not true on mobile? Connecting a wallet on localhost on mobile didn't work last time so we can just leave this for now and if it's there in prod it's not a big deal / not a blocker
3. Resolved
4. IMO they're still pretty big and take focus, but they shouldn't be a focus. They should be roughly the same size as the X, they don't need to be header elements. You can also try making them a shade of gray for less emphasis. Not gonna block on this though
5. Resolved

New one: there is a container/background of a different color on dark mode, but not on light mode, which we should unify. Prob no container for both?

image

image

garyghayrat: 2. Gold outline should really be gone now https://github.com/ScopeLift/umbra-protocol/pull/491/commits/db8e00453c51adcc585c24922ddc9d5ee9816c40
4. Made 'X'es stand out much less https://github.com/ScopeLift/umbra-protocol/pull/491/commits/fbbb59333bb200d9ccf41551124f83ca721b9bc6
New one: good catch and not sure how it propped up but is fixed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/f4570718db6b52e9fc529f84e66d0e689051ba00

frontend/src/css/app.sass
garyghayrat: Added a `isValidAmount` checker https://github.com/ScopeLift/umbra-protocol/pull/491/commits/d8b4b648262e510547fee0b2dcf451f9841940a6

frontend/src/components/BaseButton.vue
```diff
@@ -17,7 +17,7 @@
       :text-color="textColor"
```
apbendi, line 17: *arbitrary comment location*

This is *really* coming together @garyghayrat. Testing it out on Goerli and overall it's working great. I really tried to hammer on it to find edge case behavior. Here are few things I discovered:

* Once I submit the transaction and it's pending, it should no longer be possible to click the "x" to remove that row
* Same with the "+ Add Send" button. It shouldn't be possible to push this while a transaction is pending.
* After the transaction completes, the form doesn't clear. On Single send we clear the form, which I *think* makes more sense for this case too. Curious what you think @mds1.
* It's currently possible to do batch send with only 1 row. This doesn't really make sense and we should probably prevent users from doing it by preventing the send button from being clicked unless there are two valid rows.
* While a transaction is pending I can still switch back between the tabs. This seems a little odd—should we disallow it? Curious what you think.
* After completing a batch send, if I switch back to the single send tab, the form no longer has a default asset selected:

Screen Shot 2023-05-19 at 6 04 17 PM

* Furthermore, after completing a batch send, if I switch back to the single send form and fill it out, the send button on that form remains disabled. I have to reload before I can do a single send.

garyghayrat: >     * Once I submit the transaction and it's pending, it should no longer be possible to click the "x" to remove that row
> 
>     * Same with the "+ Add Send" button. It shouldn't be possible to push this while a transaction is pending.
> 
>     * While a transaction is pending I can still switch back between the tabs. This seems a little odd—should we disallow it? Curious what you think.
> 

I think it makes sense to disable switching tabs while tx is pending. Disabled all mentioned during `isSending` in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bf8ce3e8d391fffd4100878a9d47f20ecb8cb18f


>     * After the transaction completes, the form doesn't clear. On Single send we clear the form, which I _think_ makes more sense for this case too. Curious what you think @mds1.
> 
>     * It's currently possible to do batch send with only 1 row. This doesn't really make sense and we should probably prevent users from doing it by preventing the send button from being clicked unless there are two valid rows.
> 
Feel like we could make a case for both, but clearing the form for now to keep the consistency in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/bfcadb449f639b4623eb79072235541032fec505
 In the same commit, the `Send` button is disabled if there's only 1 row. Also now the batchSend page starts with 2 rows.

>     * After completing a batch send, if I switch back to the single send tab, the form no longer has a default asset selected:
> 
>     * Furthermore, after completing a batch send, if I switch back to the single send form and fill it out, the send button on that form remains disabled. I have to reload before I can do a single send.
These are fixed in https://github.com/ScopeLift/umbra-protocol/pull/491/commits/407ac6f90d7b5cea9c3154fd5df832724c54332e

frontend/src/pages/AccountSend.vue
```diff
@@ -649,32 +1092,43 @@ function useSendForm() {
...
+    Sends,
     sendAdvancedButton,
     sendFormRef,
+    sendingString,
     sendMax,
     setHumanAmountMax,
     showAdvancedSendWarning,
     showAdvancedWarning,
+    summaryTotalString,
+    tab,
```
mds1, line 1131: This file is was already pretty big at ~700 lines, and now it's over 60% bigger at 1167 lines. As a result this is probably going to be hard to maintain. I don't want to block this PR so we can clean it up separately, but we definitely should have split this up into components, such as:
- `AccountSend.vue` is basically the page layout and any shared data that's needed by both child components, managing the tabs, etc.
- `AccountSendSingle.vue` is the "Send" tab component/logic
- `AccountSendBatch.vue` is the "Batch Send" tab component/logic
- `accountSendUtils.ts` can house shared methods/logic

frontend/src/pages/AccountSend.vue
```diff
@@ -279,12 +517,20 @@ import {
 } from 'src/utils/ethers';
 import { humanizeTokenAmount, humanizeMinSendAmount, humanizeArithmeticResult } from 'src/utils/utils';
 import { generatePaymentLink, parsePaymentLink } from 'src/utils/payment-links';
+import { SendBatch } from 'app/../umbra-js/build/src/types';
```
mds1, line 520: We should avoid relative paths for imports and be consistent with the other umbra-js imports, by only importing things exported by the package. So if the below doesn't work we should update umbra-js to export the types. In other words, the current import syntax only works for us in a monorepo, the below import syntax would work for anyone
```suggestion
import { SendBatch } from '@umbracash/umbra-js';
```

frontend/src/pages/AccountSend.vue
```diff
@@ -279,12 +517,20 @@ import {
...
 import { Provider, TokenInfoExtended, supportedChains } from 'components/models';
 import { ERC20_ABI } from 'src/utils/constants';
 import { toAddress } from 'src/utils/address';
 
+interface SendData {
```
mds1, line 525: Now that this component mixes both regular and batch send logic, let's try to use comments for documenting, or use more self-documenting names
```suggestion
interface BatchSendData {
```

```suggestion
// Batch send data
interface SendData {
```

frontend/src/pages/AccountSend.vue
```diff
@@ -309,6 +555,7 @@ function useSendForm() {
 
   // Form refs for triggering validation via form items' `validate()` method.
   const recipientIdBaseInputRef = ref<InstanceType<typeof BaseInput> | null>(null);
+
```
mds1, line 558:
```suggestion
```

frontend/src/pages/AccountSend.vue
```diff
@@ -325,11 +572,21 @@ function useSendForm() {
   const attemptedNetworkChange = ref(false);
   const connectRedirectTo = ref('send');
 
+  // Batch Send Form Parameters
+  const Sends = ref<SendData[]>([]);
```
mds1, line 576: variables should be lower case
```suggestion
  const batchSends = ref<SendData[]>([]);
```

frontend/src/pages/AccountSend.vue
```diff
@@ -393,6 +716,8 @@ function useSendForm() {
       } else if (nativeTokenValue.chainId !== prevNativeTokenValue.chainId || isLoadingValue !== prevIsLoadingValue) {
         // When network finally connects after page load, we need to re-parse the payment link data.
         await setPaymentLinkData(); // Handles validations.
+        batchSendIsSupported.value = batchSendSupportedChains.includes(Number(currentChain.value?.chainId));
```
mds1, line 719: chainId is a `uint256` so let's treat it as such and avoid using `Number()` to cast. In practice this won't matter for us because (1) we only support a few chains and all have chain IDs smaller than `Number.MAX_SAFE_INTEGER` and (2) AFAIK no chains currently have a chain ID that big anyway, but better to be cautious
```suggestion
        const chainId = BigNumber.from(currentChain.value?.chainId).toNumber()
        batchSendIsSupported.value = batchSendSupportedChains.includes(chainId);
```

frontend/src/pages/AccountSend.vue
```diff
@@ -431,6 +761,22 @@ function useSendForm() {
         await setNetwork(chain[0]);
       }
     }
+
+    batchSendIsSupported.value = batchSendSupportedChains.includes(Number(linkChainId));
```
mds1, line 765: Same comment. But also, seems weird that we're doing this inside `setPaymentLinkData`? They feel unrelated. Right now this logic runs twice on mount because we call `setPaymentLinkData` which runs this line, then this same line exists directly in `onMounted`

frontend/src/pages/AccountSend.vue
```diff
@@ -469,15 +815,22 @@ function useSendForm() {
...
     if (!val || !(Number(val) > 0)) return tc('Send.enter-an-amount');
-    if (!token.value) return tc('Send.select-a-token');
 
-    const { address: tokenAddress, decimals } = token.value;
+    let tokenToUse: TokenInfoExtended | null | undefined;
+    // Use batchSend token if it exists
+    if (tokenInput) tokenToUse = tokenInput;
+    // Otherwise use the current Send token
+    else tokenToUse = token.value;
+    if (!tokenToUse) return tc('Send.select-a-token');
```
mds1, line 827: this feels clunky, the below should be equivalent right?
```suggestion
    const tokenToUse: tokenInput || token.value;
    if (!tokenToUse) return tc('Send.select-a-token');
```

frontend/src/pages/AccountSend.vue
```diff
@@ -469,15 +815,22 @@ function useSendForm() {
...
+
+    const { address: tokenAddress, decimals } = tokenToUse;
+
     const minAmt = getMinSendAmount(tokenAddress);
     if (Number(val) < minAmt && isNativeToken(tokenAddress)) return `${tc('Send.send-at-least')} ${minAmt} ${NATIVE_TOKEN.value.symbol}`; // prettier-ignore
-    if (Number(val) < minAmt && !isNativeToken(tokenAddress)) return `${tc('Send.send-at-least')} ${minAmt} ${token.value.symbol}`; // prettier-ignore
+    if (Number(val) < minAmt && !isNativeToken(tokenAddress)) return `${tc('Send.send-at-least')} ${minAmt} ${tokenToUse.symbol}`; // prettier-ignore
```
mds1, line 833: Only one of these can be true, so `else if` seems more clear
```suggestion
    if (Number(val) < minAmt && isNativeToken(tokenAddress)) return `${tc('Send.send-at-least')} ${minAmt} ${NATIVE_TOKEN.value.symbol}`; // prettier-ignore
    else if (Number(val) < minAmt && !isNativeToken(tokenAddress)) return `${tc('Send.send-at-least')} ${minAmt} ${tokenToUse.symbol}`; // prettier-ignore
```

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
   }
 
-  function resetForm() {
+  // Send funds
+  async function onBatchSendFormSubmit() {
+    try {
+      // Form validation
+      for (let i = 0; i < Sends.value.length; i++) {
+        if (!Sends.value[i].receiver || !Sends.value[i].token || !Sends.value[i].amount)
+          throw new Error('please complete the form');
```
mds1, line 959: Also should these errors have i18n translations? Not sure offhand what we normally do there
```suggestion
          throw new Error('Please complete the form');
```

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+      }
+      if (!signer.value) throw new Error('wallet not connected');
```
mds1, line 961:
```suggestion
      if (!signer.value) throw new Error('Wallet not connected');
```

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+      if (!umbra.value) throw new Error('Umbra instance not configured');
+
+      // Verify the recipient ID is valid. (This throws if public keys could not be found. This check is also
+      // done in the Umbra class `send` method, but we do it here to throw before the user pays for a token approval.
+      // This should usually be caught by the isValidId rule anyway, but is here again as a safety check)
+      const ethersProvider = provider.value as Provider;
+      for (let i = 0; i < Sends.value.length; i++) {
+        await umbraUtils.lookupRecipient(Sends.value[i].receiver as string, ethersProvider);
```
mds1, line 969: remember that `await` in loops are very bad for performance, this should be refactored to a map with `Promise.all`. We should also consider a tracking issue to change these lookups to a single multicall for even better performance and less RPC usage

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+      }
+      // Ensure user has enough balance. We re-fetch token balances in case amounts changed since wallet was connected.
+      // This does not account for gas fees, but this gets us close enough and we delegate that to the wallet
+      // await getTokenBalances();
+
+      for (let i = 0; i < Sends.value.length; i++) {
```
mds1, line 975:
```suggestion
      // await getTokenBalances();
      for (let i = 0; i < Sends.value.length; i++) {
```

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+        const token: TokenInfoExtended | null | undefined = Sends.value[i].token;
+        if (!token) throw new Error('please complete the form');
```
mds1, line 977:
```suggestion
        if (!token) throw new Error(`Please select a token for send {i + 1}`);
```

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+      if (!umbra.value) throw new Error('Umbra instance not configured');
+
+      // Verify the recipient ID is valid. (This throws if public keys could not be found. This check is also
+      // done in the Umbra class `send` method, but we do it here to throw before the user pays for a token approval.
+      // This should usually be caught by the isValidId rule anyway, but is here again as a safety check)
+      const ethersProvider = provider.value as Provider;
+      for (let i = 0; i < Sends.value.length; i++) {
+        await umbraUtils.lookupRecipient(Sends.value[i].receiver as string, ethersProvider);
+      }
+      // Ensure user has enough balance. We re-fetch token balances in case amounts changed since wallet was connected.
```
mds1, line 971: Contrary to this comment I don't see a call that actually re-fetches token balances

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+        if (token.symbol !== NATIVE_TOKEN.value.symbol) {
+          // Check allowance
+          const tokenContract = new Contract(token.address, ERC20_ABI, signer.value);
+          const batchSendAddress = umbra.value.batchSendContract!.address;
+          const allowance = (await tokenContract.allowance(userAddress.value, batchSendAddress)) as BigNumber;
+          // If insufficient allowance, get approval
+          if (tokenAmount.gt(allowance)) {
+            const approveTx = await tokenContract.approve(batchSendAddress, MaxUint256);
+            void txNotify(approveTx.hash as string, ethersProvider);
+            await approveTx.wait();
```
mds1, line 1005: Lots of awaits in loops
- the allowance checks should be pulled out into a map + Promise.all
- we can send off all approval txs WITHOUT calling `await approveTx.wait()`
- then we can await all of them at at the end for better UX: Right now if a user is doing a batch send with 2 tokens the flow is approve tx > wait for it to confirm > approve tx > wait for it to confirm > batch send. But it can be approve tx > approve tx > wait for both to confirm > batch send

frontend/src/pages/AccountSend.vue
mds1: Think my suggestion got messed up here by highlighting too many lines but I think you get the idea

frontend/src/pages/AccountSend.vue
```diff
@@ -590,20 +943,110 @@ function useSendForm() {
...
+      void txNotify(tx.hash, ethersProvider);
+      await tx.wait();
+      resetBatchSendForm();
+    } finally {
+      isSending.value = false;
+    }
+  }
+
+  async function resetForm() {
+    const { token: paymentToken } = await parsePaymentLink(NATIVE_TOKEN.value);
```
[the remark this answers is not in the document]
garyghayrat: Oh it's just to get the default token for the chain we're on. Maybe there is a more efficient way to do this? Just thought duplicating the following code that's inside `parsePaymentLink` would be too much
```
  // If no `token` symbol was given, we can return with the payment data.
  let tokenSymbol = params.get('token')?.toLowerCase();
  if (!tokenSymbol) return paymentData;

  // Parsing the `token` symbol has some additional logic.
  const tokens = await getTokens(nativeToken); // Get list of supported tokens.
  const chainId = BigNumber.from(nativeToken.chainId || 1).toNumber();

  if (tokenSymbol === 'eth' && chainId === 137) {
    // If the token is ETH, and we're on Polygon, use WETH, since the native token is MATIC.
    tokenSymbol = 'weth';
    paymentData['token'] = tokens.filter((token) => token.symbol.toLowerCase() === tokenSymbol)[0];
  } else if (tokenSymbol === 'matic' && chainId !== 137) {
    // If the token is MATIC, and we're not on polygon, clear token and amount.
    paymentData['token'] = null;
    paymentData['amount'] = null;
  } else {
    // Otherwise, find the matching `TokenInfoExtended` object
    paymentData['token'] = tokens.filter((token) => token.symbol.toLowerCase() === tokenSymbol)[0];
  }
```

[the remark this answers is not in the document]
mds1: The native token info is defined in a constant here: https://github.com/ScopeLift/umbra-protocol/blob/c2333115bf52ea1493d3db5f03eb46eabe704815/frontend/src/components/models.ts#L42

But I still don't understand why the original version of the method wasn't sufficient. Offhand I can't recall why we reset the token to undefined and not the native token, but presumably that would be the only change we need

frontend/src/components/BaseButton.vue
garyghayrat: It's fixed here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/45c87b79f3f77595afa567a38e064c3db3e8f713

frontend/src/pages/AccountSend.vue
garyghayrat: Makes sense, thank you! Fixed it here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/4ba9948658c97f5f08836e3701b9ce46eeaf7385

frontend/src/pages/AccountSend.vue
garyghayrat: Renamed it to BatchSendData here : https://github.com/ScopeLift/umbra-protocol/pull/491/commits/7c37eca7abdc6b67f495a8154b17c3ac417361c0

frontend/src/pages/AccountSend.vue
garyghayrat: Removed the extra space

frontend/src/pages/AccountSend.vue
garyghayrat: Thanks for the reminder. Renamed it here: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/7c4e67a7730b4a83fbcb6dd87f3ad74cafbce2ad

frontend/src/pages/AccountSend.vue
garyghayrat: Good to know, fixed it here https://github.com/ScopeLift/umbra-protocol/pull/491/commits/a3c2970279e53167566ddb40bc6d834ea7eefbbe

frontend/src/pages/AccountSend.vue
garyghayrat: Looks like a duplicate leftover from me trying to fix the flakiness around showing/hiding batchSend when switching networks. Removed the line from `setPaymentLinkData` and everything is still working as expected. https://github.com/ScopeLift/umbra-protocol/pull/491/commits/85765a48270e8cb1953f8fa41932a7dce1544fe2

frontend/src/pages/AccountSend.vue
garyghayrat: Indeed https://github.com/ScopeLift/umbra-protocol/pull/491/commits/f34b12a298489db5489c62ed096389876b209bde

frontend/src/pages/AccountSend.vue
garyghayrat: Like this? https://github.com/ScopeLift/umbra-protocol/pull/491/commits/7ac43663c8820bb6153315566b17291a95d40114

frontend/src/pages/AccountSend.vue
garyghayrat: Yes! Added translations here https://github.com/ScopeLift/umbra-protocol/pull/491/commits/28cbcfdc97fce7c07c890f793abd1159c6415181

frontend/src/pages/AccountSend.vue
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/28cbcfdc97fce7c07c890f793abd1159c6415181

frontend/src/pages/AccountSend.vue
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/491/commits/7c1c4d29e37f189ddf1e0245c1286acb9e56e722

frontend/src/pages/AccountSend.vue
garyghayrat: Uncommented the line that refetches balance https://github.com/ScopeLift/umbra-protocol/pull/491/commits/14f3447edb1b8fd70b273bdd1752da87e425c110

frontend/src/pages/AccountSend.vue
garyghayrat: Added here along with translation support https://github.com/ScopeLift/umbra-protocol/pull/491/commits/28cbcfdc97fce7c07c890f793abd1159c6415181

frontend/src/pages/AccountSend.vue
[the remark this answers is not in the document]
garyghayrat: The original version wasn't sufficient because after completing a tx, the token field of the page will become an empty field instead of reverting back to the default token. What if i just did this? seem to be working as expected https://github.com/ScopeLift/umbra-protocol/pull/491/commits/b7a699b9b97e173adb815eeec6f3b967bb7ba652

frontend/src/pages/AccountSend.vue
garyghayrat: yea, I have a terrible habit of this but should be fixed here https://github.com/ScopeLift/umbra-protocol/pull/491/commits/6bce5d4a1a5581d1757361ce405cd01d37814bdc

frontend/src/pages/AccountSend.vue
garyghayrat: How does this look? https://github.com/ScopeLift/umbra-protocol/pull/491/commits/66644c31227edaa495faf3c895537973a62933b4

frontend/src/pages/AccountSend.vue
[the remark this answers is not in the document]
mds1: That lgtm and is what I'd have expected, thanks!

frontend/src/pages/AccountSend.vue
mds1: you don't need the `async` keyword in your map, because we're synchronously returning a promise—the below snippet changes `async (send)` to just `(send)`
```typescript
await Promise.all(
  batchSends.value.map((send) => {
    return umbraUtils.lookupRecipient(send.receiver as string, ethersProvider);
  })
);
```

frontend/src/pages/AccountSend.vue
mds1: Really great! Two small comments
- I think we need `promises.push(Promise.resolve(Zero))` instead of `promises.push(Zero)`. The latter isn't pushing a promise, whereas the former is pushing one that will immediately resolve. If the code is working, perhaps `Promise.all` is smart enough to handle the current code, though I haven't tested
- `batchSendAddress` can be defined as a constant outside of the loops instead of redeclaring the same value in all loop iterations

frontend/src/pages/AccountSend.vue
garyghayrat: Makes sense. Looks like I caught it in later commits though, latest code don't have the `async` https://github.com/ScopeLift/umbra-protocol/blob/96389cd21c05143a62fd4450470c0fc9e9b744e9/frontend/src/pages/AccountSend.vue#L989

frontend/src/pages/AccountSend.vue
garyghayrat: - Pushing a promise def makes more sense. 
- Pulled `batchSendAddress` out of the loop
https://github.com/ScopeLift/umbra-protocol/pull/491/commits/38f23b03cc53a6ea369aba89182e6bf3b7fa9c43

apbendi (approved):
Amazing work on this @garyghayratk. It's been a long road! Congrats, and approved 🎉

mds1 (approved):
🚢🚢🚢 nice job @garyghayrat!

Issue #497, opened 2023-04-18
Fetch Stealth Address Balances Through Multicall


apbendi:
Currently, when the app completes Scanning events for transactions sent to the current user, the Receive table loads and checks the token balances of every stealth address via individual RPC calls. See: https://github.com/ScopeLift/umbra-protocol/blob/master/frontend/src/components/AccountReceiveTable.vue#L512-L516

We should refactor this to fetch for these balances via a multicall to reduce load on the user's wallet RPC. This *might* also resolve #496.

PR #498, opened 2023-04-19
perf: use multicall to lookup stealth balances

mds1:
Closes https://github.com/ScopeLift/umbra-protocol/issues/497


apbendi (approved):
Looks good and working well 👍

PR #499, opened 2023-04-25
Warn withdrawl to gitcoin contributor address

### LINKED ISSUE #469: Warn users when they are withdrawing to an address that has contributed to a Gitcoin Grant
If an address has contributed to a Gitcoin Grant (referring here to the centralized Grants application, not the new decentralized protocol) it should be considered doxxed. Therefore, if a user attempts to withdraw funds received via Umbra to such an address, we ought to warn them.

The API now includes an `is-gitcoin-contributor` endpoint. For example, `https://mainnet.api.umbra.cash/is-gitcoin-contributor/0x4C49862AD4A3d36a7d3624Fb0FF78A57434DB39D` now returns:

```json
{
  "umbraApiVersion": {
    "major": 1,
    "minor": 3,
    "patch": 0
  },
  "isContributor": false
}
```

We should use this endpoint to display a warning to the user at the same time as other such warnings.

image

If the endpoint fails/returns an error, the UI should fail silently while logging the error in the Console. It should not display an error message in the frontend nor block the user from proceeding in any way.

alexkeating:
#### Description

- Add gitcoin contributor warning

[Screencast from 2023-04-25 08-41-26.webm](https://user-images.githubusercontent.com/12705146/234279909-dde6b3b1-1180-49c2-8d07-5e89e323b96c.webm)

Closes #469

### DIFF
--- a/frontend/src/i18n/locales/en-US.json
+++ b/frontend/src/i18n/locales/en-US.json
@@ -347,7 +347,8 @@
       "same-addr-as-wallet": "the same address as the connected wallet",
       "same-addr-as-stealth": "the same address as the stealth address",
       "address-it-resolves-to": "The address it resolves to",
-      "has-poap-tokens": "has POAP tokens"
+      "has-poap-tokens": "has POAP tokens",
+      "gitcoin-contributor": "is a gitcoin contributor"
     },
     "Dialog": {
       "warning": "Warning"
--- a/frontend/src/i18n/locales/zh-CN.json
+++ b/frontend/src/i18n/locales/zh-CN.json
@@ -348,7 +348,8 @@
       "same-addr-as-wallet": "与连接的钱包地址相同",
       "same-addr-as-stealth": "与隐形地址相同",
       "address-it-resolves-to": "它解析为的地址",
-      "has-poap-tokens": "具有 POAP 令牌"
+      "has-poap-tokens": "具有 POAP 令牌",
+      "gitcoin-contributor": "是Gitcoin的贡献者"
     },
     "Dialog": {
       "warning": "警告！"
--- a/umbra-js/src/utils/utils.ts
+++ b/umbra-js/src/utils/utils.ts
@@ -586,3 +586,5 @@ export async function assertSupportedAddress(recipientId: string) {
   if (await contract.isSanctioned(address)) throw new Error(errMsg);
   return true;
 }
+
+export const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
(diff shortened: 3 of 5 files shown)


frontend/src/i18n/locales/zh-CN.json
```diff
@@ -347,7 +347,8 @@
       "same-addr-as-wallet": "与连接的钱包地址相同",
       "same-addr-as-stealth": "与隐形地址相同",
       "address-it-resolves-to": "它解析为的地址",
-      "has-poap-tokens": "具有 POAP 令牌"
+      "has-poap-tokens": "具有 POAP 令牌",
+      "gitcoin-contributor": "是Gitcoin的贡献者"
```
mds1, line 351: @garyghayrat can you confirm these?

umbra-js/src/utils/utils.ts
```diff
@@ -586,3 +586,5 @@ export async function assertSupportedAddress(recipientId: string) {
   if (await contract.isSanctioned(address)) throw new Error(errMsg);
   return true;
 }
+
+export const jsonFetch = (url: string) => fetch(url).then((res) => res.json());
```
mds1, line 590: Can we keep this `jsonFetch` definitionin the frontend? Because:
1. That's where we use it, and it's not really related to this umbra-js file
2. It uses fetch, which is defined in browsers (i.e. frontend folder) but doesn't exist in node environments. The umbra-js code should work in both browsers and node, and this breaks that

mds1 (commented):
lgtm, two small comments

apbendi (approved):
Functionality wise, this seems to be working fine. Aside from addressing @mds1's feedback, I think this is good to go.

umbra-js/src/utils/utils.ts
alexkeating: Ah yes, removed from umbra-js

frontend/src/i18n/locales/zh-CN.json
garyghayrat: lgtm!

mds1 (approved):

frontend/src/i18n/locales/en-US.json
```diff
@@ -346,7 +346,8 @@
       "same-addr-as-wallet": "the same address as the connected wallet",
```
apbendi, line 346: *arbitrary comment location*

Hey @alexkeating, we updated the format of the gitcoin contributor endpoint to be `/addresses/:address/is-gitcoin-contributor`. Please update this PR at your convenience to reflect the change. Thanks!

PR #504, opened 2023-05-01
Add deploy script

garyghayrat:
How to use:
1. Create and fill out all fields listed in `.env.example`
2. In your terminal use command 
- `cast nonce <Deployer's address> --rpc-url <URL>` to find your nonce
4. Change `EXPECTED_NONCE` of `DeployBatchSend` script to match your nonce.
5. Inside `contracts-periphery` folder, run 
- `forge test --mc DeployBatchSendTest --sender <Deployer's address>` to make sure the script test passes.
6. Pass in the `private key` to dry run deploying the contract across 6 networks. You should see gas estimates for 6 networks, otherwise the address might be taken.
- `forge script DeployBatchSend --private-key <Private Key> `
7. Actually deploy the contracts.
- `forge script DeployBatchSend --private-key <Private Key> --broadcast`


contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,29 @@
...
   address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
 
-  function run() public {
+  function deploy() internal {
     vm.broadcast();
     new UmbraBatchSend(IUmbra(UMBRA));
   }
 }
+
+contract ScriptBase is Deploy {
```
mds1, line 18: I think you have things flipped here? `contract Deploy is ScriptBase` would be more logical. If we plan to use the base script for other deploy scripts too, let's pull it out into a separate file that does not have the `.s` extension, e.g. `script/BaseScript.sol`

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,29 @@
...
+  string[] public networks = ["mainnet", "optimism", "arbitrum_one", "polygon", "goerli", "sepolia"];
+
+  function run(address expectedContractAddress) public {
```
mds1, line 21: Can we compute the `expectedContractAddress` in the script instead? Feels a lot safer that way. We'll know the deployer address, we'll know the expected nonce, and we can use `computeCreateAddress` from forge-std. This also simplifies the deploy command by removing the need for the `--sig` arg since `run()` is the default sig it looks for

garyghayrat: I knew this must be possible but couldn't find 'computeCreateAddress' easily from the Foundrybook. Maybe when I get the feature where chatGPT can crawl URLs that you provide, I would be able to find it and find it faster.

contracts-periphery/script/Deploy.s.sol
garyghayrat: ohhh hah yes that def makes more sense

contracts-periphery/script/Deploy.s.sol
mds1: yea it's part of forge-std, not foundry, so might not be in the book: https://github.com/foundry-rs/forge-std/blob/73d44ec7d124e3831bc5f832267889ffb6f9bc3f/src/StdUtils.sol#L93

contracts-periphery/script/Deploy.s.sol
garyghayrat: Like this https://github.com/ScopeLift/umbra-protocol/pull/504/commits/d5716543b4c003378cfa32abf56ed0935dac6a69 ?

contracts-periphery/script/Deploy.s.sol
garyghayrat: It's actually in the book: https://book.getfoundry.sh/reference/forge-std/compute-create-address?highlight=computeCreateAddress#computecreateaddress

Implemented changes here: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/a6862bcf242ee41664cb58fd0810cba3bc65003e

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,29 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.16;
 
-import "forge-std/Script.sol";
+import {Script} from "forge-std/Script.sol";
+import {StdChains} from "forge-std/StdChains.sol";
 import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
 import {IUmbra} from "src/interface/IUmbra.sol";
 
 contract Deploy is Script {
```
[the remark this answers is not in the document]
garyghayrat: Added comments https://github.com/ScopeLift/umbra-protocol/pull/504/commits/56b4e9684ccbace6a95c17e26a2f22644773327e

contracts-periphery/script/Deploy.s.sol
mds1: Not exactly, sorry let me clarify:
- `Base*` typically means that thing (whether a contract, class, etc) is meant to be inherited from and used at the base for other things
- In the forge scripting case, a `BaseScript` would contain methods, tools, or utilities that you want all your forge scripts to share
- What are those in this case? To me, it seems like it's (1) calling `setFallbackToDefaultRpcUrls(false)`, and (2) checking if a script is already deployed and deploying it if not
- Based on that logic, the current architecture swaps the content of what should be in the deploy script and the base script: `BaseScript` has the deploy method for the batch send contract which doesn't make much sense IMO

However, both of these scripts are simple enough, that we're probably over-abstracting. Let's instead create a new file called `DeployBatchSend.s.sol` and give the contract the same name. This makes it clear what this file does, where as the current file names are ambiguous—what is `Deploy.s.sol` actually deploying? Then we can put all code for this script into the `run` method of that contract

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,35 @@
...
-import "forge-std/Script.sol";
-import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
-import {IUmbra} from "src/interface/IUmbra.sol";
+/// @title Deploy a contract to multiple networks.
+/// @notice Run the script with `forge script Deploy --private-key <privateKey>`
+/// and add `--broadcast` to broadcast the tx.
+/// @dev The contract that's being deployed is specified in `BaseScript.sol`.
+/// Specify additional netowrk aliases in [rpc_enpoints] in `forge.toml`
+/// before adding them to the list of networks below.
+/// Add corresponding `env` variables in a `.env` file, such as ETH_RPC_URL=<URL>
```
mds1, line 10: A file can have multiple contracts, so this should be directly above the `contract` line. Since this script is for devs only we can just do this (note some typo fixes):
```suggestion
/// @dev Deploys the UmbraBatchSend contract to all supported networks networks, skipping
/// networks that already have the contract
/// @dev Run the script with `forge script Deploy --private-key <privateKey>`
/// and add `--broadcast` to broadcast the tx.
/// @dev Specify additional network aliases in [rpc_endpoints] in `forge.toml`
/// before adding them to the list of networks below, and add corresponding environment
/// variables to the `.env` and `.env.template` files.
```

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,35 @@
...
 
-contract Deploy is Script {
-  address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+import {StdChains} from "forge-std/StdChains.sol";
+import {BaseScript} from "script/BaseScript.sol";
 
+contract Deploy is BaseScript {
+  /// The list of networks to deploy to.
```
mds1, line 16: Everything else can just be regular comments, we don't actually need natspec here, it's just useful for the top-level contract documentation because the `@dev` tags are nice visual separators for different things you want to convey to the devs running/reading/editing the script
```suggestion
  // The list of networks to deploy to.
```

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,35 @@
...
+
+    /// Turn off fallback to default RPC URLs since they can be flaky.
+    setFallbackToDefaultRpcUrls(false);
+
+    /// Check if the contract is already deployed on the network, if not, deploy it.
+    for (uint256 i; i < networks.length; i++) {
+      vm.createSelectFork(getChain(networks[i]).rpcUrl);
+      bool isDeployed = address(expectedContractAddress).code.length > 0;
+      if (!isDeployed) deploy();
+    }
```
mds1, line 33: There's a few issues here:
1. Nonces differ by chain, but we check the nonce once upfront, before we select a fork, and use that for all chains.
2. With this logic, we'll always deploy to every network. What this does is compute the expected address for the `msg.sender` using their current nonce, which is the nonce used by the next tx. By definition, there will never be a contract there

We need something like this, which assumes the expected nonce we want to deploy at is 1. Le me know if you have any questions on this. I'd also suggest writing some tests for your script so you can verify it does what you expect :) 

```solidity
contract Deploy is BaseScript {
  // TODO Get a real value here. The nonce we need to deploy the contract on each chain.
  uint256 constant EXPECTED_NONCE = 1;

  // The list of networks to deploy to.
  string[] public networks = ["mainnet", "optimism", "arbitrum_one", "polygon", "goerli", "sepolia"];

  function run() public {
    address expectedContractAddress = computeCreateAddress(msg.sender, EXPECTED_NONCE);

    for (uint256 i; i < networks.length; i++) {
      vm.createSelectFork(getChain(networks[i]).rpcUrl);
      bool isDeployed = address(expectedContractAddress).code.length > 0;

      if (isDeployed) {
        console2.log("Skipping '%s': contract already deployed", networks[i]);
        continue;
      }

      uint256 nonce = vm.getNonce(msg.sender);
      if (nonce > EXPECTED_NONCE) {
        console2.log(
          "Skipping '%s': current nonce %d > expected nonce %d", networks[i], nonce, EXPECTED_NONCE
        );
        continue;
      }

      if (nonce < EXPECTED_NONCE) {
        console2.log(
          "Skipping '%s': current nonce %d < expected nonce %d", networks[i], nonce, EXPECTED_NONCE
        );
        continue;
      }

      deploy();
      console2.log("Deployed contract to '%s' at %s", networks[i], expectedContractAddress);
    }
  }
}
```

contracts-periphery/script/Deploy.s.sol
```diff
@@ -1,15 +1,35 @@
...
+/// @notice Run the script with `forge script Deploy --private-key <privateKey>`
+/// and add `--broadcast` to broadcast the tx.
+/// @dev The contract that's being deployed is specified in `BaseScript.sol`.
+/// Specify additional netowrk aliases in [rpc_enpoints] in `forge.toml`
+/// before adding them to the list of networks below.
+/// Add corresponding `env` variables in a `.env` file, such as ETH_RPC_URL=<URL>
 
-contract Deploy is Script {
-  address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
+import {StdChains} from "forge-std/StdChains.sol";
```
mds1, line 12: Unused import
```suggestion
```

garyghayrat: removed in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/663b1c2987ea5cb67ed057e61fff71d69ebf60bd

contracts-periphery/script/Deploy.s.sol
garyghayrat: I see, thank you! should be fixed in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/663b1c2987ea5cb67ed057e61fff71d69ebf60bd

contracts-periphery/script/Deploy.s.sol
garyghayrat: Makes sense, thanks! https://github.com/ScopeLift/umbra-protocol/pull/504/commits/663b1c2987ea5cb67ed057e61fff71d69ebf60bd

contracts-periphery/script/Deploy.s.sol
garyghayrat: I was making the assumption that the nonce will be the same across chains since we're trying to deploy the contract to have the same addresses. But even then, after `console2.log`ging I see how 1. can be an issue. And 2. was definitely a big miss.

My only question then for the smoothest DX, should we ask the deployer to manually edit the script file to enter the `EXPECTED_NONCE`, or should we make it an argument in `run()`, or is there a better way?

mds1: Just hardcode IMO. Once it’s set it should never change, so let’s keep it simple. Once the value is set and shouldn’t be changed we can update the code comment to reflect that

garyghayrat: Cool. Updated here: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/9bfbf40edf11a8894d7efadeb07a8b313fee229a

contracts-periphery/script/Deploy.s.sol
garyghayrat: Thanks for the clarification! Ik something didn't feel right cause the base contract seemed too specific to `batchSend` in this case, but your explanation made it clear. I've merged the scripts for now with the `DeployBatchSend.s.sol` name here: https://github.com/ScopeLift/umbra-protocol/commit/9bfbf40edf11a8894d7efadeb07a8b313fee229a

contracts-periphery/script/DeployBatchSend.s.sol
```diff
@@ -0,0 +1,69 @@
...
+/// which you can get using `cast nonce --rpc-url <url> <address>`
+/// @dev Run the script with `forge script Deploy --private-key <privateKey>`
+/// and add `--broadcast` to broadcast the tx. We *MUST* deploy using the `--private-key`
+/// flag to ensure `msg.sender` is the deployer address, which is used to check the
+/// nonce and compute the expected contract address.
+/// @dev Specify additional network aliases in [rpc_endpoints] in `forge.toml`
+/// before adding them to the list of networks below, and add corresponding environment
+/// variables to the `.env` and `.env.template` files.
+contract Deploy is Script {
+  uint256 constant EXPECTED_NONCE = 0; // Edit this with the nonce of the deployer address
```
mds1, line 21: let's make it obvious there's an outstanding TODO here
```suggestion
  uint256 constant EXPECTED_NONCE = 0; // TODO Edit this with the nonce of the deployer address
```

contracts-periphery/script/DeployBatchSend.s.sol
```diff
@@ -0,0 +1,69 @@
...
+
+      deploy();
+      console2.log("Deployed contract to '%s' at %s", networks[i], expectedContractAddress);
+    }
+  }
+
+  function deploy() private {
+    vm.broadcast();
+    new UmbraBatchSend(IUmbra(UMBRA));
+  }
```
mds1, line 68: I'd just inline this into the other method for simplicity, since this method is super simple anyway

garyghayrat: Sounds good https://github.com/ScopeLift/umbra-protocol/pull/504/commits/b62b1a08e70fd154776ac2507ef855007bb00668

contracts-periphery/script/DeployBatchSend.s.sol
garyghayrat: Added https://github.com/ScopeLift/umbra-protocol/pull/504/commits/b62b1a08e70fd154776ac2507ef855007bb00668

contracts-periphery/test/DeployBatchSend.t.sol
```diff
@@ -0,0 +1,25 @@
...
+  function setUp() public {
+    assertEq(EXPECTED_NONCE, EXPECTED_TEST_NONCE);
+    assertTrue(expectedContractAddress.code.length == 0);
+  }
+
+  function test_checkContractIsDeployed() public {
+    console2.log("expectedContractAddress is ", expectedContractAddress);
+    run();
+    for (uint256 i; i < networks.length; i++) {
+      vm.createSelectFork(getChain(networks[i]).rpcUrl);
```
mds1, line 21: didn't we already create the forks in the `run` method? here we should just be selecting an already existing fork, not creating a new one

contracts-periphery/test/DeployBatchSend.t.sol
```diff
@@ -0,0 +1,25 @@
...
+import {console2} from "forge-std/console2.sol";
+
+contract DeployBatchSendTest is Deploy, Test {
+  uint256 constant EXPECTED_TEST_NONCE = 4; // TODO Edit this with the nonce of the
+  address expectedContractAddress = computeCreateAddress(msg.sender, EXPECTED_NONCE);
+
+  function setUp() public {
+    assertEq(EXPECTED_NONCE, EXPECTED_TEST_NONCE);
+    assertTrue(expectedContractAddress.code.length == 0);
+  }
```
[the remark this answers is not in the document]
garyghayrat: Should be fixed in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/964caa73ebe402ebe5f5e0640cb87f7f5137c9fc

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: Latest test in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/964caa73ebe402ebe5f5e0640cb87f7f5137c9fc

contracts-periphery/Makefile
```diff
@@ -10,7 +10,8 @@ endif
 build :; forge build
 clean :; forge clean
 install :; $(INSTALL_CMD)
-test :; forge test
+test :; forge test --nmc DeployBatchSendTest && make test-scripts
 test-gas :; forge test --match-path *.gas.t.sol
+test-scripts :; forge test --mc DeployBatchSendTest --sender 0x4f78F7f3482D9f1790649f9DD18Eec5A1Cc70F86
```
mds1, line 15: Do we need two separate commands here, or can we just use that sender for all tests and change the `test` command to use that sender? Ideally we wouldn't need to fragment things.

If we do fragment things, the `test` command in package.json needs to be updated because it only runs `make test`

contracts-periphery/foundry.toml
```diff
@@ -21,7 +21,12 @@
   invariant = { runs = 1000 }
 
 [rpc_endpoints]
+  arbitrum = "${ARBITRUM_RPC_URL}"
```
mds1, line 24: In `.env.example` you call it `ARBITRUM_ONE_RPC_URL`. I believe StdChains would also look for `ARBITRUM_ONE_RPC_URL` so that's probably what we should use

contracts-periphery/test/DeployBatchSend.t.sol
```diff
@@ -0,0 +1,31 @@
...
+import {Test} from "forge-std/Test.sol";
+import {console2} from "forge-std/console2.sol";
+import {UmbraBatchSend} from "src/UmbraBatchSend.sol";
+import {IUmbra} from "src/interface/IUmbra.sol";
+
+contract DeployBatchSendTest is DeployBatchSend, Test {
+  uint256 constant EXPECTED_TEST_NONCE = 0; // TODO Edit this with the nonce of the
+  address expectedContractAddress = computeCreateAddress(msg.sender, EXPECTED_NONCE);
+  UmbraBatchSend umbraBatchSend = new UmbraBatchSend(IUmbra(UMBRA));
+  bytes batchSendCode = address(umbraBatchSend).code;
```
mds1, line 14: Can this all be moved to the `setUp` method instead? Not really best practice to have logic like contract deploys in the constructor like this, `setUp` is more clear and conventional

contracts-periphery/test/DeployBatchSend.t.sol
```diff
@@ -0,0 +1,31 @@
...
+
+  function test_checkNonce() public {
+    assertEq(EXPECTED_NONCE, EXPECTED_TEST_NONCE);
```
mds1, line 17: What is this really testing? Just that we updated a hardcoded value in two places? Why not use the same hardcoded value from the script in these tests so we don't need to define it twice

contracts-periphery/test/DeployBatchSend.t.sol
```diff
@@ -0,0 +1,31 @@
...
+    assertTrue(expectedContractAddress.code.length == 0);
```
mds1, line 18: Similarly this is testing the contract isn't deployed on forge's vm by default, which seems like a weird thing to test, so maybe you were trying to test something else here?

contracts-periphery/foundry.toml
garyghayrat: great catch!

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: True true, just using `EXPECTED_NONCE` from the script directly now since we inherit it.

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: Moved to `setUp`

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ed081a587b4cf48808b00edf7bb85c4b932608ad

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ed081a587b4cf48808b00edf7bb85c4b932608ad

contracts-periphery/test/DeployBatchSend.t.sol
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ed081a587b4cf48808b00edf7bb85c4b932608ad

contracts-periphery/foundry.toml
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ed081a587b4cf48808b00edf7bb85c4b932608ad

contracts-periphery/Makefile
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ed081a587b4cf48808b00edf7bb85c4b932608ad
 Using teh same sender for all forge tests is causing a couple `batchSend` tests to fail... gonna look into it and report back

.github/workflows/ci.yaml
```diff
@@ -12,7 +12,12 @@ env:
   OPTIMISTIC_ETHERSCAN_API_KEY: ${{ secrets.OPTIMISTIC_ETHERSCAN_API_KEY }}
```
apbendi, line 12: *Arbitrary comment location*

Overall looking great @garyghayrat and simulation worked locally. Nice job! One nit: can we add some notes to the contracts-periphery/README file about how to use the script? Thanks!

garyghayrat: Thanks! Update `README` here [@netlify
Header rules - jolly-shaw-20fe62 Completed in 2m — No header rules processed
Required
[Details](https://github.com/ScopeLift/umbra-protocol/pull/504/checks?check_run_id=13327010322)](https://github.com/ScopeLift/umbra-protocol/pull/504/commits/a22e1d7c7ac9757fabeff71f5a1cf36e4b885ad7)

contracts-periphery/README.md
```diff
@@ -3,9 +3,31 @@
...
+- `cast nonce <Deployer's address> --rpc-url <URL>` to find your nonce
+
+4. Change `EXPECTED_NONCE` of `DeployBatchSend` script to match your nonce.
+5. To make sure the script test passes, run
+
+- `forge test --mc DeployBatchSendTest --sender <Deployer's address>`
+
+6. Pass in the `private key` to dry run deploying the contract across 6 networks. You should see gas estimates for 6 networks, otherwise the address might be taken.
+
+- `forge script Deploy --private-key <Private Key> `
```
apbendi, line 29: I think this should say "DeployBatchSend" right?

contracts-periphery/README.md
```diff
@@ -3,9 +3,31 @@
...
+
+7. Actually deploy the contracts.
+
+- `forge script Deploy --private-key <Private Key> --broadcast`
```
apbendi, line 33: Same as above, should say "DeployBatchSend" I believe

contracts-periphery/README.md
garyghayrat: Yes! Fixed in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/473efe7ffda600f5bc3599dd2e64606410b79fc9

contracts-periphery/README.md
garyghayrat: Fixed in https://github.com/ScopeLift/umbra-protocol/pull/504/commits/473efe7ffda600f5bc3599dd2e64606410b79fc9

contracts-periphery/Makefile
mds1: Looks like there's now only one test command—what was the issue here?

contracts-periphery/README.md
```diff
@@ -3,9 +3,31 @@
...
+2. In your terminal use command
+
+- `cast nonce <Deployer's address> --rpc-url <URL>` to find your nonce
+
+4. Change `EXPECTED_NONCE` of `DeployBatchSend` script to match your nonce.
+5. To make sure the script test passes, run
+
+- `forge test --mc DeployBatchSendTest --sender <Deployer's address>`
+
+6. Pass in the `private key` to dry run deploying the contract across 6 networks. You should see gas estimates for 6 networks, otherwise the address might be taken.
```
mds1, line 27: I think you are trying to say either (1) there's already code at the address, or (2) there was a mismatch between current and expected nonce? If so maybe we clarify a bit, since "the address might be taken" isn't clear (e.g. it could refer to the deployer addr or contract addr, and what does "taken" mean?)

contracts-periphery/Makefile
garyghayrat: What do you mean? Should I have kept the `test-scripts` one?

garyghayrat: > [ed081a5](https://github.com/ScopeLift/umbra-protocol/commit/ed081a587b4cf48808b00edf7bb85c4b932608ad) Using the same sender for all forge tests is causing a couple `batchSend` tests to fail... gonna look into it and report back

This was caused by `expectCall` in batchSend tests, so just fixed them up.

mds1: > Using the same sender for all forge tests is causing a couple batchSend tests to fail... gonna look into it and report back

I was just asking about resolution to this, since you said using the same sender caused the batchSend tests to fail. But sounds like it was really the `expectCall` issue?

contracts-periphery/README.md
garyghayrat: You're right, included both possibilities here: https://github.com/ScopeLift/umbra-protocol/pull/504/commits/ec1753106506070f5e078c6fcc99988ab2009154 And then a minor update https://github.com/ScopeLift/umbra-protocol/pull/504/commits/503836c318fbb3a3a24f473d1517e89b206a79b3

contracts-periphery/README.md
```diff
@@ -3,9 +3,31 @@
...
+
+- `forge script DeployBatchSend --private-key <Private Key> `
+
+7. Actually deploy the contracts.
```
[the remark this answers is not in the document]
garyghayrat: Fixed here: https://github.com/ScopeLift/umbra-protocol/commit/ec1753106506070f5e078c6fcc99988ab2009154

contracts-periphery/test/UmbraBatchSend.t.sol
```diff
@@ -89,19 +89,16 @@ abstract contract UmbraBatchSendTest is DeployUmbraTest {
     uint256 totalAmount = uint256(amount) + amount2;
     uint256 totalAmount2 = uint256(amount3) + amount4;
 
-    sendData.push(UmbraBatchSend.SendData(alice, address(token), amount, pkx, ciphertext));
-    sendData.push(UmbraBatchSend.SendData(bob, address(token), amount2, pkx, ciphertext));
     sendData.push(UmbraBatchSend.SendData(alice, address(token2), amount3, pkx, ciphertext));
     sendData.push(UmbraBatchSend.SendData(bob, address(token2), amount4, pkx, ciphertext));
+    sendData.push(UmbraBatchSend.SendData(alice, address(token), amount, pkx, ciphertext));
+    sendData.push(UmbraBatchSend.SendData(bob, address(token), amount2, pkx, ciphertext));
```
mds1, line 95: Just curious why was this order changed?

mds1 (approved):
LGTM, thanks!

contracts-periphery/test/UmbraBatchSend.t.sol
garyghayrat: I was getting the `NotSorted` error. Switching the order was an easy way to sort it. Which do makes me wonder if the tests were not running properly before yet still passing?

contracts-periphery/Makefile
garyghayrat: Yes! Seems that way

contracts-periphery/test/UmbraBatchSend.t.sol
mds1: If the mock tokens are tokens that we deploy (as opposed to a fork test with constant token addrs), then more likely is that the nonce changed due to the deploy txs that now happen first, and as a result the token addresses changed and were no longer sorted

contracts-periphery/Makefile
mds1: got it, thanks 👌

contracts-periphery/test/UmbraBatchSend.t.sol
garyghayrat: oh cool, that makes sense. Thank you!

apbendi (approved):
Looks good @garyghayrat!

Issue #512, opened 2023-05-22
Fallback to showing standard address if ENS resolution fails


apbendi:
Currently, on the receive table, if there is an error during ENS name resolution it prevents the table from rendering. An example could include resolving a name with a [disallowed codepoint](https://sourcegraph.com/npm/ethersproject/hash/-/blob/src.ts/ens-normalize/lib.ts?L102), which at least one user has reported experiencing.

We should catch errors in name resolution and degrade gracefully by showing the raw address instead if they occur.

PR #518, opened 2023-05-24
fix: handle ENS names with invalid characters

mds1:
Resolves https://github.com/ScopeLift/umbra-protocol/issues/512


apbendi (approved):
Looks good. User reports the fix works for them. Local testing seems normal. Thanks for tracking this down!

PR #520, opened 2023-05-25
Hook up advanced block filtering to the graph subgraph

### LINKED ISSUE #513: Fix: fetching with block number range in advanced mode
Currently, the advanced mode option for fetching Announcements within a certain block range is broken. I believe it has been broken for a while and needs to be implemented for The Graph. In addition to being a useful feature for advanced mode users, it will also be needed to implement local Announcement caching strategies, so we ought to fix it.

Screen Shot 2023-05-22 at 12 07 45 PM

alexkeating:
#### Description

- A user can now specify the start and end block in umbra and see the results for only those blocks

Closes #513

### DIFF
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -176,9 +176,13 @@ function useScan() {
     // Fetch announcements
     const overrides = { startBlock: startBlockLocal.value, endBlock: endBlockLocal.value };
     let allAnnouncements: AnnouncementDetail[] = [];
+    console.log('overrides', overrides);
+    // Have to build and do a relative path to fix
     try {
       allAnnouncements = await umbra.value.fetchAllAnnouncements(overrides);
+      console.log('Fetching done');
     } catch (e) {
+      console.log('error');
       scanStatus.value = 'waiting'; // reset to the default state because we were unable to fetch announcements
       throw e;
     }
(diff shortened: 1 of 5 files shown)


frontend/src/pages/AccountReceive.vue
```diff
@@ -135,7 +135,8 @@ function useScan() {
       const isFormValid = await settingsFormRef.value?.validate(true);
```
apbendi, line 135: @alexkeating this is working great! Nice work.

One downside I noticed testing it is I don't think we have a way to get to the scan settings until *after* the scanning is complete. (Am I wrong about that?). If you're trying to use block restrictions to save time it doesn't really do much good.

What do you think of adding something like what I have mocked up below to the fetching/scanning screens, in advanced mode only. The sentence shown is the same one as shown above the table, just in a different tense.

image

image

If this is easy, we can tack it on to this PR. If it's non-trivial for any reason, let's ship this PR and move this into its own issue.

umbra-js/src/classes/Umbra.ts
```diff
@@ -420,7 +415,12 @@ export class Umbra {
           token
           txHash
         }
-      }`
+      }`,
+      [],
+      {
+        startBlock: startBlock,
+        endBlock: endBlock,
```
mds1, line 422: nit
```suggestion
        startBlock,
        endBlock,
```

mds1 (approved):

umbra-js/src/classes/Umbra.ts
alexkeating: https://github.com/ScopeLift/umbra-protocol/pull/520/commits/61c8907e5845922844186ea5337ddd145004bd3b

frontend/src/pages/AccountReceive.vue
alexkeating: https://github.com/ScopeLift/umbra-protocol/issues/522

apbendi (approved):
Nice work @alexkeating

PR #524, opened 2023-05-31
Add account encryption to batch send

alexkeating:
#### Description

- Add history button to send page
- Store send in account send history

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -451,17 +451,6 @@
             </div>
 
             <!-- Send button -->
-            <div>
-              <router-link
-                v-if="false"
-                :class="{ 'no-text-decoration': true, 'dark-toggle': true }"
-                :to="{ name: 'sent' }"
-              >
-                <div class="row items-center justify-center q-pa-xs link-container">
-                  {{ $t('Send.send-history') }}
-                </div>
-              </router-link>
-            </div>
             <div class="batch-send-buttons">
               <base-button
                 class="button-container"
@@ -476,6 +465,13 @@
                 type="submit"
                 :disable="batchSends.length == 1 || isSending"
               />
+              <div class="q-mt-sm">
+                <router-link :class="{ 'no-text-decoration': true, 'dark-toggle': true }" :to="{ name: 'sent' }">
+                  <div class="row items-center justify-center q-pa-sm link-container rounded-borders">
+                    {{ $t('Send.send-history') }}
+                  </div>
+                </router-link>
+              </div>
             </div>
           </q-form>
         </q-page>
@@ -517,7 +513,7 @@ import { SendBatch } from '@umbracash/umbra-js';
 import { Provider, TokenInfoExtended, supportedChains } from 'components/models';
 import { ERC20_ABI } from 'src/utils/constants';
 import { toAddress } from 'src/utils/address';
-import { storeSend } from 'src/utils/account-send';
+import { batchStoreSend, storeSend } from 'src/utils/account-send';
 
 interface BatchSendData {
   id: number;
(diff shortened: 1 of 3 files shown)


frontend/src/utils/account-send.ts
```diff
@@ -52,6 +52,16 @@ export type StoreSendArgs = {
...
 
+type BatchStoreSendArgs = {
+  chainId: number;
+  viewingPrivateKey: string;
+  advancedMode: boolean;
+  usePublicKeyChecked: boolean;
+  batches: { batch: SendBatch; pubKey: string }[];
+  batchTxHash: string;
+  senderAddress: string;
+};
```
mds1, line 63: Can this just be `type BatchStoreSendArgs = StoreSendArgs[]`? And really in that case we don't even need a new `BatchStoreSendArgs` type. Arguably it's less efficient since we're duplicating tx hash/etc, but that difference will be negligible, and it feels simpler that way

frontend/src/pages/AccountSend.vue
```diff
@@ -476,6 +465,13 @@
                 type="submit"
                 :disable="batchSends.length == 1 || isSending"
               />
+              <div class="q-mt-sm">
+                <router-link :class="{ 'no-text-decoration': true, 'dark-toggle': true }" :to="{ name: 'sent' }">
+                  <div class="row items-center justify-center q-pa-sm link-container rounded-borders">
+                    {{ $t('Send.send-history') }}
+                  </div>
+                </router-link>
+              </div>
```
mds1, line 474: was this element changed or just moved?

frontend/src/pages/AccountSend.vue
```diff
@@ -1067,6 +1067,36 @@ function useSendForm() {
...
+          const batches = [];
+          for (const item of values) {
+            if (item.status === 'fulfilled') {
+              batches.push(item.value as { batch: SendBatch; pubKey: string });
+            }
+          }
+          return batches;
+        });
+
+        await batchStoreSend({
```
mds1, line 1090: related to the above, wondering if we can simplify/dedupe some code by always using the batch version evne if the length of sends is just 1 (the single send)? maybe the vars we use above are too discrete to enable  this

frontend/src/pages/AccountSend.vue
alexkeating: Yes, it was feature flagged to be hidden so it wasn't showing up on the frontend. I moved it to be below the send button like we have it on the regular send page, but also happy to move if that looks weird.

frontend/src/utils/account-send.ts
alexkeating: In this case I think we could combine both functions since we are already iterating through all of the sends before hand to lookup the public key. The only downside is duplication of data along with transforming the batch send data which we are already doing somewhat.

frontend/src/pages/AccountSend.vue
alexkeating: Related https://github.com/ScopeLift/umbra-protocol/pull/524#discussion_r1214469868

alexkeating: I removed batch and handled everything in store https://github.com/ScopeLift/umbra-protocol/pull/524/commits/6e68909bd31bc23e5bcb2037a89dd643a8faeb5b

frontend/src/utils/account-send.ts
alexkeating: Instead I removed batch store and consolidated here https://github.com/ScopeLift/umbra-protocol/pull/524/commits/6e68909bd31bc23e5bcb2037a89dd643a8faeb5b

mds1: nice this looks a lot simpler now, thanks!

mds1 (commented):
code lgtm! have not tested yet so did not approve

PR #528, opened 2023-06-01
Fix batchSend summary

garyghayrat:
Fixes Umbra fee not being adding to summary total amount
Before:
Screen Shot 2023-06-05 at 1 02 41 PM
After:
Screen Shot 2023-06-05 at 1 04 46 PM

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -730,7 +730,7 @@ function useSendForm() {
       } else if (nativeTokenValue.chainId !== prevNativeTokenValue.chainId || isLoadingValue !== prevIsLoadingValue) {
         // When network finally connects after page load, we need to re-parse the payment link data.
         await setPaymentLinkData(); // Handles validations.
-        const chainId = BigNumber.from(currentChain.value?.chainId).toNumber();
+        const chainId = BigNumber.from(currentChain.value?.chainId || 0).toNumber();
         batchSendIsSupported.value = batchSendSupportedChains.includes(chainId);
         if (batchSends.value.length > 0) {
           batchSends.value[0].token = nativeToken.value;
@@ -749,7 +749,7 @@ function useSendForm() {
       { id: 1, receiver: '', token: nativeToken.value, amount: '' },
       { id: 2, receiver: '', token: nativeToken.value, amount: '' }
     );
-    const chainId = BigNumber.from(currentChain.value?.chainId).toNumber();
+    const chainId = BigNumber.from(currentChain.value?.chainId || 0).toNumber();
     batchSendIsSupported.value = batchSendSupportedChains.includes(chainId);
   });
 
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -572,7 +572,6 @@ function useSendForm() {
   const recipientIdWarning = ref<string>();
   const useNormalPubKey = ref(false);
   const token = ref<TokenInfoExtended>();
-  const nativeToken = ref<TokenInfoExtended>();
   const humanAmount = ref<string>();
   const isValidForm = ref(false);
   const isValidRecipientId = ref(true); // for showing/hiding bottom space (error message div) under input field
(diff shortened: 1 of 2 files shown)


frontend/src/pages/AccountSend.vue
```diff
@@ -743,8 +748,13 @@ function useSendForm() {
     if (amount) humanAmount.value = amount;
 
     // For token, we always default to the chain's native token if none was selected
-    if (paymentToken?.symbol) token.value = paymentToken;
-    else token.value = tokenList.value[0];
+    if (paymentToken?.symbol) {
+      token.value = paymentToken;
+      nativeToken.value = paymentToken;
```
mds1, line 753: I don't understand the code changes on in this PR:
- `NATIVE_TOKEN` is a constant, defined per-chain
- `nativeToken, based on the variable name, sounds like it would be identical
- But it's not because if we have a payment link we set native token to the token in the payment link?
- So now we have:
    - `NATIVE_TOKEN` which is always the native token
    - `token` which is the selected token
    - `nativeToken` which is also the selected token?

mds1: Possibly related, when I connect to sepolia and directly load http://localhost:8080/send I see an error

image

garyghayrat: - So this was my work around to get `NATIVE_TOKEN` as `TokenInfoExtedned`, which directly casting it using `as` didn't work, I created `nativeToken` to be the `TokenInfoExtended` native token. 
- Payment link looks like a bug created from trying to use the same `setPaymentLinkData` function to set the default token.

mds1: would just doing `NATIVE_TOKEN.value as unknown as TokenInfoExtended` fix the issue without needing a new variable? if so, you should add a comment about why the intermediate cast to `unknown` is safe, presumably here it's because we don't need the `minSendAmount` field of `TokenInfoExtended` and we just need the type as `TokenInfoExtended` for reason XYZ

garyghayrat: Fixed the error here: https://github.com/ScopeLift/umbra-protocol/pull/528/commits/8cc3c59e4f6901de1f6a8029e4f94ebb632128d9

mds1: It seems we still have the new `nativeToken` variable and it's not clear to me why, would be great if we can remove that

garyghayrat: Sounds good, i'll work on that next

garyghayrat: Removed `nativeToken` here: https://github.com/ScopeLift/umbra-protocol/pull/528/commits/4e9622b4008cd0783bf0ba3ac96c380bf8658cb6

apbendi (approved):
LGTM

frontend/src/pages/AccountSend.vue
```diff
@@ -600,15 +600,20 @@ function useSendForm() {
 
   // Batch Send Computed Form Parameters
   const batchSendHumanTotalAmount = computed(() => {
-    const ethTotal = summaryAmount.value.get(NATIVE_TOKEN.value);
+    let ethTotal;
+    for (const token of summaryAmount.value.keys()) {
+      if (token.symbol === NATIVE_TOKEN.value.symbol) {
+        ethTotal = summaryAmount.value.get(token);
+      }
+    }
```
mds1, line 608: why was this changed? isn't calling `.get` for a only single condition equivalent to what we had if we tweaked it like this:

```suggestion
const ethTotal = summaryAmount.value.get(NATIVE_TOKEN.value) || '0'
```

garyghayrat: This is actually relevant to this https://github.com/ScopeLift/umbra-protocol/pull/528#discussion_r1218364190 to `.get` you need to provide a `TokenInfoExtended` and `NATIVE_TOKEN` isn't one. 

And casting it as `unknown` as `TokenInfoExtended` didn't work because it looks like the `LogoURI` is different between `NATIVE_TOKEN` and the selected native token. I tried changing that to be the same but still it was `NATIVE_TOKEN` couldn't be used to get from the `map.` 

Felt like it wasn't worth it to further investigate and this was my solution because we know the native token's `symbol` will be the same.

Issue #316, opened 2022-03-30
bug: receive table does not always sort properly


mds1:
> It seems the received funds are not sorted properly. You can see the most recently received funds are at the bottom, but they should be at the top
> ![](https://user-images.githubusercontent.com/17163988/159098931-db84e14e-a19e-4919-93e0-92e4cd5fbb02.png)

> So weird. I saw this a bit testing last week, but can't reproduce it today. My guess is it's a race-condition issue of some kind, and related to web workers. I think make an issue for this one and we can tackle it at some point later.

_Originally posted in https://github.com/ScopeLift/umbra-protocol/pull/293#discussion_r830413250_

PR #536, opened 2023-06-08
Add Gnosis Chain support

garyghayrat:


frontend/src/components/models.ts
```diff
@@ -100,6 +100,21 @@ export const supportedChains: Array<Chain> = [
...
+      address: NATIVE_TOKEN_ADDRESS,
+      name: 'xDAI',
+      symbol: 'xDAI',
+      decimals: 18,
+      logoURI: 'https://docs.gnosischain.com/img/tokens/xdai.png',
+    },
+    rpcUrls: ['https://rpc.ankr.com/gnosis'],
+    blockExplorerUrls: ['https://gnosisscan.io'],
+    iconUrls: ['https://docs.gnosischain.com/img/tokens/gno.png'],
+    logoURI: 'https://docs.gnosischain.com/img/tokens/gno.png',
```
mds1, line 116: I think that might be the icon for the GNO token for staking, but we want the chain logo which I think is green in color? https://uploads-ssl.webflow.com/63692bf32544bee8b1836ea6/636a6e764bdb11a70341fab4_owl-forest.png

frontend/src/components/models.ts
```diff
@@ -72,7 +72,7 @@ export const supportedChains: Array<Chain> = [
   },
   {
     chainId: '0xaa36a7',
-    chainName: 'Sepolia',
+    chainName: 'Sepolia', // 11155111 as hex
```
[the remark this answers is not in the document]
garyghayrat: Fixed it in https://github.com/ScopeLift/umbra-protocol/pull/536/commits/7c12ebbe61e9afed82647c1deafb4e0afaf1f22c

frontend/src/components/models.ts
garyghayrat: You're absolutely right! Updated the logo here: https://github.com/ScopeLift/umbra-protocol/pull/536/commits/7c12ebbe61e9afed82647c1deafb4e0afaf1f22c

frontend/src/components/models.ts
```diff
@@ -71,7 +71,7 @@ export const supportedChains: Array<Chain> = [
     logoURI: ETH_NETWORK_LOGO,
```
apbendi, line 71: *Arbitrary comment location*

Given that the toll is 0.05 xDAI, I think the minimum send amount should probably be set to something like 0.5 xDAI


Screen Shot 2023-06-13 at 3 25 27 PM

frontend/src/components/models.ts
```diff
@@ -71,7 +71,7 @@ export const supportedChains: Array<Chain> = [
...
   },
   {
```
apbendi, line 73: *arbitrary comment location*

Sorting by date received is broken on the receive table for Gnosis Chain only (it seems).

Sends are sorted with oldest first (should be newest), and the table order does not react to clicking the "Date Received" header.

We've [seen this before](https://github.com/ScopeLift/umbra-protocol/issues/316) but couldn't pin it down. Is this the same thing?

apbendi (commented):
Hey @garyghayrat, this is working great overall! Nice job. I did notice a couple things while testing it more extensively. One should be a quick change (min send amount) but the other is a weird sorting bug—maybe one we've seen before?

frontend/src/components/models.ts
apbendi: Screen Shot 2023-06-13 at 3 33 56 PM

garyghayrat: Sorting by Date Received is fix by https://github.com/ScopeLift/umbra-protocol/pull/536/commits/b17ee653db20221a7f169c22aedfdbefbe21852b
Having trouble reproducing wrong sort result, but added additional sorting https://github.com/ScopeLift/umbra-protocol/pull/536/commits/e27b2cd2c250a722cef8a8cab2c3d069c0651b8e to try to prevent it from occurring too often

apbendi (approved):
Looking good @garyghayrat!

PR #543, opened 2023-06-22
Fix overlapping messages

### LINKED ISSUE #525: Negative margins are causing formatting issues
When directly copying in an ENS with confusables or directly copying an ENS name that is not supported the negative makes the error message difficult to read. It is probably a race condition.

image

_Originally posted by @mds1 in https://github.com/ScopeLift/umbra-protocol/pull/523#discussion_r1211999580_

alexkeating:
### Description

- There is a gap, but there shouldn't be overlapping messages.

[Screencast from 2023-06-22 13-01-51.webm](https://github.com/ScopeLift/umbra-protocol/assets/12705146/64506a78-3981-4cdc-b5d3-fa30250fc65d)

Closes #525

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -115,25 +115,16 @@
             :disable="isSending"
             placeholder="vitalik.eth"
             lazy-rules
+            :hideBottomSpace="true"
             :rules="isValidId"
             ref="recipientIdBaseInputRef"
           />
-          <div
-            class="flex q-pt-sm text-caption warning-container"
-            v-if="recipientIdWarning"
-            :style="!recipientId || isValidRecipientId ? 'margin-top:-2em' : ''"
-          >
+          <div class="flex row text-caption warning-container q-pb-sm" v-if="recipientIdWarning">
             {{ recipientIdWarning }}
           </div>
 
           <!-- Identifier, advanced mode tooltip -->
-          <div
-            v-if="advancedMode"
-            class="row items-center text-caption q-pt-sm q-pb-lg"
-            :style="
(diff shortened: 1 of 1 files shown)


garyghayrat (approved):
lgtm! Seems like only 'm' from the alphabet is considered a confusable character?

frontend/src/pages/AccountSend.vue
```diff
@@ -115,25 +115,16 @@
...
 
           <!-- Identifier, advanced mode tooltip -->
-          <div
-            v-if="advancedMode"
-            class="row items-center text-caption q-pt-sm q-pb-lg"
-            :style="
-              !recipientId || isValidRecipientId || (!isValidRecipientId && recipientIdWarning) ? 'margin-top:-2em' : ''
-            "
-          >
+          <div v-if="advancedMode" class="row items-center text-caption q-pt-sm">
```
alexkeating, line 127: (arbitrary thread)

@garyghayrat Yes, it seems like some ascii characters are marked as confusable. m is considered confusable because it is close to rn. I think it is overly strict and it may make sense to create an issue to relax. @mds1 Any thoughts on this?

mds1: Hmm my initial reaction is to agree it's overly strict for our use case. Does ENS or ensjs have guidance for apps on how they recommend handling this? Wonder what most popular apps do here

alexkeating: ENS and ensjs handle detecting confusables in the normalization process. They have an esip and library we could use, but it is heavier than only detecting confusables.

https://docs.ens.domains/ens-improvement-proposals/ensip-15-normalization-standard#wholes

https://github.com/adraffy/ens-normalize.js/tree/main

apbendi (approved):
LGTM thanks for resolving this!

PR #544, opened 2023-06-23
Add provider so copy works again

### LINKED ISSUE #501: Copying the Stealth Receiver and Sender Address in Receive Table is broken
When viewing transaction history on the Receive Screen, the address shown in the Sender column and the Stealth Address column can be clicked such that the address in question will be copied to the clipboard. Currently this functionality seems to be broken. While the icon shows on mouse over, clicking does not result in the address being copied.

alexkeating:
#### Description

- Fix copy behavior on the receive table

Closes #501


garyghayrat (approved):
It's working again!

mds1 (approved):
Tested and confirmed it works

apbendi (approved):
LGTM! Just needs a rebase before merging @alexkeating

Issue #266, opened 2022-01-06
Umbra Scanning Bottleneck Possible Mitigations


apbendi:
Meta issue to document possible mitigations to the bottleneck of announcement scanning:

- [x] Async w/ progress bar (notification when completed)
- [x] Parrallelization
- [ ] Rust based crypto
- [x] Only scanning blocks after when the user set their stealth keys
- [ ] Caching previously decrypted messages
- [x] Lazy loading in UI (show detected announcements as they're found rather than waiting for all of them)
- [ ] Sharding the receive table between withdrawn/non-withdrawn
- [ ] Optional server side scanning via viewing key

mds1:
#377 is another possible mitigation but requires a new contract. We may want a new contract anyway to implement that + NFT support ([1](https://twitter.com/VitalikButerin/status/1556548617636556800), [2](https://ethresear.ch/t/erc721-extension-for-zk-snarks/13237/2))

PR #549, opened 2023-06-26
Scan from user registered block

garyghayrat:
One of the improvements listed in https://github.com/ScopeLift/umbra-protocol/issues/266

### DIFF
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -37,14 +37,14 @@
               <div class="row justify-start q-col-gutter-md">
                 <base-input
                   v-model.number="startBlockLocal"
-                  @blur="setScanBlocks(startBlockLocal, endBlockLocal)"
+                  @blur="setScanBlocks(startBlockLocal!, endBlockLocal!)"
                   class="col-xs-12 col-6"
                   :label="$t('Receive.start-block')"
                   :rules="isValidStartBlock"
                 />
                 <base-input
                   v-model.number="endBlockLocal"
-                  @blur="setScanBlocks(startBlockLocal, endBlockLocal)"
+                  @blur="setScanBlocks(startBlockLocal!, endBlockLocal!)"
                   class="col-xs-12 col-6"
                   :label="$t('Receive.end-block')"
                   :rules="isValidEndBlock"
@@ -84,7 +84,7 @@
 import { computed, defineComponent, onMounted, ref, watch } from 'vue';
 import { QForm } from 'quasar';
 import { UserAnnouncement, KeyPair, AnnouncementDetail, utils } from '@umbracash/umbra-js';
-import { BigNumber, JsonRpcSigner, computeAddress, isHexString } from 'src/utils/ethers';
+import { BigNumber, computeAddress, isHexString } from 'src/utils/ethers';
 import useSettingsStore from 'src/store/settings';
 import useWalletStore from 'src/store/wallet';
 import useWallet from 'src/store/wallet';
@@ -180,11 +180,9 @@ function useScan() {
     const overrides = { startBlock: startBlockLocal.value, endBlock: endBlockLocal.value };
     let allAnnouncements: AnnouncementDetail[] = [];
     try {
-      allAnnouncements = await umbra.value.fetchSomeAnnouncements(
-        signer.value as JsonRpcSigner,
-        userWalletAddress.value as string,
-        overrides
-      );
+      if (!signer.value) throw new Error('signer is undefined');
+      if (!userWalletAddress.value) throw new Error('userWalletAddress is undefined');
+      allAnnouncements = await umbra.value.fetchSomeAnnouncements(signer.value, userWalletAddress.value, overrides);
     } catch (e) {
       scanStatus.value = 'waiting'; // reset to the default state because we were unable to fetch announcements
       throw e;
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -182,7 +182,11 @@ function useScan() {
     try {
       if (!signer.value) throw new Error('signer is undefined');
       if (!userWalletAddress.value) throw new Error('userWalletAddress is undefined');
-      allAnnouncements = await umbra.value.fetchSomeAnnouncements(signer.value, userWalletAddress.value, overrides);
+      // When private key is provided in advanced mode, we fetch all announcements
+      if (advancedMode.value && scanPrivateKey.value)
+        allAnnouncements = await umbra.value.fetchAllAnnouncements(overrides);
+      else
+        allAnnouncements = await umbra.value.fetchSomeAnnouncements(signer.value, userWalletAddress.value, overrides);
     } catch (e) {
       scanStatus.value = 'waiting'; // reset to the default state because we were unable to fetch announcements
       throw e;
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -84,8 +84,9 @@
 import { computed, defineComponent, onMounted, ref, watch } from 'vue';
 import { QForm } from 'quasar';
 import { UserAnnouncement, KeyPair, AnnouncementDetail, utils } from '@umbracash/umbra-js';
-import { BigNumber, computeAddress, isHexString } from 'src/utils/ethers';
+import { BigNumber, JsonRpcSigner, computeAddress, isHexString } from 'src/utils/ethers';
 import useSettingsStore from 'src/store/settings';
+import useWalletStore from 'src/store/wallet';
 import useWallet from 'src/store/wallet';
 import { filterUserAnnouncements } from 'src/worker/worker';
 import AccountReceiveTable from 'components/AccountReceiveTable.vue';
@@ -101,6 +102,7 @@ function useScan() {
   // Start and end blocks for advanced mode settings
   const { advancedMode, startBlock, endBlock, setScanBlocks, setScanPrivateKey, scanPrivateKey, resetScanSettings } =
     useSettingsStore();
+  const { signer, userAddress: userWalletAddress } = useWalletStore();
   const startBlockLocal = ref<number>();
   const endBlockLocal = ref<number>();
   const scanPrivateKeyLocal = ref<string>();
@@ -178,7 +180,11 @@ function useScan() {
     const overrides = { startBlock: startBlockLocal.value, endBlock: endBlockLocal.value };
     let allAnnouncements: AnnouncementDetail[] = [];
     try {
-      allAnnouncements = await umbra.value.fetchAllAnnouncements(overrides);
+      allAnnouncements = await umbra.value.fetchSomeAnnouncements(
+        signer.value as JsonRpcSigner,
+        userWalletAddress.value as string,
+        overrides
+      );
     } catch (e) {
       scanStatus.value = 'waiting'; // reset to the default state because we were unable to fetch announcements
       throw e;
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -411,21 +411,7 @@ export class Umbra {
     // Get start and end blocks to scan events for
     const startBlock = overrides.startBlock || registeredBlockNumebr || this.chainConfig.startBlock;
     const endBlock = overrides.endBlock || 'latest';
-
-    // Try querying events using the Graph, fallback to querying logs.
-    // The Graph fetching uses the browser's `fetch` method to query the subgraph, so we check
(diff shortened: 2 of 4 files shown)


umbra-js/src/classes/Umbra.ts
```diff
@@ -370,8 +370,8 @@ export class Umbra {
   }
 
   /**
-   * @notice Fetches all Umbra event logs using The Graph, if available, falling back to RPC if not
-   * @param overrides Override the start and end block used for scanning; ignored if using The Graph
+   * @notice Fetches all Umbra event logs using Goldsky, if available, falling back to RPC if not
+   * @param overrides Override the start and end block used for scanning; ignored if using Goldsky
```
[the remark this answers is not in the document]
alexkeating: Yes, we no longer ignore using goldsky

umbra-js/src/utils/utils.ts
```diff
@@ -241,6 +242,17 @@ export async function lookupRecipient(
   return { spendingPublicKey: publicKey, viewingPublicKey: publicKey };
 }
 
+export async function getBlockNumberUserRegistered(address: string, startblock = 0, Signer: JsonRpcSigner) {
+  address = getAddress(address); // address input validation
+  const { chainId } = await Signer.provider.getNetwork();
+  const txHistoryProvider = new TxHistoryProvider(chainId);
+  const history = await txHistoryProvider.getHistory(address, startblock);
+  // Address of the StealthKeyRegistry is the same on all supported networks
+  const StealthKeyRegistryAddress = '0x31fe56609C65Cd0C510E7125f051D440424D38f3';
```
alexkeating, line 251: It might be good to move this reference and the reference in `StealthKeyRegistry.ts` to a constants file.

umbra-js/src/classes/Umbra.ts
```diff
@@ -391,7 +397,25 @@ export class Umbra {
...
+   * @returns A list of Announcement events supplemented with additional metadata, such as the sender, block,
+   * timestamp, and txhash
+   */
+  async fetchSomeAnnouncements(
+    Signer: JsonRpcSigner,
+    address: string,
+    overrides: ScanOverrides = {}
+  ): Promise<AnnouncementDetail[]> {
+    const registeredBlockNumebr = await getBlockNumberUserRegistered(address, 0, Signer);
```
alexkeating, line 410: `registeredBlockNumber` instead of `registeredBlockNumebr`

frontend/src/pages/AccountReceive.vue
```diff
@@ -178,7 +180,11 @@ function useScan() {
     const overrides = { startBlock: startBlockLocal.value, endBlock: endBlockLocal.value };
     let allAnnouncements: AnnouncementDetail[] = [];
     try {
-      allAnnouncements = await umbra.value.fetchAllAnnouncements(overrides);
+      allAnnouncements = await umbra.value.fetchSomeAnnouncements(
+        signer.value as JsonRpcSigner,
```
alexkeating, line 184: Rather than cast it might be better to use `!` to indicate it won't be undefined.

alexkeating: And same with the cast below

alexkeating: To be even more strict you wrap this call in an if statement and handle the situation where both these variables are undefined.

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
garyghayrat: Awesome, thanks!

umbra-js/src/classes/Umbra.ts
garyghayrat: nice catch fixed https://github.com/ScopeLift/umbra-protocol/pull/549/commits/5c9d0b6f28897cdd7b868c03da62604dfb1d6f03

frontend/src/pages/AccountReceive.vue
garyghayrat: Addressed here: https://github.com/ScopeLift/umbra-protocol/pull/549/commits/d4c80ebaaeaf08f903008c081b31bd74a2a28a10

umbra-js/src/utils/utils.ts
garyghayrat: Moved it to constants here: https://github.com/ScopeLift/umbra-protocol/pull/549/commits/b9b27f8a7559f6434be4a9356a5578e2f9509861

umbra-js/src/utils/utils.ts
```diff
@@ -241,6 +243,15 @@ export async function lookupRecipient(
...
 }
 
+export async function getBlockNumberUserRegistered(address: string, startblock = 0, Signer: JsonRpcSigner) {
+  address = getAddress(address); // address input validation
+  const { chainId } = await Signer.provider.getNetwork();
+  const txHistoryProvider = new TxHistoryProvider(chainId);
+  const history = await txHistoryProvider.getHistory(address, startblock);
+  const registryBlock = history.find((tx) => tx.to === STEALTH_KEY_REGISTRY_ADDRESS);
+  return registryBlock?.blockNumber;
+}
```
mds1, line 253: An alternative approach here is query the StealthKeyRegistry for the `StealthKeyChanged`, and filter on the registrant, which is an indexed param. So your log query will return all `StealthKeyChanged` events for that user, and you'd return the oldest block number in that list.

You'd do this with `provider.getLogs(filter)` where the filter is something like: 
```typescript
const filter = stealthKeyRegistry.filters.StealthKeyChanged(user, null, null, null, null)`
```

My gut is to prefer the event approach to minimize our reliance on Etherscan's API, and I suspect it will be fast too because it's going to return a lot less data especially for large accounts.

I also suspect that for users with large numbers of txs, `.getHistory` won't return all of them and there may be some pagination, so this approach might fail sometimes

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
mds1: I unresolved this because it seems the comment is still out of date on this branch

umbra-js/src/classes/Umbra.ts
```diff
@@ -391,7 +397,25 @@ export class Umbra {
   }
 
   /**
-   * @notice Fetches all Umbra event logs using The Graph
+   * @notice Fetches Umbra event logs starting from the block user registered their stealth keys in using Goldsky, if available, falling back to RPC if not
```
mds1, line 400: Line length should be capped at 120 chars

garyghayrat: Broke it up here: https://github.com/ScopeLift/umbra-protocol/pull/549/commits/222d7af6be812c9cf7f8d63de8b8413b75bdd7c3

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
garyghayrat: My bad, updated it in https://github.com/ScopeLift/umbra-protocol/pull/549/commits/222d7af6be812c9cf7f8d63de8b8413b75bdd7c3

umbra-js/src/utils/utils.ts
garyghayrat: Made the changes here: `provider.getLogs(filter)` didn't seem to return any logs so used `registry._registry.queryFilter(filter)`. Seems to work just as well?

apbendi (commented):
Looking good so far @garyghayrat, this will be a big UX boost. Identified some edge cases in initial testing.

umbra-js/src/utils/utils.ts
mds1: This looks great, thanks! I see you use `const registryBlock = stealthKeyLog[0].blockNumber`—I changed the address param to `null` to verify that logs are returned by oldest first, meaning we're correct in using index 0 here, but would great for you to verify too (if you didn't already)

Similarly, a nit: `stealthKeyLogs` (plural) is a better var name since this is an array

frontend/src/pages/AccountReceive.vue
```diff
@@ -37,14 +37,14 @@
               <div class="row justify-start q-col-gutter-md">
```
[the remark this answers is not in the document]
garyghayrat: The error was fixed in https://github.com/ScopeLift/umbra-protocol/pull/549/commits/194834e49f178ff58dcb219493fc201df6e508bc Seems like in prod we scan for all announcements when scanning for a not yet registered address. Are we sure we want to just scan for an empty list of announcements? e.g. I'm thinking of the rare cases when a user registers stealth keys twice and how that might change the receive table.

frontend/src/pages/AccountReceive.vue
```diff
@@ -37,14 +37,14 @@
...
                 <base-input
```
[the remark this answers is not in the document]
garyghayrat: Now it will scan for all announcements when advanced mode is on and a private key is provided https://github.com/ScopeLift/umbra-protocol/commit/194834e49f178ff58dcb219493fc201df6e508bc

umbra-js/src/utils/utils.ts
garyghayrat: Thank you! Yes, can confirm the oldest blocks are returned first in the array. Updated var name in https://github.com/ScopeLift/umbra-protocol/pull/549/commits/c7ec660a2d2b4fd0efe601b7374a06597cc7255b

frontend/src/pages/AccountReceive.vue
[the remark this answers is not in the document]
apbendi: Seems to work well, thanks!

frontend/src/pages/AccountReceive.vue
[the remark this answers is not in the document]
garyghayrat: I was just thinking of the edge case we received on Discord -- where the user accidentally re-registers stealth keys and when they scan, they only see their previous txs but not the latest tx that they've sent after they're re-registered. I was wondering if returning an empty announcement[] will cause even their previous txs to disappear from the receive table. I hope that made sense, but I think it's too much of an edge case anyways and agree that it would be a huge waste for most unregistered users to scan for all announcements.

Made the changes here: https://github.com/ScopeLift/umbra-protocol/pull/549/commits/62605e02ed9a75468eda928ed700de7e3c5e85e5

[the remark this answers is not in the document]
apbendi: > where the user accidentally re-registers stealth keys and when they scan, they only see their previous txs but not the latest tx that they've sent after they're re-registered.

The reason the users saw different things in this case is because they are scanning with a different key. For a user tha has called setStealthKeys more than one time, the correct behavior for this PR would be to fetch and scan all announcements from the first time they called it. Is this what the PR is doing?

> I was wondering if returning an empty announcement[] will cause even their previous txs to disappear from the receive table.

Well, it would. But why would we return an empty array for these users. The only time we should return an empty list is if the user has *never* registered stealth keys. Is that what we're doing?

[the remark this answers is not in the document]
garyghayrat: > The reason the users saw different things in this case is because they are scanning with a different key. For a user tha has called setStealthKeys more than one time, the correct behavior for this PR would be to fetch and scan all announcements from the first time they called it. Is this what the PR is doing?

Yes, https://github.com/ScopeLift/umbra-protocol/pull/549#discussion_r1251144933 we're using index 0 of the `stealthKeyLogs` which is first time someone registering a stealth key.

> Well, it would. But why would we return an empty array for these users. The only time we should return an empty list is if the user has _never_ registered stealth keys. Is that what we're doing?

Yes, if `stealthKeyLogs` see the user hasn't registered and if they haven't specified a start block in advanced mode, it will return an empty array.

> I was wondering if returning an empty announcement[] will cause even their previous txs to disappear from the receive table.

Therefore, I think I was wrong to assume this, and it's a non-issue.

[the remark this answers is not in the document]
mds1: > If the user has *never* registered a stealth key, and they're not scanning with a private key in advance mode, then there's no chance they've ever received a payment and there's no point in doing the scanning.

Technically there is still a chance they received a payment. They could have signed the same message and generated the keys (with the umbra SDK, or by just copying out the logged pubkeys from the console on the receive page), and just not registered those keys on-chain. Then they share those keys with the sender directly, which the sender pastes into the app or again used the SDK. Though, this is very unlikely and may have never happened, but just want to clarify it's not "no chance" (unless I misunderstood something)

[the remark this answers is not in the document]
apbendi: Yes this is technically correct @mds1, but at that point we have to assume they are experts who know what they're doing. They could still force the scan with advanced mode -> custom block number.

alexkeating (approved):
Working, for me

apbendi (approved):
Nice work on this @garyghayrat. Will be a big UX improvement for users.

PR #553, opened 2023-06-27
Add gnosis for onboard

alexkeating:
#### Description

- Improve safe app experience


alexkeating:
Closing because it doesn't seem like we can support Gnosis as is. We need to add a warning letting users know.

PR #555, opened 2023-06-28
Use ens normalize instead of confusable

### LINKED ISSUE #551: The warning about ENS Confusables is not shown for ENS names entered on the batch send page

### LINKED ISSUE #552: Relax the strictness of confusable characters
Currently, some ASCII characters are treated as confusable and we want to remove those from being considered confusable while conforming with how other projects treat ENS confusable characters.

_Originally posted by @alexkeating in https://github.com/ScopeLift/umbra-protocol/pull/543#discussion_r1241302670_

alexkeating:
#### Description

- Use ens normalize to detect confusable, illegal characters and illegal mixtures of unicode characters. This should conform better with ENS compared to `unicode-confusables`.

Closes #552

### DIFF
--- a/frontend/src/utils/validation.ts
+++ b/frontend/src/utils/validation.ts
@@ -37,6 +37,10 @@ export const assertValidEnsName = (name: string) => {
   try {
     ens_normalize(name);
   } catch (err) {
+    // A couple example error messages from ens-normalize:
+    // ‌hi.eth - Invalid label "{200C}hi"‎: disallowed character: {200C}
+    // aα.eth - Invalid label "aα"‎: illegal mixture: Latin + Greek "α"‎ {3B1}
+    console.log((err as Error)?.message);
     throw new Error(
       `We detected a${
         (err as Error)?.message.split(':')[1]
--- a/frontend/src/utils/validation.ts
+++ b/frontend/src/utils/validation.ts
@@ -42,9 +42,10 @@ export const assertValidEnsName = (name: string) => {
     // aα.eth - Invalid label "aα"‎: illegal mixture: Latin + Greek "α"‎ {3B1}
     console.log((err as Error)?.message);
     throw new Error(
-      `We detected a${
-        (err as Error)?.message.split(':')[1]
-      } in the ENS name. Check the ENS name in order to avoid a potential scam.`
+      `The given ENS name is invalid due to: ${(err as Error)?.message
+        .split(':')
+        .slice(1)
+        .join('')} in the ENS name. Check the ENS name in order to avoid a potential scam.`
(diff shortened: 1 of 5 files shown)


frontend/src/utils/validation.ts
```diff
@@ -33,9 +33,14 @@ export const assertValidHexString = (hex: string, length: number, errorMsg?: str
...
-  const hasConfusables = isConfusing(name);
-  if (hasConfusables) {
-    throw new Error(errorMsg || 'Name contains confusables');
+export const assertValidEnsName = (name: string) => {
+  try {
+    ens_normalize(name);
+  } catch (err) {
+    throw new Error(
+      `We detected a${
+        (err as Error)?.message.split(':')[1]
```
mds1, line 42: What's going on here, the way this works is that if it can't be normalized the special character is immediately after the colon in the error message? Would be good to show a sample error message as a comment for maintainability purposes, since relying on the format of a deps error message feels a bit flaky

alexkeating: Yes, will do. The docs say the error type can be extracted after the first colon of the error message. From some testing this pattern seems to be stable. I will add some example comments.

mds1: I see you added:
```typescript
// A couple example error messages from ens-normalize:
// ‌hi.eth - Invalid label "{200C}hi"‎: disallowed character: {200C}
// aα.eth - Invalid label "aα"‎: illegal mixture: Latin + Greek "α"‎ {3B1}
```

but:

```typescript
x = `Invalid label "{200C}hi"‎: disallowed character: {200C}`
x.split(':')[1] // <-- returns ' disallowed character'
```

Am I misinterpreting the comment?

mds1: Ohh I see, we're just showing an error of `We detected a disallowed character in the ENS name.`, but we are not showing what the disallowed character is

Wdyt about changing the error message to `The given ENS name is invalid due to: '${x}'. Check the ENS name in order to avoid a potential scam.` Where `x` would be `disallowed character: {200C}` or `illegal mixture: Latin + Greek "α"‎ {3B1}` with the above examples. This way the error is more specific and helpful

alexkeating: Yes, definitely I will add that

alexkeating: https://github.com/ScopeLift/umbra-protocol/pull/555/commits/ae44c46020baeeecd16b68bf7e06276a24e6d01c

apbendi (approved):
Looks good to me, works well. Good to merge after rebase!

PR #556, opened 2023-06-29
Sort amount correctly

### LINKED ISSUE #550: Fix Amount Based Sorting on Receive Page
It's currently unclear what value amount based sorting is using. We should change it to be something consistent and intuitive. The easiest initial fix would be to sort based on the real-value (i.e. after decimal conversion) amount, regardless of the denomination/value of the token. For example, if someone receives  28 MATIC, 0.1 WETH, 20 USDC, 18 DAI, and 27 USDT it should sort ascending as:

0.1 WETH
18 DAI
20 USDC
27 USDT
28 MATIC

alexkeating:
#### Description

- Sort amount by real value

Closes #550

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -456,6 +456,15 @@ function useReceivedFundsTable(announcements: UserAnnouncement[], spendingKeyPai
       name: 'amount',
       sortable: true,
       format: toString,
+      sort: (a: BigNumber, b: BigNumber, rowA: { token: string }, rowB: { token: string }) => {
+        const tokenA = tokens.value.find((token) => {
+          return token.address === rowA.token;
+        });
+        const tokenB = tokens.value.find((token) => {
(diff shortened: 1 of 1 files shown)


alexkeating:
FYI, I don't have an account with all the varieties of tokens in @apbendi's example, but this seems to be working with a single token. If anyone has test tokens they want to send me on Sepolia I can test with more.

mds1:
> If anyone has test tokens they want to send me on Sepolia I can test with more.

The DAI and USDC tokens we use on sepolia can be minted by anyone, so you can mint as much as you want to your wallet: https://github.com/ScopeLift/umbra-api/blob/9631aa785eae26e0a9b5df58bf95ad3b4cbde6ce/src/constants/tokens.ts#L10-L11

alexkeating:
Tested with multiple tokens and seems to be working

mds1 (approved):
code seems right to me

apbendi (approved):
Looks good, working well. Thanks!

PR #558, opened 2023-06-29
Lazy load announcements 

garyghayrat:
One of the improvements listed in https://github.com/ScopeLift/umbra-protocol/issues/266

Changes made in this PR:
- Change functions `fetchAllAnnouncements` and `fetchSomeAnnouncements` to `asyncGenerator` functions that return 1000 annoucements as each 1000 are fetched from the subgraph.
- Announcements are collected inside `announcementsQueue` and once `announcementsCount` hits 10,000, we start the initial scan by calling `filterUserAnnouncementsAsync`
-  `filterUserAnnouncementsAsync` is an async function that create web workers using the previously existing `filterUserAnnouncements` and it will only `resolve()` once all web workers are done scanning. The `filterUserAnnouncementsAsync` wrapper function exists because we want to be able to keep fetching for additional announcements beyond the first 10,000 as the workers are scanning in parallel.
- We call `filterUserAnnouncementsAsync` again to scan the rest of the announcements inside `announcementsQueue` only when the first set of web workers are done. This is because creating addition web workers when the first scan isn't yet complete will cause new workers to take away resources from the first scan web workers, slowing down the first scan process dramatically.
- While filtering announcements/ scanning for funds, if announcements for the user is found in `filterUserAnnouncements`, we increment `tableKey.value` to force refresh `account-receive-table` to update the table.

Changes in the templates are mostly in terms of reordering scanning/fetching/receive-table so that receive table is at top, and second round of fetching and scanning will be shown under the table. Additionally `tableKey` ref is added to help refresh the `account-receive-table` component when new user announcements are found.

This video with `console.log`s might be a helpful visualization:

https://github.com/ScopeLift/umbra-protocol/assets/61768337/af7f4694-a1f7-4882-87ef-e3661261f1ad


garyghayrat:
Current dependencies on/for this PR:
* master
  * **PR #558** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/558" target="_blank">Graphite</a>  👈
    * **PR #573** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/573" target="_blank">Graphite</a> 
      * **PR #574** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/574" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/558?utm_source=stack-comment).

frontend/src/pages/AccountReceive.vue
```diff
@@ -60,21 +60,21 @@
         </q-card>
```
apbendi, line 60: Hey @garyghayrat, this seems like a great start. It's definitely a better experience to only have to wait for a small, recent tranche of announcements to find recent transactions. There are some UX issues we're going to have to figure out though.

First, I'm not a fan of the way the second tranche of scanning takes place with a full UI below the table. It's a bit confusing. It's also not accurate to show the message that says "Scanning Completed Successfully" while the scanning is still ongoing.

There are lots of ways we could solve this, but here's one mockup I through together:

scanning-mockup

Another issue is what happens when the second tranche of scanning completes. As it is now, the full table completely reloads. This leads to really bad UX of the screen flashing, and it's very problematic if the user is in the middle of attempting or completing a withdraw, because the forms/modals are lost. See this video to see what I mean:

receive-reset-issues

I think the right way to do this is to *update* the table contents with additional records after they're found rather than completely reseting/re-rendering the table. I'm not sure if this is easy to do with the components we're using, but does it make sense?

garyghayrat: That makes complete sense, and love the mockup. 
> update the table contents with additional records after they're found rather than completely reseting/re-rendering the table. 
I'll look into this more and iterate, then I'll get back to you!

mds1: Given that web workers run in parallel, and we want to update the table in place to prevent jumping around or other UX issues, does this result in us losing ordering guarantees in the table? Web worker 2 might return logs 1000-2000 before web worker 1 finishes returning logs 0-1000, so I think we have to choose between "stable table ordering when events come in" and "sort table by timestamp"?

frontend/src/pages/AccountReceive.vue
```diff
@@ -178,40 +179,92 @@ function useScan() {
...
+
+    try {
+      if (!signer.value) throw new Error('signer is undefined');
+      if (!userWalletAddress.value) throw new Error('userWalletAddress is undefined');
+
+      let announcementsCount = 0; // Track the count of announcements
+      let announcementsQueue: AnnouncementDetail[] = []; // Announcements to be filtered
+      let firstScanPromise = Promise.resolve();
+      // When in advanced mode, we fetch all announcements
+      if (advancedMode.value) {
```
mds1, line 225: Why do we always fetch all announcements in advanced mode? I think we can remove this `if` block and use the same logic regardless of advanced mode

frontend/src/pages/AccountReceive.vue
```diff
@@ -178,40 +179,92 @@ function useScan() {
...
-      allAnnouncements,
-      (percent) => {
-        scanPercentage.value = Math.floor(percent);
-      },
-      (filteredAnnouncements) => {
-        userAnnouncements.value = filteredAnnouncements.sort(function (a, b) {
-          return parseInt(a.timestamp) - parseInt(b.timestamp);
-        });
+    // Wrapper for `filterUserAnnouncements` so we can await for the first batch of web workers to finish before
+    // creating new workers.
```
mds1, line 190: What do you mean by "the first batch of web workers"? I'm not too familiar with them so would appreciate more details in the comments for my own understanding / to simplify future updates here. Is the number of web workers in the first batch equal to the max number of web workers allowed, or is it 1, or do we limit it to same ratio of the available web workers so we have others available for future incoming announcements?

umbra-js/src/classes/Umbra.ts
```diff
@@ -745,8 +754,10 @@ async function recursiveGraphFetch(
     body: JSON.stringify({
       query: query(`
         first: 1000,
+        orderBy: timestamp,
+        orderDirection: desc,
         where: {
-          ${fromId ? `id_gt: "${fromId}",` : ''}
+          ${fromTimestamp ? `timestamp_lt: "${fromTimestamp}",` : ''}
```
mds1, line 760: Why did we change from using the log ID to using timestamp? Not sure if it matters, just wondering. For reference, we have the IDs in the subgraph set to `event.block.number.toString() + '-' + event.logIndex.toString()`

umbra-js/src/classes/Umbra.ts
```diff
@@ -759,8 +770,10 @@ async function recursiveGraphFetch(
 
   // If there were results on this page then query the next page, otherwise return the data
   /* eslint-disable @typescript-eslint/no-unsafe-return */
-  if (!json.data[key].length) return [...before];
-  else return await recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides);
+  if (json.data[key].length) {
+    yield json.data[key]; // yield the data for this page
+    yield* recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides); // yield the data for the next pages
+  }
```
mds1, line 776: Was the `else` block intentionally dropped here? Seems like a bug was introduced
```suggestion
  if (json.data[key].length) yield json.data[key]; // yield the data for this page
  else yield* recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides); // yield the data for the next pages
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -759,8 +770,10 @@ async function recursiveGraphFetch(
...
   /* eslint-enable @typescript-eslint/no-unsafe-return */
```
mds1, line 777: Was the `else` block intentionally dropped here? Seems like a bug may have been introduced. Also now that we use `yield`, maybe the eslint hint
```suggestion
  // If there were results on this page then query the next page, otherwise return the data
  if (json.data[key].length) yield json.data[key]; // yield the data for this page
  else yield* recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides); // yield the data for the next pages
``` is outdated?

File: frontend/src/worker/worker.ts [hunk 7]
```diff
@@ -88,9 +88,6 @@ export const filterUserAnnouncements = (
           // sorting based on `timestamp`.
           // in Umbra.ts `fetchAllAnnouncementFromLogs` turns `timestamp` from number to string,
           // and therefore `timestamp` can be cast to number here.
-          userAnnouncements.sort(function (a, b) {
-            return parseInt(a.timestamp) - parseInt(b.timestamp);
-          });
           completion(userAnnouncements);
```
mds1, line 91: Why was this deleted, and guessing the comment also should have been deleted?

frontend/src/pages/AccountReceive.vue
garyghayrat: I think to giving more context before answering those questions will be helpful here. 
So the initial ideal goal was to scan every 1000 announcements as they come in and whenever new announcements are found, update the table with the new user announcements.

Challenges there were that creating and teriminating web workers cost resources and time, and what was actually happening was, before the first batch of maximum allowed web workers are created and have the time to process the first 1000 announcements, 10-20 requests to create new sets of workers all trying to process their share of 1000 announcements were taking so much resource and time that it actually slowed down the whole process. 

I tried using a library like [workerpool](https://www.npmjs.com/package/workerpool) which was suppose to help create a pool of workers, and instead of terminating them once they done scanning (terminating and creating workers take resources), they would wait for new announcements to be added to their task queue. But after installing that library, I'm assuming because we are trying to call pieces of code from a dependency like `umbra-js` I had webpack related issues that I thought wasn't worth the time investigating. 

Ben suggested splitting up the process in 2 instead, where you first scan 10,000 announcements, which should cover a reasonable timeframe where most of our users will have received funds, and then scan rest of the announcements.

frontend/src/pages/AccountReceive.vue
mds1: Thanks for the context in https://github.com/ScopeLift/umbra-protocol/pull/558#discussion_r1267213396! Moving the discussion here to avoid de-railing the scope of that thread.

I want to make sure I fully understand the flow. Specifically, I think we have:
1. First fetch 10k announcements, this may takes multiple HTTP requests to the subgraph/log so this happens serially with no web workers
2. Pass those 10k logs to `n` web workers to start decrypting/processing the logs, where `n < maxWorkers`. Question: How is `n` chosen though?
3. While 2 is processing, repeat step 1 to fetch 10k more logs with on web workers.
4. Pass the next batch to `m` web workers to start decrypting/processing the logs, where `m <= maxWorkers - n`. Question: How is `m` chosen? And how do we know if we need to reserve more workers for another batch of 10k?
5. Repeat the above steps until we've fetched all logs and allocated all workers to decrypting
6. As a worker finishes processing it's batch of logs, immediately show them in the table. Repeat this step until all logs are processed 

So let me know if that flow/understanding is correct. There's also some questions I wrote up in the above flow.

Also, if that's correct, it feels the answer to my question is "yes, we have to choose between 'stable table ordering when events come in' and 'sort table by timestamp'", because we can't guarantee that e.g. step 2 finishes before step 4. Is that also correct?

frontend/src/pages/AccountReceive.vue
mds1: Moving this convo to https://github.com/ScopeLift/umbra-protocol/pull/558#discussion_r1267233356 to avoid derailing this thread

frontend/src/pages/AccountReceive.vue
garyghayrat: > 2. Pass those 10k logs to n web workers to start decrypting/processing the logs, where n < maxWorkers. Question: How is n chosen?

I believe `n` is maxWorkers allowed. All of these are done inside `frontend/src/worker/worker.ts` file, I believe added with this PR https://github.com/ScopeLift/umbra-protocol/pull/289

> 3. While 2 is processing, repeat step 1 to fetch 10k more logs with no web workers.

No, we actually go on to fetch rest of the announcements with no web workers(except the ones that were created in step 2), doesn't matter if there's 1000 more to fetch or 30000 more.

4. First batch of web workers has to finish stop scanning before we do anything else because creating additional workers actually will slow down the scan of first 10,000 announcements. But meanwhile we can keep fetching until we've fetch all of the rest of the announcements.

5. Web workers finish scanning 10,000 announcements.

6. New `n` webworkers are created to process the rest of the announcements (maybe 1000, maybe 30000+, doesn't matter, this is the final scan we will do).

garyghayrat: We only refresh the receive table when the first scan finishes and when the second/last scan finishes. And they are ordered by timestamp inside `filterUserAnnouncements`

umbra-js/src/classes/Umbra.ts
garyghayrat: > subgraph set to event.block.number.toString() + '-' + event.logIndex.toString()

Does this mean the `formId`s are already ordered by their block number? if that's the case we can revert this change. The idea was just to start fetching from the most recent announcements to the oldest last.

frontend/src/pages/AccountReceive.vue
mds1: Ok getting closer! Updated understanding:

1. First fetch 10k announcements, this may takes multiple HTTP requests to the subgraph/node so this happens serially with no web workers.
2. Pass those 10k logs to `maxWorkers` web workers to start decrypting/processing the logs.
3. While 2 is processing, repeat step 1 to fetch *all* remaining announcement logs.
4. Wait:
    1. For step (2) to complete, and show any results in the table. At this point, all web workers have no jobs.
    2. For step (3) to complete. At this point, the app has all announcement logs.
5. Give all remaining logs to `maxWorkers` to process. Wait for them and show the rest in the table. 

Questions:
1. Are the web workers from (2) shut down then re-spawned in (5), or do they stay active but idle until step (5)?
2. Does step 5 update the table as workers finish processing, or does it wait for *all* remaining logs to be processed then update the table?

umbra-js/src/classes/Umbra.ts
mds1: Correct, the ordering by ID should be equivalent to ordering by timestamp. Note that ordering by timestamp is _not_ unique because with batch sends you can have multiple logs in the same timestamp. Whereas ordering by this ID is deterministic because it uses block number and tx log index

frontend/src/pages/AccountReceive.vue
garyghayrat: That's correct!

> 1. Are the web workers from (2) shut down then re-spawned in (5), or do they stay active but idle until step (5)?

I haven't changed `worker.ts` behavior, so they shut down and then re-spawn in (5).

> 2. Does step 5 update the table as workers finish processing, or does it wait for all remaining logs to be processed then update the table?

Table is updated only when all remaining logs are processed and done.

umbra-js/src/classes/Umbra.ts
garyghayrat: Oh true, that's good to know. Thanks!

frontend/src/pages/AccountReceive.vue
garyghayrat: Lmk if you know an easy or obvious way of doing this
> update the table contents with additional records after they're found rather than completely reseting/re-rendering the table.

A short search didn't reveal anything straightforward

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
garyghayrat: Rough idea from my testing during development

|         | First shown | All shown |
|--------------|:-----:|----:|
| master |     | Faster |
| This PR     |  Faster* |           |

*if the user has recent transactions within the latest 10k announcements.
The reason for `master` branch currently being faster to show full history is again because starting/terminating web workers and allocating work to them takes time.

frontend/src/pages/AccountReceive.vue
```diff
@@ -98,6 +98,7 @@ function useScan() {
   const scanStatus = ref<ScanStatus>('waiting');
   const scanPercentage = ref<number>(0);
   const userAnnouncements = ref<UserAnnouncement[]>([]);
+  const tableKey = ref(0);
```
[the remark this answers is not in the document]
garyghayrat: This is the key attached to the `account-receive-table`, when it changes it'll force a refresh on the component with the latest `userAnnouncements`.

frontend/src/pages/AccountReceive.vue
garyghayrat: Latest behavior is if `advancedMode` is on and `scanPrivateKey` is provided, we fetch for all announcements.

frontend/src/worker/worker.ts
garyghayrat: Cause I think there's duplicate work, we're also sorting in `AccountReceive.vue` https://github.com/ScopeLift/umbra-protocol/blob/3b2b98d895ba7ac4365f7813a8e0b59394682391/frontend/src/pages/AccountReceive.vue#L236C11-L239C16

umbra-js/src/classes/Umbra.ts
garyghayrat: So I played around and the suggestion code is yielding the first page first result and immediately stopping.
I think the block of code
```
  if (json.data[key].length) {
    yield json.data[key]; // yield the data for this page
    yield* recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides); // yield the data for the next pages
  }
  ```
  is correct as it would return the latest call results and then call recursively from where it left off. And if it finds no data, then do nothing because all previous data has already been returned.

Comment (garyghayrat) [reply umbra-js/src/classes/Umbra.ts]: Updated comments: https://github.com/ScopeLift/umbra-protocol/pull/558/commits/619eff1c918bd837a11b4a21fe6a5c13aeed19fa

File: frontend/src/pages/AccountReceive.vue (hunk 3 above)
Comment (garyghayrat) [reply frontend/src/pages/AccountReceive.vue:222]: Now using [watchEffect](https://vuejs.org/guide/essentials/watchers.html#watcheffect) to accomplish this

File: frontend/src/pages/AccountReceive.vue (hunk 2 above)
Comment (mds1) [reply frontend/src/pages/AccountReceive.vue]: Why do we not respect user-provided block ranges if a private key is provided in advanced mode?

File: frontend/src/pages/AccountReceive.vue (hunk 3 above)
Comment (mds1) [reply frontend/src/pages/AccountReceive.vue:222]: > That's correct!

Great, let's make sure this flow is documented somewhere, a big comment block in whatever file seems best is ok (maybe `worker.ts`?)

Comment (mds1) [reply frontend/src/pages/AccountReceive.vue:222]: > A short search didn't reveal anything straightforward

Yea I don't know of anything offhand, a watcher seems ok if that works. 

Thanks for all the back and forth here!

File: umbra-js/src/classes/Umbra.ts (hunk 4 above)
Comment (mds1) [reply umbra-js/src/classes/Umbra.ts:760]: Yea so we should probably revert this back to using IDs, since I don't think timestamp will guarantee unique ordering (it might, but no reason to rely on the ambiguity there)

Comment (mds1) [reply umbra-js/src/classes/Umbra.ts:760]: For reference, you can find our subgraph / ID definition here: https://github.com/ScopeLift/umbra-subgraph/blob/a390b27fa6a6431875402e07b3ae302e87c59ec3/src/mapping.ts#L7

File: umbra-js/src/classes/Umbra.ts (hunk 6 above)
Comment (mds1) [reply umbra-js/src/classes/Umbra.ts]: > So I played around and the suggestion code is yielding the first page result and immediately stopping.

Oh maybe because there was a typo in my suggestion? You can see the original code was:
```typescript
if (!json.data[key].length) return [...before];
else return recursiveGraphFetch(...);
```

But this PR flips the first condition to remove the `!`. I'm mainly just surprised that changing from `return` to `yield` results in a differing conditional here which is why I'm making sure I understand what's going on.

In the old approach, nothing was returned until we hit the `if (!...)` condition, but in the new approach return data is basically streamed since this is a generator, which is why we don't need that anymore?

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
mds1: Sorry, can you provide the actual numbers instead of just "faster"? Along with how many runs you did for each box 😅

The first row should have the same time in both columns, and the second row should have two different values. But I want to better understand the tradeoff we're going with and how significant the differences are (faster to first shown, slower to all shown)

> *if the user has recent transactions within the latest 10k announcements.

So presumably we'd be interested in two versions of this table, one where they do have txs in the last 10k announcements, and one where they don't?

I'm just trying to be cautious and make sure we're achieving the desired goals without introducing any significant tradeoffs

apbendi (approved):
I tested this extensively and it seems to be working great. This will be a big UX improvement for our users. Nice work @garyghayrat.

frontend/src/pages/AccountReceive.vue
garyghayrat: Actually we do. We pass in `overrides` to `fetchAllAnnouncements`, which will include user specified block ranges.

frontend/src/pages/AccountReceive.vue
garyghayrat: Thank you @alexkeating !

umbra-js/src/classes/Umbra.ts
garyghayrat: > return data is basically streamed since this is a generator, which is why we don't need that anymore?

I believe this is the reason. It's pretty much like


```
  if (!json.data[key].length) {
    return;
  }
  yield [...before, ...json.data[key]];
  yield* recursiveGraphFetch(url, key, query, [...before, ...json.data[key]], overrides);
  ```

File: umbra-js/src/classes/Umbra.ts (hunk 4 above)
Comment (garyghayrat) [reply umbra-js/src/classes/Umbra.ts:760]: So an additional change that was added was using `orderBy: timestamp` and `orderDirection: desc` whitch helped fetch the latest 1000 announcements. 
Seems like using `formId` as a filter to query makes it fetch the oldest 1k announcements. And me trying to use `orderBy : formId` is giving the following error Screenshot 2023-07-26 at 4 37 57 PM
Maybe we should break this into an issue because currently ordering by timestamps is working batchSend transactions.

Comment (garyghayrat) [reply umbra-js/src/classes/Umbra.ts:760]: Issue https://github.com/ScopeLift/umbra-protocol/issues/578

File: umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
Comment (garyghayrat) [reply umbra-js/src/classes/Umbra.ts:1]: Sorry meant to come back to this with more specific testing. The following is still not rigorous enough, but the each row is tested at least twice. It seems like at least from my tests there isn't a penalty? I'll test later tonight on my old laptop.

More than 10k annoucements(Arbitrum)
|         | First shown | All shown |
|--------------|:-----:|----:|
| master |   1:04  | 1:04 |
| This PR     |  0:19 |    1:02       |


Less than 10k annoucements(Polygon)
|         | First shown | All shown |
|--------------|:-----:|----:|
| master |   0:11 | 0:11  |
| This PR     |  0:11 |   0:11  |

File: frontend/src/pages/AccountReceive.vue [hunk 9]
```diff
@@ -256,6 +338,7 @@ function useScan() {
     startBlockLocal,
     userAddress,
     userAnnouncements,
+    tableKey,
```
alexkeating, line 341: Can this be removed now?

alexkeating (commented):
One comment seems to be working for me. Great job!

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
mds1: Oh nice this is great! For polygon, to get less than 10k announcements, did you just change the start block to something recent?

umbra-js/src/classes/Umbra.ts
mds1: Yep make sense, thanks a lot!

frontend/src/pages/AccountReceive.vue
mds1: Ok, I see the overrides being passed. So the comment in https://github.com/ScopeLift/umbra-protocol/pull/558/files#diff-27047766a3ad711f41f2a80bc8036e498999c7963b3196e7b9e4f27ae3e5804eR256-R258 needs to be updated then?

umbra-js/src/classes/Umbra.ts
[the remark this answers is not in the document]
garyghayrat: Oh I just used an account that I registered recently and knew from testing that it would only fetch <10k announcements.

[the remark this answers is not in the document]
garyghayrat: 4 Core CPU. Uuuuf, those new contracts can't come fast enough.

More than 10k annoucements(Arbitrum)
|   min:sec      | First shown | All shown |
|--------------|:-----:|----:|
| master |   4:13  | 4:13 |
| This PR     |  0:54 |    4:11       |


Less than 10k annoucements(Polygon)
|     min:sec    | First shown  | All shown |
|--------------|:-----:|----:|
| master |   0:31 | 0:31  |
| This PR     |  0:31 |   0:31  |

frontend/src/pages/AccountReceive.vue
garyghayrat: Yes, removed here: https://github.com/ScopeLift/umbra-protocol/pull/558/commits/599d33aead443ef144250df11c5d6f8714004b48

frontend/src/pages/AccountReceive.vue
garyghayrat: Updated here : https://github.com/ScopeLift/umbra-protocol/pull/558/commits/599d33aead443ef144250df11c5d6f8714004b48

alexkeating (approved):

PR #568, opened 2023-07-13
Fix Gnosis scanning error -- Block range too wide

garyghayrat:
Fix for receive page scanning on Gnosis Chain, which currently throws `Internal JSON-RPC error`

Screenshot 2023-07-13 at 3 49 08 PM

Regression: 
Filtering announcements no longer returns an empty array when the startBlock is `undefined` instead it reverts back to `chainConfig` default `startblock`. Will create an issue to handle this elsewhere using `isAccountSetup` from `walletStore`. See https://github.com/ScopeLift/umbra-protocol/issues/569


apbendi (approved):
Works! Scanning on Gnosis Chain seems to be working again. Thanks @garyghayrat

alexkeating (approved):

PR #571, opened 2023-07-13
On receive page, ask user to setup account if unregistered

### LINKED ISSUE #569: Use isAccountSetup from walletStore to prevent unnecessary fetching and scanning
Fix the regression caused by https://github.com/ScopeLift/umbra-protocol/pull/568 . 
After merging the PR, filtering announcements will no longer return an empty `userAnnouncements` array when the `startBlock` is undefined(i.e. user registered block couldn't be found, undefined). Instead it reverts back to `chainConfig` default `startblock`.

Idea to fix: check `isAccountSetup` on the receive page before requiring user signature and a scan, and prompt the user to set up their account if they are not in advanced mode.

garyghayrat:
Closes #569 

Before:

https://github.com/ScopeLift/umbra-protocol/assets/61768337/81ead8bf-39c3-490c-94ab-2fde789a0bc8

After: 

https://github.com/ScopeLift/umbra-protocol/assets/61768337/38d2f130-528b-45ab-bb42-39ddfeb6dea8

### DIFF
--- a/frontend/src/css/app.sass
+++ b/frontend/src/css/app.sass
@@ -293,7 +293,7 @@ p, div
   border: 1px solid $warning
   border-radius: 10px
 
-.border-radius
+.rounded-borders-md
   border-radius: 15px
 
 .text-break-word
--- a/frontend/src/css/app.sass
+++ b/frontend/src/css/app.sass
@@ -196,8 +196,14 @@ p, div
   .bg-muted
     background: $lightestgrey
 
+  .bg-notice
+    background: #FCEFC7
+
   .text-muted
     color: $lightgrey
+
+  .text-notice
+    color:#513C06
 
   .hyperlink
     color: $primarydark
@@ -234,8 +240,14 @@ p, div
   .bg-muted
     background: $darkestgrey
 
-  .text-muted
+  .bg-notice
+    background: #7C5E10
+
+    .text-muted
     color: $darkgrey
+
+  .text-notice
+    color:#FFFAEB
 
   .hyperlink
     color: $primaryreg
@@ -283,6 +295,9 @@ p, div
 .border-warning
   border: 1px solid $warning
   border-radius: 10px
+
+.border-radius
+  border-radius: 15px
 
 .text-break-word
   word-wrap: break-word
--- a/frontend/src/css/app.sass
+++ b/frontend/src/css/app.sass
@@ -76,8 +76,6 @@
 .danger
   color: $danger
 
-
-
 h1, h2, h3, h4, h5, h6
   margin-bottom: 0
 h4, h5
@@ -127,7 +125,6 @@ p, div
 
 .input-container
   margin: 0px 10px
-
 
 .input-container-token
   min-width: 140px
@@ -203,7 +200,7 @@ p, div
     color: $lightgrey
 
   .text-notice
-    color:#513C06
+    color: #513C06
 
   .hyperlink
     color: $primarydark
@@ -243,11 +240,11 @@ p, div
   .bg-notice
     background: #7C5E10
 
-    .text-muted
+  .text-muted
     color: $darkgrey
 
   .text-notice
-    color:#FFFAEB
+    color: #FFFAEB
 
   .hyperlink
     color: $primaryreg
@@ -301,4 +298,3 @@ p, div
 
 .text-break-word
   word-wrap: break-word
-
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -11,7 +11,19 @@
       </div>
     </div>
 
-    <div v-else class="q-mx-auto" style="max-width: 800px">
+    <div
+      v-else-if="!isAccountSetup && !advancedMode"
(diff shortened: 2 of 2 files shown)


frontend/src/pages/AccountReceive.vue
```diff
@@ -11,7 +11,19 @@
       </div>
     </div>
 
-    <div v-else class="q-mx-auto" style="max-width: 800px">
+    <div
+      v-else-if="!isAccountSetup && !advancedMode"
+      class="dark-toggle text-center text-bold q-pa-md q-mb-lg"
+      style="border-radius: 15px"
+      :style="isDark ? 'color: #FFFAEB; background-color: #7C5E10' : 'color: #513C06; background-color: #FCEFC7'"
```
alexkeating, line 18: It is better to use one of the quasar shading classes or the sass variables we have setup in the app. It allows us to keep the colors consistent in the app.

alexkeating (commented):
Working for me one small comment

frontend/src/pages/AccountReceive.vue
```diff
@@ -11,7 +11,19 @@
       </div>
     </div>
 
-    <div v-else class="q-mx-auto" style="max-width: 800px">
+    <div
+      v-else-if="!isAccountSetup && !advancedMode"
+      class="dark-toggle text-center text-bold q-pa-md q-mb-lg"
```
mds1, line 16: IMO we should keep the max-width, it looks odd to have the message take the full width of the page

image

frontend/src/pages/AccountReceive.vue
mds1: Yea try to avoid `style` as much as possible for the reason @alexkeating mentioned. I believe quasar has classes for border radius, maybe max width also

garyghayrat: Good to know, thank you! Addressed it in https://github.com/ScopeLift/umbra-protocol/pull/571/commits/d34d75c3e854848b58237035fb4eea4868304fdc
- Quasar Palette didn't really have colors that were close so created custom text and bg colors in sass file
- Quasar rounded border class wasn't round enough so also created a custom class in sass
- Quasar max-width option didn't match our widths, so just reused an existing sass class.

frontend/src/pages/AccountReceive.vue
garyghayrat: Addressed it in https://github.com/ScopeLift/umbra-protocol/commit/d34d75c3e854848b58237035fb4eea4868304fdc

frontend/src/pages/AccountReceive.vue
garyghayrat: Ran prettier on .sass https://github.com/ScopeLift/umbra-protocol/pull/571/commits/cdbf1a522403bc910a36c40d0bf0b750e7e392ea

mds1: Thanks! Since you pulled things out into classes, should we also pull out existing hardcoded classes, e.g. the `:style="isDark ? 'color: #FFFAEB; background-color: #7C5E10' : 'color: #513C06; background-color: #FCEFC7'"` line in this PR? Or a follow up PR?

garyghayrat: There seems to be a lot more of those. Let me create an issue for it and fix them in a separate PR

garyghayrat: Created an issue https://github.com/ScopeLift/umbra-protocol/issues/572

alexkeating (approved):

frontend/src/css/app.sass
```diff
@@ -76,8 +76,6 @@
 .danger
```
apbendi, line 76: *arbitrary comment location*

Hey @garyghayrat, overall this is looking good, however I'm observing the following behavior with regards to advanced mode: if the user has never registered stealth keys before, but turns on advanced mode, it now fetches and scans *all announcements*. Why is this the case? We should only need to scan all blocks if the user supplies a private key (because now we don't know when that was registered) or if the user specifies a block range (in which case we should respect it). Am I missing something?

apbendi (commented):
Looking good @garyghayrat, but a question about advanced mode behavior.

frontend/src/css/app.sass
garyghayrat: Updated here: https://github.com/ScopeLift/umbra-protocol/pull/571/commits/ac938e6ec365de35ee3c76035eb169357c5a601a

apbendi: Ok @garyghayrat, now I'm seeing the opposite problem:

* I connect with an account that *has* completed setup, and *does* have past transactions in its history
* I turn on advanced mode and scan
* It completes immediately and it shows no transactions
* I turn advanced mode off, refresh and scan
* It scans much longer and shows the past transactions

See the video below:

umbra-advanced-no-scan

garyghayrat: Ok, this should fix it: https://github.com/ScopeLift/umbra-protocol/pull/571/commits/dd8e70078af457c73b67a3589e92e25df5510809

apbendi (approved):
Seems to be working well @garyghayrat, thanks!

PR #573, opened 2023-07-17
Update fetching/scanning status w/ i18n

garyghayrat:

PR #574, opened 2023-07-17
Fix #558 CI

garyghayrat:
Fixes [CI failures](https://github.com/ScopeLift/umbra-protocol/actions/runs/5556669269/jobs/10149474025?pr=558) in #573 #558 
**Issue**: Because `typeof window` is undefined during the test, the `try` `catch` block of code is skipped, resulting in an empty `userAnnouncements` array being returned. Therefore, all tests involving the `Umbra.scan` function was failing the CI.

**Solution**: If conditions for the `try` `catch` block aren't met, yield announcements from logs:
```
397    yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
```
Because `fetchAllAnnouncements` is now a AsyncGenerator function that `yields` announcements as they are fetched, also update the previous call to `fetchAllAnnouncementFromLogs` to
```
394    yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
```

Also update variable name from ambiguous `ann` to `announcementsBatch`

### DIFF
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -393,8 +393,9 @@ export class Umbra {
       } catch (err) {
         yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
       }
+    } else {
+      yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
     }
-    yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
   }
 
   /**
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -391,9 +391,10 @@ export class Umbra {
           yield details;
         }
       } catch (err) {
-        return this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
+        yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
       }
     }
+    yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
   }
 
   /**
@@ -517,8 +518,8 @@ export class Umbra {
    */
   async scan(spendingPublicKey: string, viewingPrivateKey: string, overrides: ScanOverrides = {}) {
     const userAnnouncements: UserAnnouncement[] = [];
-    for await (const ann of this.fetchAllAnnouncements(overrides)) {
-      for (const announcement of ann) {
+    for await (const announcementsBatch of this.fetchAllAnnouncements(overrides)) {
+      for (const announcement of announcementsBatch) {
         const { amount, from, receiver, timestamp, token: tokenAddr, txHash } = announcement;
         const { isForUser, randomNumber } = Umbra.isAnnouncementForUser(
           spendingPublicKey,


garyghayrat:
Current dependencies on/for this PR:
* master
  * **PR #558** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/558" target="_blank">Graphite</a> 
    * **PR #573** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/573" target="_blank">Graphite</a> 
      * **PR #574** <a href="https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/574" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ScopeLift/umbra-protocol/574?utm_source=stack-comment).

umbra-js/src/classes/Umbra.ts
```diff
@@ -391,9 +391,10 @@ export class Umbra {
           yield details;
         }
       } catch (err) {
-        return this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
+        yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
       }
     }
+    yield await this.fetchAllAnnouncementFromLogs(startBlock, endBlock);
```
mds1, line 397: In the PR description can you explain the issue, and how this fixes it? Linking to a failed CI run demonstrating the issue would also be great

garyghayrat: My bad! Updated the description.

mds1: Got it. So node.js 18 [now](https://nodejs.org/es/blog/announcements/v18-release-announce) has the fetch API. So an alternative solution is bump the minimum required node version from v16 to v18 and make no other code changes

garyghayrat: I can update node.js version too, but isn't another reason why we check for `window` is so when running `yarn test` during CI, it doesn't actually try to fetch announcements from the subgraph. Instead it just fetches them from the log?

garyghayrat: I also think the code change in this PR wouldn't be any different? Because note that this PR's main aim is to fix the code in another PR https://github.com/ScopeLift/umbra-protocol/pull/558

mds1: IIUC you're saying in CI it's preferable to fetch logs directly from the node instead of from a subgraph? If so I don't think, why is that? Note that we also don't support getting logs from the node on certain chains because of the issues with log fetching

mds1: > I also think the code change in this PR wouldn't be any different? Because note that this PR's main aim is to fix the code in another PR https://github.com/ScopeLift/umbra-protocol/pull/558

Ah sorry I haven't reviewed that PR yet so I'm not fully caught up on that. But based on the PR description of "Issue: Because typeof window is undefined" and solution of "yield instead of return" it sounds like we should just update our node version and remove the `window` check instead

garyghayrat: No, that's my bad, I was just referring to the lines of code and def should have given more context.

> IIUC you're saying in CI it's preferable to fetch logs directly from the node instead of from a subgraph? If so I don't think, why is that? Note that we also don't support getting logs from the node on certain chains because of the issues with log fetching

In terms of this, when we're testing in `umbra-js` we send mock transactions on hardhat chain, so we can only query them from the logs right?

mds1: > In terms of this, when we're testing in `umbra-js` we send mock transactions on hardhat chain, so we can only query them from the logs right?

We shouldn't tailor library behavior to its test suite—umbra-js should work just as well in node environments as browser envs, which ultimately means upgrading node to remove the existing limitation. If a subgraph call fails we should fallback to trying logs (i.e. in tests). But this isn't really a priority now so if the solution I'm suggesting doesn't make sense (since I'm not yet caught up on #558) we can track this in a separate issue

garyghayrat: As Ben said here: https://github.com/ScopeLift/umbra-protocol/pull/573#issuecomment-1640724964 I think me using graphite inappropriately might have caused some unnecessary confusion. I'll merge in the sub branches into https://github.com/ScopeLift/umbra-protocol/pull/558 and let's continue the conversation there.

garyghayrat: > We shouldn't tailor library behavior to its test suite—umbra-js should work just as well in node environments as browser envs, which ultimately means upgrading node to remove the existing limitation. If a subgraph call fails we should fallback to trying logs (i.e. in tests). 

This makes sense. Opened this issue https://github.com/ScopeLift/umbra-protocol/issues/575, I think I can open a quick PR to fix this issue so we don't have to wait for  https://github.com/ScopeLift/umbra-protocol/pull/558 to be merged in

Issue #575, opened 2023-07-18
Update pinned node.js version to have access to the `fetch` method


garyghayrat:
- [x] Update node version https://github.com/ScopeLift/umbra-protocol/pull/577
- [ ] Remove `typeof window !== 'undefined'` in `umbra-js`

              Got it. So node.js 18 [now](https://nodejs.org/es/blog/announcements/v18-release-announce) has the fetch API. So an alternative solution is bump the minimum required node version from v16 to v18 and make no other code changes, because the reason we check for existence of the window object is because we need the fetch method: https://github.com/ScopeLift/umbra-protocol/blob/9c0f11e13cb4e0684db6f421d400c3093aacd253/umbra-js/src/classes/Umbra.ts#L383-L385

_Originally posted by @mds1 in https://github.com/ScopeLift/umbra-protocol/pull/574#discussion_r1267128299_

PR #577, opened 2023-07-24
Update to node@18.17.0

garyghayrat:


contracts-core/package.json
```diff
@@ -10,6 +10,7 @@
     "@commitlint/config-conventional": "^9.1.2",
     "@ethersproject/abstract-signer": "^5.5.0",
     "@ethersproject/bignumber": "^5.5.0",
+    "@ethersproject/bytes": "5.0.0",
```
mds1, line 13: What's this for? Seems surprising we'd need to add a dep in `contracts-core`

garyghayrat: It's a new warning that popped up during the `yarn install` process after upgrading to new node version. Thought this might be causing the failure. 
https://github.com/ScopeLift/umbra-protocol/actions/runs/5649918220/job/15306143560
```
warning "workspace-aggregator-d28f0833-fbcf-481a-914d-96f0486c469d > @umbra/contracts-core > @typechain/ethers-v5@10.2.0" has unmet peer dependency "@ethersproject/bytes@^5.0.0".
```

But note that even thought this PR is now passing, the CI has become very finicky where sometimes the `typechain` in `contracts-core` doesn't get generated before/when `umbra-js` tests are being run. Ben is saying it could be a race condition issue. Might be helpful to have you or @wildmolasses take a closer look.

wildmolasses: if `contracts-core` is specified as a dependency in `umbra-js`, then it will be built before `umbra-js` tests run. if it's not specified as a dependency, then lerna / yarn doesn't know that the two commands (building in contracts-core, tests in `umbra-js`) can't run parallel, or that there should be some order to them. I recommend adding contracts-core as a dependency. That then creates an issue for the standalone `umbra-js` package -- I don't think we've published `contracts-core` to our @umbracash namespace, so the package would be unavailable. Perhaps adding `contracts-core` as a devDependency, then, would be appropriate. Let me know if you want to sync on this in chat!

garyghayrat: Removed the `@ethersproject/bytes` package in https://github.com/ScopeLift/umbra-protocol/pull/577/commits/440e0deb48a49b4aab735fedeb007ef5ca19d915

garyghayrat: And solved the underlying failure by updating `hardhat` version in https://github.com/ScopeLift/umbra-protocol/pull/577/commits/e9b5b9b4369d1597e7c2916e6f7e792d3e460be4

contracts-core/package.json
```diff
@@ -32,9 +33,8 @@
     "ethereum-waffle": "^3.2.0",
     "ethers": "5.7.2",
     "fs-extra": "^9.0.1",
-    "hardhat": "^2.6.8",
+    "hardhat": "^2.17.1",
```
mds1, line 36: The PR description says:

> We're using hardhat version `2.17.0` instead of `2.17.1` to deal possibly monorepo-related error:

But this contradicts that. Curioius which is correct / what the possible issue was?

umbra-js/package.json
```diff
@@ -40,7 +40,7 @@
     "eslint-config-prettier": "^8.2.0",
     "eslint-plugin-chai-friendly": "^0.6.0",
     "eslint-plugin-import": "^2.20.2",
-    "hardhat": "^2.12.4",
+    "hardhat": "2.17.0",
```
mds1, line 43: oh I see, maybe only umbra-js was affected by the bug?

umbra-js/src/classes/Umbra.ts
```diff
@@ -149,7 +149,7 @@ export class Umbra {
    */
   constructor(readonly provider: EthersProvider, chainConfig: ChainConfig | number) {
     this.chainConfig = parseChainConfig(chainConfig);
-    this.umbraContract = new Contract(this.chainConfig.umbraAddress, UMBRA_ABI, provider) as UmbraContract;
+    this.umbraContract = new Contract(this.chainConfig.umbraAddress, UMBRA_ABI, provider) as unknown as UmbraContract;
```
alexkeating, line 152: I think we can remove these unknowns. Instead use the UmbraFactory from typechain to get a new contract. It should look like the below

```
this.umbraContract = UmbraFactory.connect(this.chainConfig.umbraAddress, provider);
```

umbra-js/src/classes/Umbra.ts
```diff
@@ -819,7 +819,7 @@ export async function assertSufficientBalance(signer: JsonRpcSigner | Wallet, to
   // If applicable, check that sender has sufficient token balance. ETH balance is checked on send. The isEth
   // method also serves to validate the token input
   if (!isEth(token)) {
-    const tokenContract = new Contract(token, ERC20_ABI, signer) as ERC20;
+    const tokenContract = new Contract(token, ERC20_ABI, signer) as unknown as ERC20;
```
alexkeating, line 822: Like the above use the factory class.

alexkeating (commented):
I think we can remove these unknowns

contracts-core/package.json
garyghayrat: Sorry, so when both `contracts-core` and `umbra-js` hardhat versions are set to `^2.17.1`, we get the following error only in `Umbra-js` when running `yarn test`
```
yarn test
yarn run v1.22.18
$ yarn build && yarn hardhat test
$ tsc --build
$ /Users/user/github/umbra-protocol/node_modules/.bin/hardhat test
Error HH12: Trying to use a non-local installation of Hardhat, which is not supported.
Please install Hardhat locally using npm or Yarn, and try again.

For more info go to https://hardhat.org/HH12 or run Hardhat with --show-stack-traces
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
So one way of solving this was installing a different versions of hardhat in umbra-js, so it would keep its own copy of hardhat.

umbra-js/package.json
garyghayrat: Yes, only having issues inside `umbra-js,` see comment above https://github.com/ScopeLift/umbra-protocol/pull/577#discussion_r1284556417

contracts-core/package.json
garyghayrat: I see that using pinned version `2.17.1` inside `umbra-js` instead `^2.17.1` also works. So I've done that here: https://github.com/ScopeLift/umbra-protocol/pull/577/commits/0a2b406b55ad9dceaa3ca55069e287756b0351c4

wildmolasses: @garyghayrat i'll be surprised if it works now; perhaps it works on your machine because you already had an umbra-js/node_modules/hardhat around, but I think we can expect that if umbra-js and contracts-core share the same version, we'll see the "non-local hardhat" error. hoping i get proved wrong!

umbra-js/src/classes/Umbra.ts
garyghayrat: Like this: https://github.com/ScopeLift/umbra-protocol/pull/577/commits/92de58d332ad2e4814a9569b068bdca6524da124 ?

umbra-js/src/classes/Umbra.ts
garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/577/commits/92de58d332ad2e4814a9569b068bdca6524da124

contracts-core/package.json
garyghayrat: ¯\_(ツ)_/¯

umbra-js/src/classes/Umbra.ts
alexkeating: Yes

umbra-js/src/classes/Umbra.ts
alexkeating: Yes

umbra-js/src/classes/Umbra.ts
garyghayrat: Thxxx!

alexkeating (approved):

mds1 (approved):
This PR LGTM. One question: I see you linked #575 here. Now that we're on nodejs 18 which has `fetch`, is the plan to make the corresponding umbra-js updates in a separate PR, or was that intended for this PR? I think there might only be one `typeof window !== 'undefined'` check that needs to be removed

garyghayrat:
> This PR LGTM. One question: I see you linked #575 here. Now that we're on nodejs 18 which has `fetch`, is the plan to make the corresponding umbra-js updates in a separate PR, or was that intended for this PR? I think there might only be one `typeof window !== 'undefined'` check that needs to be removed

I'll unlink the issue and remove that line in a separate PR! Wanna do some testing to make sure all is well after the changes.

PR #583, opened 2023-08-07
Use try catch block for address lookups Closes #582

### LINKED ISSUE #582: bug: receive table breaks if mainnet RPC is not working
My local node was down which I how I discovered this, but you can probably reproduce by building in dev mode and just ensuring the mainnet RPC URL doesn't work

I was on sepolia, and could not withdraw because the table got stuck in a hanging state like this. You can see two loading spinners in each row that persisted (next to the sender address, and in the withdraw button column)

image

garyghayrat:
Without a working mainnet RPC url `lookupOrReturnAddresses` was throwing an error causing the table to be stuck at a loading state with spinning icons. This PR adds a `try-catch` block to handle the error.

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -100,7 +100,7 @@
                         ? formatNameOrAddress(props.row.formattedFrom)
                         : formatNameOrAddress(props.row.from)
                     }}</span>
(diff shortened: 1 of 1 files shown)


mds1 (approved):
LGTM, also tested and seems to be working as expected!

alexkeating (approved):

PR #586, opened 2023-08-08
Fix walletconnect state

alexkeating:
#### Description

- Upgrading the libraries improved the UX


frontend/src/store/wallet.ts
```diff
@@ -216,7 +216,7 @@ export default function useWalletStore() {
       let connectedWallet;
       if (lastWallet.value) {
         [connectedWallet] = await onboard.value.connectWallet({
```
alexkeating, line 218: (arbitrary comment)

The logo is now messed up. Do we have a svg version of the logo.

PR #588, opened 2023-08-09
Upgrade wallet libs

### LINKED ISSUE #585: After Disconnecting Wallet Connect Show Main Connection Modal
If you connect with WalletConnect, then disconnect, then click connect again, it will show you the wallet connect screen when it should show you:

Screen Shot 2023-08-07 at 5 33 57 PM

alexkeating:
#### Description

- Upgrading the libraries improved the UX

Closes #585


umbra-js/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "composite": true,
     "declarationMap": true,
+    "skipLibCheck": true,
     "outDir": "build",
     "resolveJsonModule": true,
```
alexkeating, line 8: (arbitrary comment)

The logo is now messed up. Do we have a svg version of the logo.

alexkeating: It seems like the width is set to 56px. If we had an svg or a 56x56 image of the logo I think the issue would be resolved.

mds1: Which logo, the Umbra logo? If so I don't think we have one, I made the logo in powerpoint a few years ago and don't think I even have the original files anymore 😅 Might be able to import it into some image editing software and have it generate an SVG from it, but not sure

umbra-js/tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "baseUrl": "./",
     "paths": {
       "@sinonjs/fake-timers": ["../node_modules/@types/sinonjs__fake-timers"]
-    }
+    },
+    "typeRoots": ["./node_modules/@types"]
```
alexkeating, line 19: Got the idea from here 

https://stackoverflow.com/a/76830746

frontend/src/store/wallet.ts
```diff
@@ -132,9 +135,6 @@ export default function useWalletStore() {
             window.location.reload();
           });
           wallet.provider.on('chainChanged', () => {
-            if (wallet.label === 'WalletConnect' && !lastWallet.value) {
-              setLastWallet('WalletConnect');
-            }
```
mds1, line 137: This was modified recently in https://github.com/ScopeLift/umbra-protocol/pull/559#discussion_r1253300467, why was it needed then but not now?

alexkeating: When logging into wallet connect the page would reload before the last wallet was set which created a state where the user seemed to be logged out after the page reloaded. I tested and I wasn't able to create the issue the required this after upgrading the packages.

.github/workflows/ci.yaml
```diff
@@ -31,13 +31,16 @@ jobs:
           submodules: recursive
```
apbendi, line 31: *Arbitrary comment location*

Testing with Rainbow on mobile and wallet connect on desktop, I get a "Connection Failed" and "Connection Rejected" message on each immediately after scanning the QR code. This does not happen in Prod, where I'm able to connect without issue.

alexkeating: It works with Metamask's wallet, and Rainbow works for me when I am running on my local computer, but I get the same error when I am using the deploy preview.

PR #591, opened 2023-08-10
Remove check for window in Umbra-js

### LINKED ISSUE #575: Update pinned node.js version to have access to the `fetch` method

garyghayrat:
closes #575

### DIFF
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -381,9 +381,7 @@ export class Umbra {
     const endBlock = overrides.endBlock || 'latest';
 
     // Try querying events using Goldsky, fallback to querying logs.
-    // Goldsky fetching uses the browser's `fetch` method to query the subgraph, so we check
-    // that window is defined first to avoid trying to use fetch in node environments
-    if (typeof window !== 'undefined' && this.chainConfig.subgraphUrl) {
+    if (this.chainConfig.subgraphUrl) {
       try {
         for await (const subgraphAnnouncements of this.fetchAllAnnouncementsFromSubgraph(startBlock, endBlock)) {
           // Map the subgraph amount field from string to BigNumber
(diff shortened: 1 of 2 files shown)


umbra-js/src/classes/Umbra.ts
```diff
@@ -381,9 +381,7 @@ export class Umbra {
     const endBlock = overrides.endBlock || 'latest';
 
     // Try querying events using Goldsky, fallback to querying logs.
-    // Goldsky fetching uses the browser's `fetch` method to query the subgraph, so we check
-    // that window is defined first to avoid trying to use fetch in node environments
-    if (typeof window !== 'undefined' && this.chainConfig.subgraphUrl) {
+    if (this.chainConfig.subgraphUrl) {
```
mds1, line 384: Looks like CI is failing, haven't looked closely to see if it's related, but code wise the change looks right

mds1: Hey @garyghayrat just bumping this

garyghayrat: Think the local subgraph being mistakenly set to polygon one was causing the CI problems. https://github.com/ScopeLift/umbra-protocol/pull/591/commits/a4df87bd466edad4ec03c0a65c32a07d6f171dcc should fix it.

PR #592, opened 2023-08-10
Use em instead of rem for icon size

### LINKED ISSUE #584: Remove repetitive console warnings
On the receive page, once signed, there are repetitive console warnings like: 
`Unexpected value 4rem parsing height attribute.` 
`Unexpected value 1rem parsing width attribute.`
`Unexpected value 1rem parsing height attribute.`

Most likely from the loading icons that were added to the account receive table component.

garyghayrat:
closes #584

### DIFF
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -348,7 +348,7 @@
         <q-icon name="fas fa-check" class="text-positive q-mr-sm" /> {{ $t('AccountReceiveTable.scanning-complete') }}
       </div>
       <div v-else-if="scanStatus === 'scanning'" class="text-caption text-right q-mt-md" style="opacity: 0.5">
-        <progress-indicator customClass="q-mr-sm" :percentage="scanPercentage" size="2rem" />
+        <progress-indicator customClass="q-mr-sm" :percentage="scanPercentage" size="2em" />
         {{ $t('Receive.scanning') }}
       </div>
       <div v-else-if="scanStatus === 'fetching'" class="text-caption text-right q-mt-md" style="opacity: 0.5">
--- a/frontend/src/components/AccountReceiveTable.vue
+++ b/frontend/src/components/AccountReceiveTable.vue
@@ -100,7 +100,7 @@
                         ? formatNameOrAddress(props.row.formattedFrom)
                         : formatNameOrAddress(props.row.from)
                     }}</span>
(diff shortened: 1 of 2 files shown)


frontend/src/components/AccountReceiveTable.vue
```diff
@@ -100,7 +100,7 @@
                         ? formatNameOrAddress(props.row.formattedFrom)
                         : formatNameOrAddress(props.row.from)
                     }}</span>
-                    <loading-spinner v-if="!props.row.formattedFrom" size="1rem" customClass="q-ml-sm" />
+                    <loading-spinner v-if="!props.row.formattedFrom" size="16px" customClass="q-ml-sm" />
```
mds1, line 103: Hm do you know why rem was a problem? It seems odd because the quasar docs [say](https://quasar.dev/vue-components/spinners) that rem is supported. Rem is nicer than px because px is absolute and rem is relative, so if the user changes the browser font size, rem units scale accordingly but pixel units don't

garyghayrat: Ahh didn't realize rem had those benefits. Did some testings, those warnings are only showing up when using Firefox. Chromium-based browsers seem to be fine with rem.

garyghayrat: Using `em` gets rid of those errors. Should I use that instead or close this PR entirely.

mds1: Are things not rendering correctly in firefox because of the rem unit, or do they render correctly and we just want to get rid of the warning? 

Either way, em and rem are both relative so using em is just as good. But they are relative to a different base, so you might need to change the values (or not, if both bases are the same value)

garyghayrat: they are rendering correctly but wanted to get rid of the warnings.

garyghayrat: Updated to use em here: https://github.com/ScopeLift/umbra-protocol/pull/592/commits/8cd7ef181f8438eb2ab20a9afc5fc05bf927479d

alexkeating: This might be what is going on here. Although, I am surprised Firefox doesn't support SVG 2 yet

https://stackoverflow.com/questions/46336500/svg-text-layout-rendering-differently-in-chrome-firefox-edge/46337386#46337386

alexkeating (approved):

mds1 (approved):

PR #594, opened 2023-08-13
Change query to use id instead of timestamp

### LINKED ISSUE #578: Use IDs when fetching announcements
Yea so we should probably revert this back to using IDs, since I don't think timestamp will guarantee unique ordering (it might, but no reason to rely on the ambiguity there)

_Originally posted by @mds1 in https://github.com/ScopeLift/umbra-protocol/pull/558#discussion_r1273855710_

alexkeating:
Closes #578

### DIFF
--- a/umbra-js/src/classes/Umbra.ts
+++ b/umbra-js/src/classes/Umbra.ts
@@ -732,7 +732,7 @@ async function* recursiveGraphFetch(
   overrides?: GraphFilterOverride
 ): AsyncGenerator<any[]> {
   // retrieve the last ID we collected to use as the starting point for this query
-  const fromTimestamp = before.length ? (before[before.length - 1].timestamp as string | number) : false;
+  const fromId = before.length ? (before[before.length - 1].id as string | number) : false;
   let startBlockFilter = '';
   let endBlockFilter = '';
   const startBlock = overrides?.startBlock ? overrides.startBlock.toString() : '';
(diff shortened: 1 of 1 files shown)


mds1 (approved):

umbra-js/src/classes/Umbra.ts
```diff
@@ -734,7 +734,7 @@ async function* recursiveGraphFetch(
   overrides?: GraphFilterOverride
 ): AsyncGenerator<any[]> {
   // retrieve the last ID we collected to use as the starting point for this query
-  const fromTimestamp = before.length ? (before[before.length - 1].timestamp as string | number) : false;
```
garyghayrat, line 737: Fetching is ending up in some kind of a loop.

https://github.com/ScopeLift/umbra-protocol/assets/61768337/926aca5f-177d-4844-a2be-36a4f8513711

alexkeating: https://github.com/ScopeLift/umbra-protocol/pull/594/commits/35983bb1b34f5f3dcaec573dd39a19a51fcd046e

garyghayrat: https://github.com/ScopeLift/umbra-protocol/assets/61768337/4998c478-ad31-42d0-a2e4-43ae27a6f0fe

It's now not showing older txs that, on production, shows up after the second/full scan

garyghayrat: > 

Couldn't reproduce this more than once. Can ignore it for now.

garyghayrat (approved):

Issue #362, opened 2022-07-06
Friction for advanced mode


garyghayrat:
- Make sending using recipient's standard public key in advanced mode harder

Ideas:
1. When user checks the checkbox, have a pop-up warning to make sure users understand.
2. Another pop-up when user clicks Send.
3. Any other ideas?

mds1:
Related to 1 and 2, one idea is ensure the warning text makes it clear that if you mess up while using advanced mode funds may be unrecoverable, i.e. make sure the warning is scary 😅

alexkeating:
I will also pick this one up this week

PR #595, opened 2023-08-13
Close modal when submit is clicked

### LINKED ISSUE #511: Advanced form warning does not close on submit
When submitting in advanced mode with use public key checked the final confirmation modal does not close. It should close when the submit button is clicked.

https://github.com/ScopeLift/umbra-protocol/pull/494/files#diff-428f8140870a026c3dd9dc90b835bd0a7deb07a7c8bb8f629ee8e7f9f7156baaR624
image

_Originally posted by @mds1 in https://github.com/ScopeLift/umbra-protocol/pull/494#discussion_r1197941494_

alexkeating:
#### Description

When submitting in advanced mode with use public key checked the final confirmation modal does not close. Now, it should close.

Closes #511

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -873,6 +873,7 @@ function useSendForm() {
       if (!recipientId.value || !token.value || !humanAmount.value) throw new Error(tc('Send.please-complete-form'));
       if (!signer.value) throw new Error(tc('Send.wallet-not-connected'));
       if (!umbra.value) throw new Error('Umbra instance not configured');
+      showAdvancedSendWarning.value = false;
 
       // Verify the recipient ID is valid. (This throws if public keys could not be found. This check is also
       // done in the Umbra class `send` method, but we do it here to throw before the user pays for a token approval.
@@ -998,7 +999,6 @@ function useSendForm() {
       resetForm();
     } finally {
       isSending.value = false;
-      showAdvancedSendWarning.value = false;
     }
   }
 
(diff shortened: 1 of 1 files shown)


frontend/src/pages/AccountSend.vue
mds1, line 1: 1. It looks like we show the "warn about checkbox" modal twice: Once when the "Send using recipient's standard public key" box is checked, and again when hitting send. I confirmed this behavior was not changed on this PR. Regardless, I can't remember if this was intentional/desired? @apbendi do you remember?

2. Additionally, after the "advanced with box checked" send completes, it seems the "send" button stays disabled. I don't recall this happening on main but I did not verify

3. As shown below, the "send" button does not enter the "disabled with loading spinner" state immediately after I click it, but it should. In the below image you can see button is not disabled even though MM is open. Once I sign, then it enters the "disabled with loading spinner" state. I don't think this was introduced on this PR, but just mentioning it and @alexkeating will let you decide if you want to fix it here or in a separate PR

image

alexkeating: 1. From my memory this is intentional, #362 
2. I may be confused here. But, after the send completes the form is cleared which I think is what is causing the send to be disabled. From my understanding this is expected behavior in the app.
3. I can fix this in this pr. It should be as simple as moving 1 line

mds1: 1. Got it
2. Ah right of course 🤦
3. Thakns!

garyghayrat (approved):
🚢 it

frontend/src/pages/AccountSend.vue
alexkeating: https://github.com/ScopeLift/umbra-protocol/pull/595/commits/249ec718d12919c7d1801e76acf3161f750eb99b

PR #597, opened 2023-08-14
Enable BatchSend on Arbitrum and add `isValidBatchSendAmount`

### LINKED ISSUE #601: Batch send page `Amount` field doesn't check the total amount of the token that's being sent
No warnings when the total amount of ETH and DAI being sent exceeds wallet balance:

https://github.com/ScopeLift/umbra-protocol/assets/61768337/5d153916-00f3-47c5-b6a9-5f14d5a2d242

garyghayrat:
`isValidBatchSendAmount` function closes #601 

https://github.com/ScopeLift/umbra-protocol/assets/61768337/3f7c694f-ea65-4319-b506-294d6876e0ec

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -341,7 +341,7 @@
                       placeholder="0"
                       :appendButtonDisable="!Send.receiver || !isValidRecipientId"
                       lazy-rules
-                      :rules="(value : string) => isValidTokenAmount(value, Send.token)"
+                      :rules="(value: string) => isValidBatchSendAmount(value, Send.token)"
                       ref="humanAmountBaseInputRef"
                     />
                   </q-card-section>
@@ -391,7 +391,7 @@
                     :appendButtonDisable="!Send.receiver || !isValidRecipientId"
                     lazy-rules
                     :label="$t('Send.amount')"
-                    :rules="(value : string) => isValidTokenAmount(value, Send.token)"
+                    :rules="(value: string) => isValidBatchSendAmount(value, Send.token)"
                     ref="humanAmountBaseInputRef"
                     class="input-container-amount"
                   />
@@ -573,7 +573,7 @@ function useSendForm() {
   // Batch Send Form Parameters
   const batchSends = ref<BatchSendData[]>([]);
   const tab = ref('send');
-  const batchSendSupportedChains = [5, 11155111];
+  const batchSendSupportedChains = [42161, 11155111];
   const batchSendIsSupported = ref(false);
 
   // Computed form parameters.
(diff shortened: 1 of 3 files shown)


frontend/src/pages/AccountSend.vue
```diff
@@ -866,6 +866,22 @@ function useSendForm() {
...
 
+  function isValidBatchSendAmount(val: string | undefined, tokenInput?: TokenInfoExtended | null | undefined) {
+    const tokenToUse = tokenInput || token.value;
+    if (!tokenToUse) return tc('Send.select-a-token');
+
+    const { address: tokenAddress, decimals } = tokenToUse;
+
+    // Check total Batch Send amount
+    const totalBatchSendAmount = summaryAmount.value.get(tokenToUse);
+    if (totalBatchSendAmount) {
```
mds1, line 877: `totalBatchSendAmount` is a string, so this feels like an odd way to handle the conditional since `Boolean('0')` is true but if the amount is 0 we can skip this block. Maybe convert to BigNumber first and check that it's greater than 0?

frontend/src/pages/AccountSend.vue
```diff
@@ -866,6 +866,22 @@ function useSendForm() {
...
+      const totalAmount = parseUnits(totalBatchSendAmount, decimals);
+      if (totalAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
+    }
+
+    return isValidTokenAmount(val, tokenInput);
```
mds1, line 882: Can we add comments to document the input args and purpose of this method? At first I thought `val` was intended to be the total batch send amount and the method was supposed to validate the full batch send. But then I thought it's supposed to be an individual send amount for the sole purpose of being passed to `isValidTokenAmount` to validate an individual row? But actually, it looks like it does both? 

Maybe I'm missing some context, but there's no accompanying issue or PR description, and IMO there should always be at least one :)

garyghayrat: Absolutely. Linked an issue and updated PR description. Added comments to the code block in https://github.com/ScopeLift/umbra-protocol/pull/597/commits/218b5c6c5171bc9b8e044c8c049f2d57cddb495c

frontend/src/pages/AccountSend.vue
garyghayrat: It was there to check that `totalBatchSendAmount` isn't undefined. `parseUnits` below only takes in strings.
I've updated the check to     
```
if (totalBatchSendAmount && BigNumber.from(totalBatchSendAmount).gt(0)) {
```
in https://github.com/ScopeLift/umbra-protocol/pull/597/commits/218b5c6c5171bc9b8e044c8c049f2d57cddb495c

frontend/src/pages/AccountSend.vue
```diff
@@ -866,6 +866,22 @@ function useSendForm() {
     return true;
   }
 
+  function isValidBatchSendAmount(val: string | undefined, tokenInput?: TokenInfoExtended | null | undefined) {
```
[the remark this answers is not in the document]
garyghayrat: Makes sense, thank you! Updated in https://github.com/ScopeLift/umbra-protocol/pull/597/commits/218b5c6c5171bc9b8e044c8c049f2d57cddb495c

frontend/src/pages/AccountSend.vue
mds1, line 1: I have a balance of 10k DAI, seeing this error when I hit the tab key after typing 6000.

image
image

mds1: Seems like the `summaryAmount` is storing the values with commas because it calls `humanizeArithmeticResult` before putting them into the map. That method is intended for display/UI purposes only, to format values to a certain number of decimals to avoid showing too many decimals in the UI. Using that internally to place values in `summaryAmount` causes this comma issue, and also means we lose precision if a user enters a lot of decimals. 
image

garyghayrat: How about this https://github.com/ScopeLift/umbra-protocol/pull/597/commits/6ffdec91f09074e34e77e7bd48d51e6ccbfa587b for a quick fix.

garyghayrat: precision loss right now is when someone enters 14+ digits.

PR #598, opened 2023-08-14
Bump umbra-js to version 0.1.3 for publication

apbendi:


alexkeating (approved):

umbra-js/package.json
mds1, line 1: Is this PR dependent on PRs #588, #591, and #594 being merged first? Those PRs all touch umbra-js, but I'm not sure if the purpose of this PR is a hotfix to unblock something then we'll put out 0.1.4 after those are all merged, or if this is not urgent and we should wait for all of those before merging this

mds1: This commit is because we released 0.1.3 on npm, so this needs to be merged before the open PRs

mds1 (approved):

PR #600, opened 2023-08-16
Add Gitcoin round 18 banner

garyghayrat:


frontend/src/layouts/BaseLayout.vue
```diff
@@ -1,19 +1,19 @@
...
+            href="https://explorer.gitcoin.co/#/round/10/0x8de918f0163b2021839a8d84954dd7e8e151326d/0x8de918f0163b2021839a8d84954dd7e8e151326d-87"
             target="_blank"
             >{{ $t('Home.development-grant') }}</a
           >
         </template>
         <template v-slot:LearnMore>
           <a
             class="alert-banner-hyperlink"
-            href="https://twitter.com/UmbraCash/status/1651702157345976320"
+            href="https://explorer.gitcoin.co/#/round/10/0x8de918f0163b2021839a8d84954dd7e8e151326d/0x8de918f0163b2021839a8d84954dd7e8e151326d-87"
```
alexkeating, line 16: Should this stay the twitter? It is a little confusing having the same link twice.

garyghayrat: Doesn't look we tweeted anything yet. Let's update it after we have tweeted?

alexkeating: Good point that makes sense. I don't have a strong opinion so feel free to keep things as is, but it may also make sense to remove the learn more until we have tweeted.

alexkeating (approved):

frontend/src/layouts/BaseLayout.vue
apbendi: Merging as-is to get it out there. We can update later.

apbendi (approved):

PR #603, opened 2023-08-22
Support multicall with CNS

### LINKED ISSUE #507: feat: support batch CNS looksups
We use `lookupOrReturnAddresses` to batch ENS lookups into fewer RPC calls, but don't have the equivalent batch lookup for CNS names.

Ref https://github.com/ScopeLift/umbra-protocol/pull/494#discussion_r1185313690

alexkeating:
#### Description

- Support multicall for looking up CNS name

Closes #507


frontend/src/utils/address.ts
```diff
@@ -154,11 +153,65 @@ export const lookupOrReturnAddresses = async (addresses: string[], provider: Pro
...
+const lookupCNSNameBatchPolygon = async (addresses: string[]) => {
+  return await lookupCNSNameBatch(addresses, '0xa9a6A3626993D487d2Dbda3173cf58cA1a9D9e9f', POLYGON_PROVIDER);
+};
+
+const lookupCnsNameBatch = async (addresses: string[]) => {
+  try {
+    const [resultL1, resultL2] = await Promise.all([
+      lookupCNSNameBatchMainnet(addresses),
+      lookupCNSNameBatchPolygon(addresses),
+    ]);
```
mds1, line 197: Is this correct: a name can be registered on mainnet OR polygon only, so we check both, and those are the only two chains you can register on which is why we don't check any others?

frontend/src/utils/constants.ts
```diff
@@ -2,6 +2,8 @@ import { StaticJsonRpcProvider } from 'src/utils/ethers';
 
 export const MAINNET_RPC_URL = String(process.env.MAINNET_RPC_URL);
 export const MAINNET_PROVIDER = new StaticJsonRpcProvider(MAINNET_RPC_URL);
+export const POLYGON_MAINNET_RPC_URL = `https://polygon-mainnet.infura.io/v3/${String(process.env.INFURA_ID)}`;
```
mds1, line 5: Probably should make this a generic RPC URL like we do with `MAINNET_RPC_URL`, so it's easier to switch RPC providers if needed

frontend/src/utils/address.ts
alexkeating: Yes, outside of testnets. I based a lot of this off of their resolution library. It was unfortunate they didn't support batching out of the box.

https://docs.unstoppabledomains.com/smart-contracts/contract-reference/uns-smart-contracts/#unsregistry

frontend/src/utils/constants.ts
alexkeating: https://github.com/ScopeLift/umbra-protocol/pull/603/commits/2c605026ced01da3a6379484abad317afc17c3fd

apbendi (approved):
lgtm!

PR #608, opened 2023-08-25
Use BigNumber for `summaryAmount` and use `humanizeTokenAmount`

### LINKED ISSUE #607: Update `summaryAmount` to use BigNumber
Seems like the `summaryAmount` is storing the values with commas because it calls `humanizeArithmeticResult` before putting them into the map. That method is intended for display/UI purposes only, to format values to a certain number of decimals to avoid showing too many decimals in the UI. Using that internally to place values in `summaryAmount` causes this comma issue, and also means we lose precision if a user enters a lot of decimals. 
image

_Originally posted by @mds1 in https://github.com/ScopeLift/umbra-protocol/pull/597#discussion_r1300771794_

garyghayrat:
closes #607

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -600,20 +600,19 @@ function useSendForm() {
 
   // Batch Send Computed Form Parameters
   const batchSendHumanTotalAmount = computed(() => {
-    let ethTotal;
+    let ethSendAmount: BigNumber | undefined;
     for (const token of summaryAmount.value.keys()) {
       if (token.symbol === NATIVE_TOKEN.value.symbol) {
-        ethTotal = summaryAmount.value.get(token)?.toString();
+        ethSendAmount = summaryAmount.value.get(token);
       }
     }
-    if (typeof ethTotal !== 'string') return '--'; // appease TS
-    if (isNaN(Number(ethTotal))) return '--';
-    const sendAmount = BigNumber.from(ethTotal);
+    if (!ethSendAmount) return '--';
+    const sendAmount = BigNumber.from(ethSendAmount);
     const totalAmount = sendAmount.add(toll.value.mul(batchSends.value.length));
 
     return humanizeArithmeticResult(
       totalAmount,
-      [ethTotal, batchSendHumanToll.value], // subtotal and fee
+      [ethSendAmount.toString(), batchSendHumanToll.value], // subtotal and fee
       NATIVE_TOKEN.value
     );
   });
@@ -870,10 +869,7 @@ function useSendForm() {
 
     // Get total batch send amount for the token
     const totalBatchSendAmount = summaryAmount.value.get(tokenToUse) || BigNumber.from(0);
-    // Check totalBatchSendAmount is defined and is greater than zero.
-    if (totalBatchSendAmount && totalBatchSendAmount.gt(0)) {
-      if (totalBatchSendAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
-    }
+    if (totalBatchSendAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
 
     return isValidTokenAmount(val, tokenInput);
   }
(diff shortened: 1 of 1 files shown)


frontend/src/pages/AccountSend.vue
```diff
@@ -603,12 +603,12 @@ function useSendForm() {
...
+        ethTotal = summaryAmount.value.get(token)?.toString();
       }
     }
     if (typeof ethTotal !== 'string') return '--'; // appease TS
     if (isNaN(Number(ethTotal))) return '--';
-    const sendAmount = parseUnits(ethTotal, NATIVE_TOKEN.value.decimals);
+    const sendAmount = BigNumber.from(ethTotal);
     const totalAmount = sendAmount.add(toll.value.mul(batchSends.value.length));
 
     return humanizeArithmeticResult(
```
mds1, line 614: I think we can clean this up a bit now:
- On L606, remove the `.toString()` so the type of `ethTotal` is either `BigNumber | undefined`
- L609-610 can then be collapsed into `if (!ethTotal) return '--';
- L611 can be removed now because `ethTotal` is already a `BigNumber`

I haven't tested this code but it should be the same
```typescript
const batchSendHumanTotalAmount = computed(() => {
  let ethSendAmount: BigNumber | undefined;
  for (const token of summaryAmount.value.keys()) {
    if (token.symbol === NATIVE_TOKEN.value.symbol) {
      ethSendAmount = summaryAmount.value.get(token);
    }
  }
  if (!ethSendAmount) return '--';
  const totalAmount = ethSendAmount.add(toll.value.mul(batchSends.value.length));

  return humanizeArithmeticResult(
    totalAmount,
    [ethSendAmount.toString(), batchSendHumanToll.value], // subtotal and fee
    NATIVE_TOKEN.value
  );
});
```

frontend/src/pages/AccountSend.vue
```diff
@@ -872,14 +866,13 @@ function useSendForm() {
...
     // Get total batch send amount for the token
-    const totalBatchSendAmount = summaryAmount.value.get(tokenToUse)?.replace(/,/g, '') || '0';
+    const totalBatchSendAmount = summaryAmount.value.get(tokenToUse) || BigNumber.from(0);
     // Check totalBatchSendAmount is defined and is greater than zero.
-    const totalAmount = parseUnits(totalBatchSendAmount, decimals);
-    if (totalBatchSendAmount && totalAmount.gt(0)) {
-      if (totalAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
+    if (totalBatchSendAmount && totalBatchSendAmount.gt(0)) {
+      if (totalBatchSendAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
     }
```
mds1, line 876: This can also be cleaned up:
- `totalBatchSendAmount` is a `BigNumber` so the `if (totalBatchSendAmount)` portion always evaluates to true
- if `totalBatchSendAmount.gt(balances.value[tokenAddress])` is true then `totalBatchSendAmount.gt(0)` must be true, so the latter check is redundant

```suggestion
if (totalBatchSendAmount.gt(balances.value[tokenAddress])) return `${tc('Send.total-amount-exceeds-balance')}`;
```

frontend/src/pages/AccountSend.vue
```diff
@@ -1092,7 +1085,7 @@ function useSendForm() {
         if (token.symbol !== NATIVE_TOKEN.value.symbol) {
           const tokenContract = new Contract(token.address, ERC20_ABI, signer.value);
           const batchSendAddress = umbra.value?.batchSendContract!.address;
-          const parsedAmount = parseUnits(amount.replace(/,/g, '') || '0', token.decimals);
+          const parsedAmount = parseUnits(amount.toString() || '0', token.decimals);
```
mds1, line 1088: Amount is a `BigNumber` so I don't think we'll ever hit the `|| '0'` branch right?
```suggestion
          const parsedAmount = parseUnits(amount.toString(), token.decimals);
```

Also I'm a bit confused about what's going on here and not sure if this is correct. Above, we set values in the map using `const updatedAmount = previousAmount.add(parseUnits(send.amount, send.token.decimals))` which means that amount should already have the correct decimals in the map. For example, 1 USDC gets saved here as 1000000, which means that on this line we are again scaling up that amount by 6 decimals, which seems incorrect. Am I misunderstanding something here?

frontend/src/pages/AccountSend.vue
garyghayrat: Updated here: https://github.com/ScopeLift/umbra-protocol/pull/608/commits/3182021b042ae7e432d87cb848fffebf018f00f0

frontend/src/pages/AccountSend.vue
```diff
@@ -600,63 +600,56 @@ function useSendForm() {
 
   // Batch Send Computed Form Parameters
   const batchSendHumanTotalAmount = computed(() => {
-    let ethTotal;
+    let ethSendAmount: BigNumber | undefined;
```
alexkeating, line 603: I think the types can be removed here

alexkeating (approved):
Looks good, one nit

frontend/src/pages/AccountSend.vue
garyghayrat: Sounds good, removed here: https://github.com/ScopeLift/umbra-protocol/pull/608/commits/b1309b86c0f4b92fb444cef47d970b4a6a93e213

PR #609, opened 2023-08-25
Copy typechain into umbra-js

wildmolasses:
This is a quick fix to unblock api development that @apbendi, @jferas, and I discussed. By copying the typechain stuff into umbra-js before build, we don't need to publish contracts-core.


jferas (approved):
Looks good.. yarn build puts stuff where it needs to

PR #614, opened 2023-09-06
Add error msg when an ENS name doesn't have address record

garyghayrat:


umbra-js/src/utils/utils.ts
```diff
@@ -180,6 +180,7 @@ export async function toAddress(name: string, provider: EthersProvider) {
   let address: string | null = null;
   address = await resolveEns(name, provider); // will never throw, but returns null on failure
   if (address) return address;
+  if (name.endsWith('.eth')) throw 'Please check if the ENS name has a valid address record.';
```
mds1, line 183: Most valid DNS domains can also be ENS names, so if the intent here is "if this was an ENS name, but it didn't resolve to an address, throw error", then this should come after we try CNS. In other words:
1. Try resolving ENS domain
2. Try resolving CNS domain
3. If both fail, and `name` is a domain, throw error

Also, it's better to `throw new Error('my message')` instead of throwing a string like this, since throwing an Error object will include a stack trace

garyghayrat: I see, thank you! Is this a good way to check if it's a domain? https://github.com/ScopeLift/umbra-protocol/pull/614/commits/02354570ce789e75524c39fb483bf9c2d2ec0df4

mds1: Hmm, yea I think that is probably ok, I don't know of any name systems that work without periods

umbra-js/src/utils/utils.ts
```diff
@@ -185,6 +185,8 @@ export async function toAddress(name: string, provider: EthersProvider) {
   address = await resolveCns(name); // will never throw, but returns null on failure
   if (address) return address;
 
+  if (name.includes('.')) throw new Error('Please check if the ENS name has a valid address record.');
```
mds1, line 188: There are other issues that could trigger this right? Such as the name not being registered at all? So maybe we should say `Please verify the name is correct, registered, has a valid address record.`

alexkeating (approved):

mds1 (approved):

PR #623, opened 2023-10-23
Enable batchSend on all networks

garyghayrat:
- [x] Still need to add supported tokens to `approveToken` on `batchSend` contracts on each network.


apbendi (approved):
🚢 :shipit:

PR #640, opened 2024-01-22
Check recipient addresses for ENS confusables in batch send

### LINKED ISSUE #530: Add check confusable to batch send and format error message to not overlap send
Screenshot from 2023-06-05 15-48-26

Change formatting so it does not overlap and also check for confusables.

### LINKED ISSUE #551: The warning about ENS Confusables is not shown for ENS names entered on the batch send page

jferas:
Resolves #551 
Resolves #530

### DIFF
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -321,6 +321,7 @@
                       <br />
                     </div>
                     <div class="text-caption warning-container" v-if="batchSends[index].warning">
+                      <br />
                       {{ batchSends[index].warning }}
                     </div>
                     <!-- Token -->
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -826,7 +826,6 @@ function useSendForm() {
 
     // Check if recipient ID is valid
     try {
-      // Check if ENS name is valid
       await umbraUtils.lookupRecipient(val, provider.value as Provider, {
         advanced: shouldUseNormalPubKey.value,
       });
@@ -1228,7 +1227,7 @@ function useSendForm() {
         assertValidEnsName(recipientIdString);
       }
       setWarning('', index); // clear warning
-    } catch (e: unknown) {
+    } catch (e) {
       if (e instanceof Error && e.message) {
         setWarning(e.message, index);
       }
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -320,7 +320,7 @@
                     <div v-if="batchSends[index].validationError">
                       <br />
                     </div>
-                    <div class="warning-container" v-if="batchSends[index].warning">
+                    <div class="text-caption warning-container" v-if="batchSends[index].warning">
                       {{ batchSends[index].warning }}
                     </div>
                     <!-- Token -->
@@ -414,8 +414,10 @@
                   <div><br /></div>
                 </div>
                 <div class="batch-send" v-if="batchSends[index].warning">
-                  <div class="batch-send-warning-container">
-                    <br />
+                  <div class="text-caption batch-send-warning-container">
+                    <div v-for="n in numberOfWarningBreaksNeeded" :key="n">
+                      <br v-if="batchSends[index].validationError" />
+                    </div>
                     {{ batchSends[index].warning }}
                   </div>
                   <p class="input-container-token"></p>
@@ -478,7 +480,6 @@
                 :flat="true"
                 :label="$t('Send.add-send')"
               />
-
               <div>
                 <router-link :class="{ 'no-text-decoration': true, 'dark-toggle': true }" :to="{ name: 'sent' }">
                   <div class="row items-center justify-center q-pa-sm link-container rounded-borders">
@@ -590,6 +591,7 @@ function useSendForm() {
   const tab = ref('send');
   const batchSendSupportedChains = [1, 10, 100, 137, 42161, 11155111];
   const batchSendIsSupported = ref(false);
+  const numberOfWarningBreaksNeeded = ref(0);
 
   // Computed form parameters.
   const showAdvancedWarning = computed(() => advancedAcknowledged.value === false && useNormalPubKey.value === true);
@@ -821,6 +823,14 @@ function useSendForm() {
   }
 
   function setWarning(warning: string, index: number | undefined) {
+    // Width setpoints for extra break(s) needed before warning message on the batch send page
+    // These are eeded to prevent the error message from overlapping with the warning field field,
+    // as page width and layout dictate the number of error lines (and therefore) breaks needed.
+    const warningWidths = [1140, 955, 790, 710, 685, 645, 630];
+    numberOfWarningBreaksNeeded.value = 0;
+    for (const width of warningWidths) {
+      if (window.innerWidth < width) numberOfWarningBreaksNeeded.value += 1;
+    }
     if (index !== undefined) {
       batchSends.value[index].warning = warning;
     } else {
@@ -1278,6 +1288,7 @@ function useSendForm() {
     isValidTokenAmount,
     isValidBatchSendAmount,
     NATIVE_TOKEN,
+    numberOfWarningBreaksNeeded,
     onFormSubmit,
     onBatchSendFormSubmit,
     paymentLinkParams,
--- a/frontend/src/pages/AccountSend.vue
+++ b/frontend/src/pages/AccountSend.vue
@@ -116,7 +116,7 @@
             placeholder="vitalik.eth"
             lazy-rules
             :hideBottomSpace="true"
-            :rules="isValidId"
+            :rules="(value: string) => isValidId(value)"
             ref="recipientIdBaseInputRef"
           />
           <div class="flex row text-caption warning-container q-pb-sm" v-if="recipientIdWarning">
@@ -315,7 +315,7 @@
                       :disable="isSending"
                       placeholder="vitalik.eth"
                       lazy-rules
-                      :rules="isValidId"
+                      :rules="(value: string) => isValidId(value)"
                     />
 
                     <!-- Token -->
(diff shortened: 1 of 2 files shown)


frontend/src/pages/AccountSend.vue
```diff
@@ -806,11 +806,14 @@ function useSendForm() {
   // Validators
   async function isValidId(val: string | undefined) {
     // Return true if nothing is provided
-    checkConfusables();
     if (!val) return true;
 
     // Check if recipient ID is valid
     try {
+      // Check if confusable chars in string, throws with warning if so
+      checkConfusables(val);
```
alexkeating, line 814: I think this change causes a confusable character to show an error on the frontend rather than a warning. This will prevent users from sending to users that have confusable characters in their ENS name. @garyghayrat is this the behavior we want?

frontend/src/pages/AccountSend.vue
```diff
@@ -366,7 +366,7 @@
                       placeholder="vitalik.eth"
                       :label="$t('Send.receiver-addr-ens')"
                       lazy-rules
-                      :rules="isValidId"
+                      :rules="(value: string) => isValidId(value)"
```
alexkeating, line 369: Is this change necessary? Is there a different behavior when the rules are defined this way?

frontend/src/pages/AccountSend.vue
```diff
@@ -1206,18 +1209,11 @@ function useSendForm() {
...
-      recipientIdWarning.value = undefined;
-    } catch (e) {
-      if (e instanceof Error) {
-        recipientIdWarning.value = e.message;
-      }
+  function checkConfusables(recipientIdString: string | undefined) {
+    if (recipientIdString && recipientIdString.endsWith('eth')) {
+      assertValidEnsName(recipientIdString);
     }
+    recipientIdWarning.value = undefined;
```
alexkeating, line 1216: Is this needed? Will this supress all warnings?

frontend/src/pages/AccountSend.vue
jferas: > Is this change necessary? Is there a different behavior when the rules are defined this way?

I think this is necessary because we need to pass the value of the text input object being changed into the function `isValidId` (and ultimately into `checkConfusables`). If we left it the way it was, inside `checkConfusables`.. where we do the following:
```
    const recipientIdString = recipientId.value || '';
```
We end up always getting the contents of the text object on the send page but not one of the objects on the batch send page. So we end up checking confusables on the wrong string.

frontend/src/pages/AccountSend.vue
jferas: > I think this change causes a confusable character to show an error on the frontend rather than a warning. This will prevent users from sending to users that have confusable characters in their ENS name. @garyghayrat is this the behavior we want?

Ah... I see... looks like I've made this too restrictive.. thanks.. I'll fix

garyghayrat: I'd say warning is sufficient.

frontend/src/pages/AccountSend.vue
jferas: Gone: [34740c2](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/34740c2777b044c67badac8536a056d80aadd2ca)

frontend/src/pages/AccountSend.vue
jferas: Restored to warning.. with a new warning element associated with each batch send entry: [34740c2](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/34740c2777b044c67badac8536a056d80aadd2ca)

frontend/src/pages/AccountSend.vue
```diff
@@ -116,7 +116,7 @@
             placeholder="vitalik.eth"
             lazy-rules
             :hideBottomSpace="true"
-            :rules="isValidId"
+            :rules="(value: string) => isValidId(value, undefined)"
```
alexkeating, line 119: (Arbitrary thread)

- Still reviewing, but I found a couple weird scenarios

- It takes a while for the error to show up on this one.
Screenshot from 2024-01-26 09-43-34

- Some state must not be getting cleared
[Screencast from 2024-01-26 11-01-27.webm](https://github.com/ScopeLift/umbra-protocol/assets/12705146/1de2d19d-9f5b-490d-86be-dffc0e712026)

jferas: Thanks for the review! The warning state not getting cleared issue is resolve here: [f8ff428](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/f8ff428ffc58a3d2219fa6b13724c43538b3e840)

The issue related to the slow error appearance on the batch send form looks like it is also present in the existing release, and appears to be related to the way that the "batchSend"s are "watched" and validations processed.
Might be a good idea to make a separate PR for this issue.

frontend/src/pages/AccountSend.vue
```diff
@@ -803,14 +808,25 @@ function useSendForm() {
...
+    // Check if confusable chars in string, throws with warning if so
+    checkConfusables(val, index);
+
     // Return true if nothing is provided
-    checkConfusables();
     if (!val) return true;
 
     // Check if recipient ID is valid
     try {
+      // Check if ENS name is valid
```
alexkeating, line 829: I think this can be removed now

frontend/src/pages/AccountSend.vue
```diff
@@ -1206,16 +1222,15 @@ function useSendForm() {
...
     try {
       if (recipientIdString && recipientIdString.endsWith('eth')) {
         assertValidEnsName(recipientIdString);
       }
-      recipientIdWarning.value = undefined;
-    } catch (e) {
-      if (e instanceof Error) {
-        recipientIdWarning.value = e.message;
+      setWarning('', index); // clear warning
+    } catch (e: unknown) {
```
alexkeating, line 1231: I think this `unknown` can be removed. Explicit type declarations can cause issues down the line if they are un-needded

alexkeating (commented):
Looks good! A couple nits and this should be ready

alexkeating (approved):

frontend/src/pages/AccountSend.vue
jferas: Ended up putting the fix for this in the PR after all, as it was so small: https://github.com/ScopeLift/umbra-protocol/pull/640/commits/62a4bf2f9fb0ebf688b32ff9627ab114a334713e

alexkeating: The formatting is a little weird in the batch view. This should probably be resolved if possible.

Screenshot from 2024-01-30 14-48-52
Screenshot from 2024-01-30 14-49-57
Screenshot from 2024-01-30 14-50-24

jferas: Good point.. I think the "error overlapping the ID object below" issue also exist in the current release.. but the warning issue didn't (as that's what this PR was intended to solve.. the missing warning). I should be able to get both resolved in the next commit to this PR.

jferas: Fix the overlaps on both batch send mobile and desktop..however, if the page is very narrow (just before it transitions to mobile layout) there is still an overlap of the error message onto the warning message .. this is because the base-input component is dropping down a very narrow paragraph (because of the screen width) of the long error text directly beneath it. [8459e97](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/8459e97d09a7a8e2c77bce4b128362e9dfbbfd53)

garyghayrat: Sorry for the last min nits, but 
1. Could you reduce the distance between the error message and the confusables warning message on Desktop? Ideally to the same distance that they are on the Send page.



2. Could you make the fonts size of the yellow warning message same as the red error message? Maybe they already are and it's just my browser?

alexkeating: Looking good @jferas, I think after these final nits we this will be ready

jferas: Done: fixed the warning/error layout issue for desktop at various widths:https://github.com/ScopeLift/umbra-protocol/pull/640/commits/f783e6d49a57cb3e9afef52eae28e1e003264a2c
Also fixed a small issue that only showed on an actual device where error overlapped warning: https://github.com/ScopeLift/umbra-protocol/pull/640/commits/964b574f18d446ad024bf310670faf227a8cad8a

garyghayrat: A couple more nits, 
1. the x-axis padding or margins are a bit larger than the error message.
SCR-20240205-lwoc

2. On mobile view, the batch send text box distance between the error and the warning are bit larger than Single Send. 

**Single Send:**
image

**Batch Send:**
image

3. Weird behavior where if you switch between Single Send and Batch Send while putting in an unregistered address with confusables, Single Page reacts by hiding the error message, and moving the warning message up closer to the address input box. Batch Send page keeps the warning message in the same spot, with a huge margin where the error message would be. 

**Single Send**:
image

**Batch Send**:
image

jferas: Nits 1 and 2 were relatively minor.. done: [8bb038e](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/8bb038e59c6075f371ddc77b9c7c57a53e162b95)

Nit 3 was kind of tricky.. the error message was automatically being removed by Vue on the tab switch because it was part of the recipient ID base-input component, but the warning message (which was generated into its own object was not getting removed on tab switch, so added logic to do that so that error and warning appeared to be handled the same way.. let me know if you think there's a better way of doing this: [462fac7](https://github.com/ScopeLift/umbra-protocol/pull/640/commits/462fac7ad460f99bc3e316367f31e72709192814)

alexkeating (approved):

garyghayrat (approved):
lgtm, thanks John!

frontend/src/pages/AccountSend.vue
alexkeating: I am getting the slightest of overlaps

Screenshot from 2024-02-12 15-49-05
Screenshot from 2024-02-12 15-48-53

garyghayrat: @jferas will create an issue to track this.

PR #642, opened 2024-01-24
Add button to stop scanning announcements

garyghayrat:
Video shows the following in order:
1. Pausing when fetching latest announcements -> Goes back to receive signing page
2. Pausing when scanning latest announcements -> Goes to the empty table
3. Pausing when fetching all announcements -> ditto.
4. Pausing when scanning all announcements -> ditto.

https://github.com/ScopeLift/umbra-protocol/assets/61768337/dbca894f-6fa5-4fb0-9023-7b1558dc9a56


frontend/src/pages/AccountReceive.vue
```diff
@@ -101,6 +101,14 @@
         </div>
         <div v-else class="text-center text-italic">{{ $t('Receive.fetching-latest') }}</div>
       </div>
+      <div>
+        <base-button
```
alexkeating, line 105: What is the best way to test this? I can't get it to show on my receive page.

garyghayrat: Do you mean you don't see the `Stop` button at all?

alexkeating: I don't see it on the table view, don't know what scenario should trigger it.

image

garyghayrat: The video in the description of this PR should give you some idea, but basically the button should show up when the announcements are being fetched or being scanned.

frontend/src/pages/AccountReceive.vue
```diff
@@ -296,6 +322,8 @@ function useScan() {
         }
         // Wait for the first batch of web workers to finish scanning before creating new workers
         await firstScanPromise;
+        // Clear out existing workers
+        workers.length = 0;
```
jferas, line 326: Just verifying.. we need this because we're gonna fall out and execute this asynce code upon workers being terminated by the 'Stop' button and the fact that workers array is instantiated now in this file and not in the workers.ts anymore means we need to empty the array here?

jferas (approved):
LGTM!  Just left a comment to verify my understanding of how it's working.

alexkeating (approved):

PR #647, opened 2024-02-15
Better frontend handling when API is down

### LINKED ISSUE #367: Better frontend handling when API is down
- Frontend should fail more gracefully and provide better error message when the Umbra-api is unreachable or returning error messages.
- Ideally users should still be able to perform actions relating to the base asset, i.e. sending and withdrawing Eth.

jferas:
Resolves #367 

Made inability to connect to Umbra API a softer error so native tokens can be transferred

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -22,6 +22,7 @@ import { ERC20_ABI, MAINNET_PROVIDER, MULTICALL_ABI, MULTICALL_ADDRESS } from 's
 import { BigNumber, Contract, ExternalProvider, Web3Provider, parseUnits } from 'src/utils/ethers';
 import { UmbraApi } from 'src/utils/umbra-api';
 import { getChainById } from 'src/utils/utils';
+import { notifyUser } from 'src/utils/alerts';
 import useSettingsStore from 'src/store/settings';
 import enLocal from 'src/i18n/locales/en-US.json';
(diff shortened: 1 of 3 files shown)


frontend/src/store/wallet.ts
```diff
@@ -22,6 +22,7 @@ import { ERC20_ABI, MAINNET_PROVIDER, MULTICALL_ABI, MULTICALL_ADDRESS } from 's
 import { BigNumber, Contract, ExternalProvider, Web3Provider, parseUnits } from 'src/utils/ethers';
```
garyghayrat, line 22: _Random placement for comment_

This is actually great, thanks for picking this up. 

One nit is, could we put in a placeholder for the Token logo and Amount when API is not able to fetch those details on the receive table? And maybe handle 'LogoURI' errors a bit better?
image

jferas: Thanks Gary! Great catch.. fixed: [ddf1367](https://github.com/ScopeLift/umbra-protocol/pull/647/commits/ddf136714f66f9f238f2c2239f4013ce75baa7b0)

alexkeating (approved):

frontend/src/store/wallet.ts
garyghayrat: Looks good now 
image

garyghayrat (approved):
lgtm!

PR #648, opened 2024-02-16
Fix fleek deploy failure

alexkeating:
#### Description

- It looks like fleek was failing to deploy because typescript was missing.
- Also, only deploy on master


garyghayrat (approved):
lgtm!

PR #653, opened 2024-02-28
Detect Umbra relayer/API version change and force reload

### LINKED ISSUE #309: Force users to refresh the frontend when there has been a breaking backend change
### Background

In the event that there is ever a serious bug with the frontend or backend, we want to be able to force users to update their clients.

Imagine the following:
* user X loads app.umbra.cash in a new browser tab
* we discover that there is a serious bug in the frontend (and/or how it interacts with the backend)
* we patch the bug and deploy the new code to the server that distributes it
* user X still has the old, vulnerable code running in his/her browser tab
* user X does not refresh his/her tab before using the frontend
* user X uses the stale frontend and BAD THING happens

Unless we have a mechanism to force refresh clients, there will always be a long tail of outdated versions in the wild.

### Implementation

There are two obvious triggers we could use for refreshes:
1. refresh after a certain amount of time (e.g. a week) even if there has been no backend update
2. refresh each time there is a (major|minor|patch) version update to the backend

Trigger 1 has the benefit of refreshing the app even if no backend calls have been made. For example, after you've connected to the network, you can send funds with the protocol without touching the backend again. You just sign the transaction that the frontend creates for you, and your wallet sends it to the network.

I think 2 would probably be all that we need, though. Most wallets will require you to reconnect after a prolonged period away from a dapp, and doing so will hit the `/tokens` route of the backend.

As it is, the backend already returns a version number with each response which we could use:

```js
{
    name: 'Umbra',
>>  version: { major: 0, minor: 0, patch: 1 },
    keywords: [ 'umbra', 'stealth', 'payments', 'defi' ],
    tokens: [...],
    timestamp: '2021-04-23T20:11:22.472Z'
}
```

My suggestion would be to refresh on every major or minor release of the backend. Refreshing on patch releases would likely just result in a lot of refreshing for our users, which isn't great UX.

To implement this, it would probably be sufficient to just store the backend version in a cookie or local storage or something, TBD. When the frontend first loads, it stores the first version it sees from the backend. Then, each time it makes a request to the backend, it confirms that the version in the cookie matches the version coming back from the API. If it doesn't, we force the user to refresh. A power user could always manipulate his/her cookies to get around the check. But I think at that point, it's on them and we just let them do what they want. The protocol is ultimately open anyway.

An alternative to this approach would be to give the frontend a version too. Then on each API request we check that the frontend version is >= the backend version. Then we don't have to worry about power users getting around our protections by modifying cookies (now they'd have to modify the javascript itself). But the downside of this approach is that we'd be coupling frontend and backend deploys: it wouldn't be possible to deploy the backend without first deploying the frontend. We'd also need to be careful: if the backend was deployed before the frontend, users could be stuck in an infinite refresh loop where the frontend keeps telling them to refresh b/c its version lags the backend version, but when it does refresh it just pulls frontend code with a version that is still out of date.

### UX

We probably want to put an overlay on the app to disable it, with copy to the effect of "Your application is out of date, please reload [button]". This is what google sheets does, for example, when I've had a tab open for months. It's straightforward. You just click the button (it's all you can do) and that's it.

jferas:
Resolves #309 

Added most recently obtained version of Umbra relayer/API to local storage..

At startup, on creation of Umbra relayer/API.. and on every interaction between the front-end and the API.. the API version is returned in the API's response to the front-end, and the version and local storage are used in the following way:

1. Attempt to read the most recently saved API version from local storage.
2. If the version does not exist in local storage, save the most recently received version from the API to local storage.
3. If the version does exist in local storage, compare it to the version received from the API.. if the major/minor numbers differ, the following actions are taken:
   a. The saved API version in local storage is cleared.
   b. The user is alerted.
   c. The user is forced to reload the front-end web app.
   d. On that reload, because the saved version in local storage is not present, the save described in step 2 above will occur.

Currently the alert to the user is just that.. a javascript "alert".. but can probably be done in a more "Vue-centric" way.

### DIFF
--- a/frontend/src/utils/umbra-api.ts
+++ b/frontend/src/utils/umbra-api.ts
@@ -40,7 +40,7 @@ export class UmbraApi {
       if (apiVersionFromSettings.major != version.major || apiVersionFromSettings.minor != version.minor) {
         console.log('UmbraAPI: version mismatch, clearing settings and going to force a page reload');
         clearUmbraApiVersion();
-        alert('UmbraAPI: version mismatch, clearing settings and going to force a page reload');
+        alert('UmbraAPI: version outdated, please reload the page');
         window.location.reload();
       }
     }
--- a/frontend/src/utils/umbra-api.ts
+++ b/frontend/src/utils/umbra-api.ts
@@ -28,7 +28,7 @@ export class UmbraApi {
   ) {}
 
   static checkUmbraApiVersion(version: UmbraApiVersion) {
-    const apiVersionFromSettings: UmbraApiVersion | null = getUmbraApiVersion();
+    const apiVersionFromSettings = getUmbraApiVersion();
     if (!apiVersionFromSettings) {
       console.log(`UmbraAPI: no saved version setting, using backend value: ${version.major}.${version.minor}`);
       setUmbraApiVersion(version);
--- a/frontend/src/utils/umbra-api.ts
+++ b/frontend/src/utils/umbra-api.ts
@@ -10,16 +10,41 @@ import {
   RelayResponse,
   TokenListResponse,
   WithdrawalInputs,
+  UmbraApiVersion,
 } from 'components/models';
 import { jsonFetch } from 'src/utils/utils';
 
+import useSettingsStore from 'src/store/settings';
+
+const { getUmbraApiVersion, setUmbraApiVersion, clearUmbraApiVersion } = useSettingsStore();
+
 export class UmbraApi {
   static baseUrl = 'https://mainnet.api.umbra.cash'; // works for all networks
+  // static baseUrl = 'http://localhost:3000';  // use this for testing with a local Umbra API
   constructor(
     readonly tokens: TokenInfoExtended[],
     readonly chainId: number,
     readonly nativeTokenMinSendAmount: string | undefined
   ) {}
+
+  static checkUmbraApiVersion(version: UmbraApiVersion) {
+    const apiVersionFromSettings: UmbraApiVersion | null = getUmbraApiVersion();
+    if (!apiVersionFromSettings) {
+      console.log(`UmbraAPI: no saved version setting, using backend value: ${version.major}.${version.minor}`);
+      setUmbraApiVersion(version);
+    } else {
+      console.log(`UmbraAPI: major.minor version fetched from backend: ${version.major}.${version.minor}`);
+      console.log(
+        `UmbraAPI: major.minor version fetched from setting: ${apiVersionFromSettings.major}.${apiVersionFromSettings.minor}`
+      );
+      if (apiVersionFromSettings.major != version.major || apiVersionFromSettings.minor != version.minor) {
+        console.log('UmbraAPI: version mismatch, clearing settings and going to force a page reload');
+        clearUmbraApiVersion();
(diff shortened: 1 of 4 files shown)


frontend/src/utils/umbra-api.ts
```diff
@@ -10,17 +10,42 @@ import {
...
+      setUmbraApiVersion(version);
+    } else {
+      console.log(`UmbraAPI: major.minor version fetched from backend: ${version.major}.${version.minor}`);
+      console.log(
+        `UmbraAPI: major.minor version fetched from setting: ${apiVersionFromSettings.major}.${apiVersionFromSettings.minor}`
+      );
+      if (apiVersionFromSettings.major != version.major || apiVersionFromSettings.minor != version.minor) {
+        console.log('UmbraAPI: version mismatch, clearing settings and going to force a page reload');
+        clearUmbraApiVersion();
+        alert('UmbraAPI: version mismatch, clearing settings and going to force a page reload');
```
garyghayrat, line 43:
```suggestion
        alert('UmbraAPI: version outdated, please reload the page');
```

jferas: Done: https://github.com/ScopeLift/umbra-protocol/pull/653/commits/fb40c3de0d1d8b75c2a656407265a07e74d7d6a9

frontend/src/utils/umbra-api.ts
```diff
@@ -10,17 +10,42 @@ import {
...
 } from 'components/models';
 import { jsonFetch } from 'src/utils/utils';
 
+import useSettingsStore from 'src/store/settings';
+
+const { getUmbraApiVersion, setUmbraApiVersion, clearUmbraApiVersion } = useSettingsStore();
+
 export class UmbraApi {
   static baseUrl = 'https://mainnet.api.umbra.cash'; // works for all networks
+  // static baseUrl = 'http://localhost:3000';  // use this for testing with a local Umbra API
```
alexkeating, line 23: May  be better to move this to comment above the baseUrl variable. Something like `To test locally change to ...`

frontend/src/utils/umbra-api.ts
```diff
@@ -10,17 +10,42 @@ import {
...
   constructor(
     readonly tokens: TokenInfoExtended[],
     readonly chainId: number,
     readonly nativeTokenMinSendAmount: string | undefined
   ) {}
 
+  static checkUmbraApiVersion(version: UmbraApiVersion) {
+    const apiVersionFromSettings: UmbraApiVersion | null = getUmbraApiVersion();
```
alexkeating, line 31: I think the return type can be removed here. It is better to be less explicit with types in typescript as these can hide type errors.

alexkeating (commented):
A couple small nits then should be good to merge.

frontend/src/utils/umbra-api.ts
garyghayrat: Oh interesting, is that a general tip for typescript? I thought using implicit types can hide type errors.

alexkeating: Yes, in my experience explicit types can lead to mistyping

jferas: Done: https://github.com/ScopeLift/umbra-protocol/pull/653/commits/c64845cd8120990d046476de35b3791f89ac1271

alexkeating (approved):

PR #655, opened 2024-03-05
Added a retry mechanism to getRegisteredStealthKeys

### LINKED ISSUE #654: Add retry to query for exising stealth address for user.
Sometimes it has been seen that the query fails, thus presenting the user with a "setup" button, which, if they press it causes their previously setup information to be lost.

jferas:
Resolves #654

### DIFF
--- a/frontend/src/store/wallet.ts
+++ b/frontend/src/store/wallet.ts
@@ -640,6 +640,9 @@ async function getRegisteredStealthKeys(account: string, provider: Provider) {
     } catch (err) {
       window.logger.warn(err);
       retryCounter++;
+      if (retryCounter < 3) {
+        await new Promise((resolve) => setTimeout(resolve, 2000)); // Wait for 2 seconds
+      }
     }
   }
   return null;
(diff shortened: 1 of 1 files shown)


frontend/src/store/wallet.ts
```diff
@@ -613,11 +613,16 @@ const hasSetPublicKeysLegacy = async (name: string, provider: Provider) => {
...
 // Helper method to check if user has registered public keys in the StealthKeyRegistry
 async function getRegisteredStealthKeys(account: string, provider: Provider) {
-  try {
-    const stealthPubKeys = await utils.lookupRecipient(account, provider); // throws if no keys found
-    return stealthPubKeys;
-  } catch (err) {
-    window.logger.warn(err);
-    return null;
+  let retryCounter = 0;
+  while (retryCounter < 3) {
```
garyghayrat, line 617: 1. Should we add timeout between tries?
2. I think we've discussed this, maybe we should try to query public providers if the first one fails. But maybe that won't work either because public providers are prob heavily rate limited.
3. I think the long term solution to this is querying a subgraph instead, which we should have an issue for, if not we should create one.

jferas: Not sure option 3 is something we'd want to expend effort on in Umba v1, especially as we've seen this failure pretty infrequently, correct?

jferas: Added a delay of 2 seconds for each of the 3 retries: https://github.com/ScopeLift/umbra-protocol/pull/655/commits/2f0d54b76f472a0230787bf69eea43a3a09450de

garyghayrat (approved):

alexkeating (approved):

PR #676, opened 2024-05-07
Use ContractScan for multichain deployment links

Destiner:
Replaces Blockscan with ContractScan. The latter only highlights the networks where the code is actually deployed (vs those that have no code but some state e.g. balance), supports over 100 networks (vs 50ish), and is open-source.

> Disclaimer: author of ContractScan

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,9 +135,9 @@ Umbra contracts are deployed at the same address on each network where Umbra res
 
 |Contract          |Type     |Address|
 |------------------|---------|-------|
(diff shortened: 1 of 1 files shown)


garyghayrat:
> Replaces Blockscan with ContractScan. The latter only highlights the networks where the code is actually deployed (vs those that have no code but some state e.g. balance), supports over 100 networks (vs 50ish), and is open-source.
> 
> > Disclaimer: author of ContractScan

Hi @Destiner , thank you for the suggestion and the PR. We currently don't have any plans to migrate over to a new service, but will keep your product in mind for future projects. Thanks!

PR #685, opened 2024-06-18
feat: save user announcements to local storage and last fetched block

marcomariscal:
## Description
Updates to fetch and scan only the latest blocks. The latest blocks are those after the initial scan, where each subsequent scan saves the latest fetched block to local storage, allowing for appropriate updating of the scan range:

1. Initial scan --> fetch/scan blocks 1 to x
2. Next scan  --> fetch/scan blocks x + 1 to latest block (y)
3. Subsequent scan  --> fetch/scan blocks y + 1 to latest block
etc.

### Methodology:
- User announcements are updated in local storage on each fetch/scan
- The latest fetched block is updated in local storage on each fetch/scan
- The most recent announcement block data is updated in local storage on each fetch/scan to accurately show if there are no announcements fetched in the current scan
- The `startBlock` for the scan is appropriately updated to the latest fetched block 

### Other Updates:
1. Show fetching/scanning latest verbiage above the top right of the `AccountReceiveTable`, intentionally omitting the large scan indicator at the bottom of the table
2. Implemented a clear local storage button with auto app refresh

### Comments:
- There were instances where querying the subgraph potentially hung (using goldsky, so outdated); while this may not be directly affected by this PR, it's worth noting for testing purposes

### To Test:
1. Perform the initial scan using account A
2. Navigate to a different route/page using account A
3. Validate: Navigate back to the receive page to confirm:
   a) The initial user announcements are present
   b) The fetching/scanning sequence is reflected appropriately (it's fetching/scanning only the latest blocks)
4. Perform a send transaction from a different account to account A, then navigate to the receive view connected with account A; validate as in step 3
5. Test the clear local storage functionality and ensure the app refreshes correctly


jferas (changes requested):
Ran into an interesting UX loose end.. connected my wallet and did scan for received transactions.. had to sign to get them.. did so and got the list displayed.. went back to home and hit the receive tab again.. got the list immediate. Very nice...

Then  closed that browser tab.. went back to the app via the deploy preview URL.. went to scan for receives, and got the list I had previously as well as a request to sign.  Something might need to be cleared?
Screenshot 2024-06-25 at 3 37 56 PM

marcomariscal:
> Ran into an interesting UX loose end.. connected my wallet and did scan for received transactions.. had to sign to get them.. did so and got the list displayed.. went back to home and hit the receive tab again.. got the list immediate. Very nice...
> 
> Then closed that browser tab.. went back to the app via the deploy preview URL.. went to scan for receives, and got the list I had previously as well as a request to sign. Something might need to be cleared? Screenshot 2024-06-25 at 3 37 56 PM

Yee, that's intentional. When the user refreshes the page, they need to sign again to scan because the keys are no longer available in memory. Sensitive data won't be cached to ensure security and privacy.

We've updated the sign-to-scan verbiage slightly for this scenario but welcome any feedback on clearer or less confusing language, and the ux in general.

jferas (approved):

jferas:
> Yee, that's intentional. When the user refreshes the page, they need to sign again to scan because the keys are no longer available in memory. Sensitive data won't be cached to ensure security and privacy.
> 
> We've updated the sign-to-scan verbiage slightly for this scenario but welcome any feedback on clearer or less confusing language, and the ux in general.

Understood.. thanks!

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -60,12 +60,45 @@
         >.
```
garyghayrat, line 60: I don't think you need to sign to withdraw funds?
image

garyghayrat: I see why you added that. But would it be too much trouble to change it to asking people to sign when they lick on withdraw instead.

marcomariscal: cool will update

marcomariscal: https://github.com/ScopeLift/umbra-protocol/pull/685/commits/29f6f6ecf25f892f33e0f4e8d027852ecf78c72a

updated the wallet store to have a `needSignature` computed val to be able to use freely throughout...intentionally used that value only in the relevant parts for this pr. used this value to update the label and disabled props on the withdraw form button.

garyghayrat: I think it would be better if we made the `Need Signature` button a `Sign` button if we don't have user signature, and then change it to `Withdraw` like it used to be once the user signs.

frontend/src/layouts/BaseLayout.vue
```diff
@@ -223,6 +223,21 @@
               option-label="label"
               rounded
             />
+
+            <!-- Clear local storage button -->
+            <div class="q-mt-md row items-center">
```
garyghayrat, line 228: Let's hide the clear storage button unless the advanced mode is on.

frontend/src/i18n/locales/en-US.json
```diff
@@ -33,6 +33,8 @@
     "contact": "Contact"
   },
   "Base-Layout": {
+    "clear-local-storage": "Clear local storage",
+    "clear-local-storage-description": "Clears all local storage data, including send history, and wallet connection.",
```
garyghayrat, line 37:
```suggestion
    "clear-local-storage-description": "Clears all local storage data, including send history, announcements, and wallet connection.",
```

frontend/src/layouts/BaseLayout.vue
marcomariscal: might be confusing because the advanced mode doesnt utilize the caching functionality? only the default fetching/scanning functionality has been updated to incorporate caching.

garyghayrat: Hmm, didn't realize it wasn't supported in advanced mode. I guess it makes sense that we don't. 

Could we move it above the language selector then?

frontend/src/i18n/locales/zh-CN.json
```diff
@@ -33,6 +33,8 @@
     "contact": "联系"
   },
   "Base-Layout": {
+    "clear-local-storage": "清除本地存储",
+    "clear-local-storage-description": "清除所有本地存储数据，包括发送历史记录和钱包连接信息。",
```
garyghayrat, line 37:
```suggestion
    "clear-local-storage-description": "清除所有本地存储数据，包括发送历史记录，公告，和钱包连接信息。",
```

frontend/src/components/AccountReceiveTable.vue
marcomariscal: updated:

https://github.com/ScopeLift/umbra-protocol/pull/685/commits/a211d4623617bfc752678eda2025280753219958

frontend/src/layouts/BaseLayout.vue
marcomariscal: https://github.com/ScopeLift/umbra-protocol/pull/685/commits/ba09324b2bb07d4115de28ac9bcbe3dcf9f07ec1

frontend/src/components/AccountReceiveTable.vue
garyghayrat: Still some issues: 

https://github.com/ScopeLift/umbra-protocol/assets/61768337/75f79027-83b3-4249-949d-e6984b0571f6

frontend/src/layouts/BaseLayout.vue
garyghayrat: Sorry for the back and forth, but if clearing the local storage is also inevitably resetting the send history, could we just make the current "Send history on/off" toggle clear the local storage when it's turned off? We can rename the toggle to "Send and Announcement history" or something and update the `?` description accordingly.

The extra clear storage button just seems so out of place no matter where it is placed.

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -60,12 +60,45 @@
...
       </div>
 
-      <div v-if="scanStatus === 'complete'" class="text-caption q-mb-sm">
-        <!-- Show the most recent timestamp and block that were scanned -->
```
garyghayrat, line 64: Another nit, could we make the bottom padding for the sign buttong the same as the same as the top padding?
image

frontend/src/layouts/BaseLayout.vue
marcomariscal: Adding a toggle like "Send and Announcement History On" would add a bit more complexity to the current announcement caching and scanning logic. I'd rather have the send history and announcement handling be separate concerns for now.

frontend/src/components/AccountReceiveTable.vue
marcomariscal: https://github.com/ScopeLift/umbra-protocol/pull/685/commits/a72ff26f40e576630de8b78bfc4417ebb4fdfb26

- made the spacing more uniform between the scanning sign verbiage, the sign button, and the most recent announcement data
- updated the advanced setting form vertical spacing
- made the receive header spacing match the send view

frontend/src/layouts/BaseLayout.vue
garyghayrat: Shouldn't it be just a change of:

`{{ sendHistorySave ? $t('Base-Layout.on') : $t('Base-Layout.off') `**&& clearLocalStorage()**` }}</span`

I haven't fully convinced myself that combining those two features is a good idea either, so yea, maybe we should leave it for now.

frontend/src/components/AccountReceiveTable.vue
marcomariscal: https://github.com/ScopeLift/umbra-protocol/pull/685/commits/d8b730b65a142fcd8beb5712d29c081a3c546a2b

frontend/src/layouts/BaseLayout.vue
marcomariscal: Mmm not sure that would work. 

For clearing send history it's simply clearing local storage when the toggle is off. But if we need similar functionality for announcements, ie. announcement toggle is off, then will need to add corresponding logic to the AccountReceive component and update the default scan to not save user announcements, etc.

Not sure if worth pursuing.

frontend/src/components/AccountReceiveTable.vue
garyghayrat: one last nit, let's rename the button name to "Withdraw" .

garyghayrat: or maybe "Sign and withdraw"

marcomariscal: https://github.com/ScopeLift/umbra-protocol/pull/685/commits/5d63f2a6a20968be854e2b8e2b58cc2bef5be880

frontend/src/components/AccountReceiveTable.vue
```diff
@@ -99,12 +132,8 @@
           <div :key="props.row.id" class="col-12">
             <q-card class="card-border cursor-pointer q-pt-md col justify-center items-center">
               <q-card-section class="row justify-center items-center">
-                <img
-                  class="q-mr-md"
-                  :src="getTokenLogoUri(props.row.token, tokens)"
-                  style="width: 1.2rem"
-                  v-if="getTokenInfo(props.row.token)"
-                />
+                <img v-if="getTokenInfo(props.row.token)" src="path/to/image.jpg" />props.row.token, tokens)"
```
garyghayrat, line 135: is that src a mistake?

garyghayrat: should it be `:src="getTokenLogoUri(props.row.token, tokens)"` ?

marcomariscal: updated to whats in master:

https://github.com/ScopeLift/umbra-protocol/pull/685/commits/b6278897deab200b572900b6ffd6ad76fb4d8d85

garyghayrat (approved):

garyghayrat:
Thank you @marcomariscal !

PR #688, opened 2024-07-03
fix: contract periphery tests

marcomariscal:


jferas (approved):
Looks good

contracts-periphery/script/DeployBatchSend.s.sol
```diff
@@ -22,15 +22,14 @@ contract DeployBatchSend is Script {
   uint256 constant EXPECTED_NONCE = 0; // TODO Edit this with the nonce of the deployer address
   address constant UMBRA = 0xFb2dc580Eed955B528407b4d36FfaFe3da685401;
   // The list of networks to deploy to.
-  string[] public networks =
-    ["mainnet", "optimism", "arbitrum_one", "polygon", "gnosis_chain", "sepolia"];
+  string[] public networks = ["mainnet", "optimism", "arbitrum_one", "polygon", "sepolia"];
```
garyghayrat, line 25: why did we remove gnosis chain here?

marcomariscal: accidentally left out for testing. reverted here:

https://github.com/ScopeLift/umbra-protocol/pull/688/commits/2ada97d152eb9278d8cee5202f4d8aaa2a90f33c

jferas: I think the RPC URL env var for gnosis chain was not supported on infura, so rather than commenting it out we should be able to use this for the env var on CI (works for me locally):
GNOSIS_CHAIN_RPC_URL=https://rpc.gnosischain.com

garyghayrat (approved):

PR #694, opened 2024-09-30
fix: show addr not registered error when no event data from subgraph

marcomariscal:
updates the error verbiage when no stealth key changed event data is returned from the subgraph to "Address has not registered stealth keys..." since they probably (near certainly) didn't register yet (hence no events).

originally it would show an error message that potentially wasn't as clear.


jferas (approved):
Looks good

garyghayrat (approved):
thanks!

PR #712, opened 2026-03-11
Support Ponder subgraph schemas in umbra-js and frontend

garyghayrat:
## Summary
- add Ponder-compatible subgraph queries for announcement and stealth key scans
- auto-detect Ponder vs legacy subgraph schemas and fall back when Ponder fields are unavailable
- wire the frontend to the new subgraph configuration and add schema compatibility coverage

## Testing
- add `umbra-js/test/subgraph-schema.test.ts`

## Results
Older accounts benefit the most, and newer accounts not as much (like the base example).
| Chain | Before | After | Time Saved | % Improvement |
|---|---:|---:|---:|---:|
| Mainnet | 18s | 7s | 11s | 61.1% |
| Optimism | 48s | 27s | 21s | 43.8% |
| Polygon | 31s | 15s | 16s | 51.6% |
| Base | 8s | 7s | 1s | 12.5% |
| Arbitrum | 61s | 42s | 19s | 31.1% |

**Average percentage improvement across chains: 40.0%**


marcomariscal (changes requested):
Requesting changes: PONDER_SUBGRAPH_URL is now required in practice, but missing it silently disables subgraph scanning. The current deploy preview bundle has no Ponder URL, so L2 receive scans fall back to RPC and throw on Optimism/Polygon/Base.

Please fail the build or keep the legacy per-chain fallback until the new env is configured everywhere. Also add a smoke test we can run against preview/prod that verifies the built app has a Ponder URL and a basic Ponder scan succeeds.

garyghayrat:
> Requesting changes: PONDER_SUBGRAPH_URL is now required in practice, but missing it silently disables subgraph scanning. The current deploy preview bundle has no Ponder URL, so L2 receive scans fall back to RPC and throw on Optimism/Polygon/Base.
> 
> Please fail the build or keep the legacy per-chain fallback until the new env is configured everywhere. Also add a smoke test we can run against preview/prod that verifies the built app has a Ponder URL and a basic Ponder scan succeeds.

https://github.com/ScopeLift/umbra-protocol/pull/712/commits/db73aba22bce0a461f6c6acd385af72c2aa1137a

marcomariscal (approved):
Looks good. Gary addressed the requested fallback/build-guard issue and added the preview/prod Ponder smoke test. I ran the smoke test against the Netlify preview and it passed.

marcomariscal (approved):
Reviewed the latest commits through 8690a895. The Netlify Ponder proxy smoke test passes, and I found no blocking issues.

apbendi (approved):
Tested locally and everything is running great. Really nice work on this @garyghayrat. Excited to get this into prod.

PR #713, opened 2026-03-11
Checkpoint receive scans to subgraph head block

garyghayrat:
## Summary
This changes the Receive page checkpointing logic so subgraph-backed scans advance to the subgraph head block instead of the RPC latest block.

## Why
On Base and other subgraph-dependent chains, the previous logic stored the provider latest block after a scan. If the subgraph lagged behind the RPC head, a newly indexed announcement could be skipped permanently because future scans would start after the missed block.

## What changed
- fetch the subgraph head block from `_meta`
- support both Ponder `_meta.status` and legacy `_meta.block.number` shapes
- use that head block as the next cached `lastFetchedBlock`
- fall back conservatively to the prior scan checkpoint if the head-block lookup fails

### DIFF
--- a/frontend/src/pages/AccountReceive.vue
+++ b/frontend/src/pages/AccountReceive.vue
@@ -391,6 +391,95 @@ function useScan() {
     return provider.getBlock('latest');
   }
 
+  function getPonderNetworkName(chainId: number): string | null {
+    switch (chainId) {
+      case 1:
+        return 'mainnet';
+      case 10:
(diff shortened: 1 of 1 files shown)


marcomariscal (approved):
Looks good. I verified these response shapes are real by testing both the Ponder _meta.status endpoint and the legacy _meta.block.number subgraph endpoints.

apbendi (approved):
looks good

PR #715, opened 2026-04-30
feat: add turnkey support

garyghayrat:


frontend/src/components/AccountReceiveTable.vue
```diff
@@ -564,8 +572,11 @@ function useReceivedFundsTable(userAnnouncements: Ref<UserAnnouncement[]>, spend
       return;
     }
     isFeeLoading.value = true;
-    activeFee.value = await relayer.value?.getFeeEstimate(tokenAddress);
-    isFeeLoading.value = false;
+    try {
+      activeFee.value = await relayer.value?.getFeeEstimate(tokenAddress);
```
marcomariscal: I think this can race. Expanding the row starts one fee request, then starting the withdrawal starts another. If the first one finishes last, it can overwrite the fee for the active withdrawal. Could we keep the fee for this withdrawal as its own value and pass that through confirmation/signing? A test with two out-of-order fee responses would catch it.

garyghayrat: https://github.com/ScopeLift/umbra-protocol/pull/715/commits/1eddafb1d3cd13a26899d200598656287d7a3d94

marcomariscal (approved):
Looks good. The fee estimate race is addressed with the isolated withdrawal fee and out-of-order response test.

apbendi (approved):
@garyghayrat Tested myself with Optimism and Mainnet. Funded the wallet across all accounts. Let's coordinate on Monday to get it rolled out to prod.

